PR #3, opened 2017-12-06
Rename converge package to be scoped to bigtest org

wwilsman:


cowboyd:
If we're calling our primitive `Convergence`, then perhaps we should name the test case that?

cowboyd:
Also, if we're moving the scope, then directories should probably be:

```
packages/convergence
packages/mocha
packages/interaction
````

wwilsman:
@cowboyd after a few git hiccups, ready for another review

Issue #2, opened 2017-10-13
Fill out README


Robdel12:
It would be nice to have a small getting started / how to use section and why this repo exists (what problem is it solving?)

PR #4, opened 2017-12-06
Add helper to converge on a series of convergences

wwilsman:
This is useful for when you need to chain multiple convergences but still need to constrain the total timeout period for all of them.

For example, when clicking a button you would first want to converge on the existence of the button, perform the click, then converge on any resulting side effects; but that whole transaction must take place within the constraints a single timeout (a `beforeEach` block for example).

I saw two options to solve this:

1. Track the first `convergeOn` start time so every `convergeOn` that is called within a short period of time can properly track the elapsed time since the first one started.

2. Set the timeout of the following convergent function to be the total timeout, minus the time it has taken any previous convergent functions to complete.

`#1` seemed sub-optimal, as you wouldn't then be able to have unrelated convergences running at the same time.

To achieve `#2`, a convergent function must be able to have it's timeout modified before being executed, so that dealing with the math involved could be handled by another helper. Using the `convergent` helper, you can now modify it's timeout after creation by changing the `timeout` property of the resulting function.

```js
let converge = convergent(() => expect(foo).to.equal('bar'));

console.log(converge.timeout);
// => 2000

converge.timeout = 50;

console.log(converge.timeout);
// => 50

// will fail if it hasn't converged after 50ms
converge().then(...).catch(...);
```

With this ability, the `convergeSeries` helper can now track the total time elapsed, and set the timeouts of other convergent functions accordingly. 

The following will resolve once `Page.clickCount` equals `1`, but _only_ after a button has been successfully clicked:

```js
convergeSeries([
  convergent(() => expect(Page.$btn).to.exist),
  () => Page.$btn.get(0).click(), // this would error if it didn't exist in the DOM
  convergent(() => expect(Page.clickCount).to.equal(1))
], 50);
```

### DIFF
--- a/packages/bigtest-converge/src/index.js
+++ b/packages/bigtest-converge/src/index.js
@@ -115,7 +115,65 @@ export function convergeOn(assertion, timeout = 2000, invert) {
  *   // setTimeout(() => total = 0, 400);
  */
 export function convergent(assertion, timeout, invert) {
-  return function(...args) {
-    return convergeOn.call(this, assertion.bind(this, ...args), timeout, invert);
+  let converge = function converge(...args) {
+    return convergeOn.apply(this, [
+      assertion.bind(this, ...args),
+      converge.timeout,
+      invert
+    ]);
   };
+
+  // allows modification of the timeout before execution
+  converge.timeout = timeout;
+
+  // flags this function as a convergent one
+  converge.convergent = true;
+
+  return converge;
+}
+
+/**
+ * Convergent helper that ensures a set of convergent functions all
+ * converge after one another within a timeout period. If a provided
+ * function is not convergent, it is simply invoked with the previous
+ * resolved value and its return value is passed to the next function
+ * in the series.
+ *
+ * This function works by modifying the timeout period of convergent
+ * functions created using the convergent helper above so that the
+ * combined timeout period does not exceed a given value.
+ *
+ * Example:
+ *   beforeEach(() => {
+ *     return convergeSeries([
+ *       convergent(() => {
+ *         let $el = $(selector);
+ *         expect($el).to.exist;
+ *         return $el.get(0);
+ *       }),
+ *       (node) => {
+ *         node.click();
+ *       },
+ *       convergent(() => {
+ *         expect(sideEffect).to.be.true;
+ *       });
+ *     ]);
+ *   });
+ *
+ * @param {[Function]} funcs - Array of possibly convergent functions
+ * @param {Number} [timeout=2000] - Timeout that all functions must
+ * converge within
+ */
+export function convergeSeries(funcs, timeout = 2000) {
+  let start = Date.now();
+
+  return funcs.reduce((promise, fn) => {
+    let ellapsed = Date.now() - start;
+
+    if (fn.convergent) {
+      fn.timeout = timeout - ellapsed;
+    }
+
+    return promise.then(fn);
+  }, Promise.resolve());
 }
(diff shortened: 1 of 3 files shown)


cowboyd:
I'd like to reconsider rethinking a convergent assertion as a `Functor` akin to `Promise`, and `Observable`. But the "computational context" instead of being a future value, or a stream of values, it's a state that is eventually met, or not met.

```javascript
let convergence = new Convergence(2000)
  .and(() => expect(1).to.equal(1))
  .and(() => expect(2).to.equal(2))
```

We can run this convergence any number of times to get a new promise:

```javascript
convergence.run().then(() => console.log('yay!!!')).catch((error) => console.log('boooo'))
convergence.run().then(() => console.log('yay!!!')).catch((error) => console.log('boooo'))
```

We can adjust the timeout by chaining onto the original convergence (which, like promise and observable does not change the original)

```javascript
convergence
  .timeout(2000).run().then(() => console.log('yay!!!')).catch((error) => console.log('boooo'))
```

It can also be a monoid

```javascript
import { reduce } from 'funcadelic';

f1() {
  expect(1).to.equal(1)
}
f2() {
  expect(2).to.equal(2)
}
f3() {
  expect(3).to.equal(3)
}

reduce(Convergence, [f1, f2, f3]).run().then(() => console.log('yay!'))
```

ultimately, this can be rolled up into input helpers:

```
interactive()
  .click('[button-selector]')
  .type('[input-selector]', 'some text')
  .scroll('[pane-selector]', {y: -300});
```

Clearly some hand-waving here, but I feel it's worth exploring.

Robdel12:
Don't mind my drive-by review but this syntax seems odd to me, but I'm going to assume I'm missing the use-case

```js
convergeSeries([
  convergent(() => expect(Page.$btn).to.exist),
  () => Page.$btn.get(0).click(), // this would error if it didn't exist in the DOM
  convergent(() => expect(Page.clickCount).to.equal(1))
], 50);
```

I'd probably write that like:

```js
beforeEach(()=> {
  Page.$btn.get(0).click(); // error if not in DOM, takes care of the first expect in the example
});

it('increments the button', () => {
  expect(Page.clickCount).to.equal(1);
});
```

I could see how this would be useful as I typed that out, but I'm struggling to think of an example

cowboyd:
@Robdel12 With the syntax I'm proposing, the mocha syntax would probably look like:

```javascript
beforeEach
  .click(Page.$btn)
```

or maybe

```javascript
import { click } from '@bigtest/helpers';

beforeEach(() => click(Page.$btn))
```

That would be the "high-level" syntax

The "low-level" syntax it would "compile" to

```javascript
beforeEach(() => {
  return new Convergence(this.timeout)
    .and(() => expect(Page.$btn).to.exist))
    .run().then(() => Page.$btn.get(0).click())
})
```

wwilsman:
@Robdel12 The issue can be seen in @cowboyd's example:

```
interactive()
  .click('[button-selector]')
  .type('[input-selector]', 'some text')
  .scroll('[pane-selector]', {y: -300});
```

If you wanted to do that in a `beforeEach`, you'd have to converge on the existence of the button and the input, but click them and adjust their values in between convergences. Say for example you couldn't type in that input until the button is clicked first.

Also, in your solution if this was a nested `describe` block the `beforeEach` would just error and stop your tests from continuing. You'd have to make sure the button exists before you click it, otherwise your tests will never run.

wwilsman:
@cowboyd I ❤️ that proposed syntax. We can just close this and that'll be my next PR tonight. 😄

PR #6, opened 2017-12-08
Convergence Class

wwilsman:
## Purpose
We need the ability to chain convergent assertions together, but still constrain their total allotted timeout amount.

## Approach
With the addition of this convergence class, you can chain convergent assertions together like promises and observables.

```js
new Convergence()
  .timeout(200)
  .once(() => expect(Page.$btn).to.exist)
  .do(() => Page.$btn.click)
  .always(() => expect(Page.$btn).to.exist)
```

Like promises and observables, each chain method creates a new instance of `Convergence` with the assertion added to a copy of the previous convergence instance's stack.

```js
// Will converge on the `$root` element existing in the DOM
let converge1 = new Convergence()
  .once(() => expect(Page.$root).to.exist);

// After converging on #1, will converge when the `$root` element exists 
// for the remainder of the timeout period
let converge2 = converge1
  .always(() => expect(Page.$root).to.exist);
```

Convergences are not run immediately. To kick off converging on all chained assertions, call the `.run()` method of the convergence instance. The `.run()` method returns a promise that will resolve with a `stats` object, and reject when any of the convergent assertions fail.

```js
// kicks off convergence #1 from above
converge1.run().then(() => {
  // resolves once the `$root` element exists in the DOM
}).catch(() => {
  // rejects when the `$root` element never exists throughout the 
  // entire timeout period
});

// kicks off convergence #2 from above
converge2.run().then(() => {
  // resolves after `$root` exists, and has existed for the duration of the timeout
}).catch(() => {
  // rejects when `$root` cannot be found, or when `$root` is removed from
  // the DOM within the timeout period
});
```

The `stats` object provided after the convergence resolves contains useful information about the execution of the convergent assertions:

- `start` - the timestamp created just before starting the convergent assertions
- `end` - the timestamp created when the final convergence resolves
- `elapsed` - the total elapsed time, in `ms`, for all assertions to converge
- `runs` - the total number of times functions within this convergence have been run
- `timeout` - the timeout used for this set of convergences
- `value` - the value returned from the final function in the stack
- `stack` - an array of stats for each assertion in the convergence instance's stack (see below)

To facilitate this stats object, the ability to track stats using `convergeOn` has also been added. A fourth argument, `useStats`, can now be provided to `convergeOn` and the promise returned will resolve with a `stats` object similar to the above:

- `start`, `end`, `elapsed`, `runs`, `timeout`, and `value` are available on this stats object as well, however the stats are for this individual convergent assertion
- `always` - true when the convergence was inverted, either by `always = true` or when using the `.always()` method of a Convergence instance

## Testing
- Added `babel-polyfill` to enable the regenerator runtime for the `async`/`await` syntax. This might also be helpful for future packages, so it was added to the `devDependencies` of the entire repo.
- The tests for the old `convergent` helper have been replaced with tests for the new `Convergence` class. But all of the old `convergent` tests have been moved to `converge-on-test.js` and modified slightly to work with `convergeOn` directly.


mocha.opts
```diff
@@ -0,0 +1,2 @@
+--compilers js:babel-register
+--require babel-polyfill
```
Robdel12: woah, you can require deps as a command line flag? That's dope

cowboyd:
Awesome! Reading the syntax in real life, it makes me wonder if we should line up the `still` `and` and with more well understood boolean / set operations.

Some alternatives:

| Current | JS Set    | Rb Set | Timish     | Horse Math |
|---------|---------|--------|----------|-------------|
| `and`    | `some` | `any`   | `once`    | `instance`  | 
| `still`    | `every` | `all`     | `always` | `invariant`  |

I think we could also push inversion into the `Convergence` definition. E.g. 

```javascript
  new Convergence()
    .still(() => expect(Page.$btn).to.exist))
    .not() //=> never.
```

packages/convergence/src/convergence.js
```diff
@@ -0,0 +1,213 @@
...
+   * this convergence instance. When a running convergence instance
+   * encounters a `.tap()`, it will invoke the callback with the value
+   * returned from the last function in the stack. The resulting
+   * return value will also be provided to the following function in
+   * the stack
+   *
+   * @param {Function} exec - the callback to execute during the convergence
+   * @returns {Convergence} a new convergence instance
+   */
+  tap(exec) {
```
cowboyd: How do you see the primary use-case of `tap`. As a debugging tool?

wwilsman: Debugging, and sometimes we might need to do something like this: 

```js
new Convergence()
  .and(() => expect(Page.$submit).to.exist)
  .tap(() => Page.submitForm(...))
  .still(() => expect(Page.$newItem).to.exist)
```

Then make assertions on this new item

wwilsman:
@cowboyd renamed the methods to `once`, `always`, and `do`. Also changed the `invert` flag for `convergeOn` to also be `always` as that better describes what the convergence does when "inverted".

PR #9, opened 2017-12-15
Add convergence docs

wwilsman:
[Rendered](https://github.com/thefrontside/bigtest/blob/ww/convergence-docs/packages/convergence/README.md)

### DIFF
--- a/packages/convergence/README.md
+++ b/packages/convergence/README.md
@@ -39,3 +39,252 @@ that condition has been met.
 And it isn't just for assertions either. Because it is a general
 mechanism for synchronizing on any observed state, It can be used to
 properly time test setup and teardown as well.
+
+## Creating a Convergence
+
+A convergent function is a function that runs repeatedly until it no
+longer returns false or throws an error. When the function is finally
+successfully executed, it is considered to be passing and a converging
+promise will resolve. However, if the converging function does not pass
+within the provided timeout period the promise will reject with the
+last error thrown from the function. A converging promise will reject
+right before the timeout so it is never reached during it's execution.
+
+### Using `Convergence`
+
+The `Convergence` class allows you to create a convergence, add
+assertions to converge on, and execute the convergent assertions all
+within a single timeout period.
+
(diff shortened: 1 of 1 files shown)


packages/convergence/README.md
```diff
@@ -39,3 +39,336 @@ that condition has been met.
...
+  stats.elapsed  // time taken to converge on the entire stack
+  stats.runs     // number of times functions in the stack ran
+  stats.timeout  // this Convergence instance's timeout
+  stats.value    // value returned from the last function in the stack
+  stats.stack    // an array of stats objects from the stack
+});
+```
+
+Because `.run()` returns a promise, and `Convergence` is immutable,
+you can call `.run()` multiple times on the same convergence instance.
```
wwilsman: I feel like this README ends kind of abruptly. Maybe some section to tie it all up? Or in the future we could link to other `@bigtest/*` packages that use `@bigtest/convergence`. Thoughts?

packages/convergence/README.md
```diff
@@ -39,3 +39,336 @@ that condition has been met.
 And it isn't just for assertions either. Because it is a general
 mechanism for synchronizing on any observed state, It can be used to
 properly time test setup and teardown as well.
+
+## Creating a Convergence
+
+A convergent function is simply a function that is ran
```
cowboyd: ~simply~

cowboyd: s/ran/run

packages/convergence/README.md
```diff
@@ -39,3 +39,336 @@ that condition has been met.
...
+repeatedly until it no longer throws an error. When the function is
```
cowboyd: > as long as it does not return false or throw an error
right? A single false will kill it.

wwilsman:
@cowboyd Updated based on feedback. I removed references to `convergeOn` from the `Convergence` docs. But I left a bit about the low-level notice for `convergeOn` as a sort of footnote to the rest of it.

PR #10, opened 2017-12-22
Per package scripts

wwilsman:
## Purpose
When adding another package, it became clear that it would be difficult to run different test runners across our packages from one command. `@bigtest/convergence` uses Mocha, but soon we will need Karma for browser tests, and maybe Selenium for iOS or Android in the future. With this in mind, our builds might also be different across packages, so just using Gulp might not cut it in the future.

## Approach
Created a script, `bin/each-pkg`, that will run any command in every package directory. This worked well enough in bash, but for deploying packages we want to compare versions, and that is easier in node. Using node also opened up better implementation of flags for the CLI and we can utilize promises to run our commands in series (or not and run them in parallel).

All of the dev dependencies of `@bigtest/convergence` have been moved into that package's `package.json` instead of resting at the workspace root. If other packages end up having the same dependencies, yarn will intelligently lift them up into the workspace's `package.json`.

With the new `each-pkg` command, with the `--changed` flag, we can now run tests only for packages that have changed. And with the `--newer` flag, we can only publish packages that have bumped their version number. The `.circleci/config.yml` has been updated to utilize workflows so that `master` can have a separate set of jobs oppose to other branches. The `master` workflow runs all tests, builds all packages, and publishes any new ones. The `branch` workflow runs tests and builds packages that have diverged from `origin/master`.

### Open Questions
- Should we run all tests anyway on other branches? Maybe it could surface issues with packages that depend on one another. But then again, if there's a breaking change that should require a new version, and the other package should still depend on the older version.

- In the future, we can run tests and builds for packages in parallel, either with the `each-pkg` command, or with Circle's workspaces (we would need a job for each package).


.circleci/config.yml
```diff
@@ -1,22 +1,124 @@
...
       - run:
           name: Install dependencies
           command: yarn
+      - save_cache:
+          name: Save cache
+          key: *cache_key
+          paths:
+            - node_modules
+            # globs don't work for save_cache :(
+            - packages/convergence/node_modules
```
wwilsman: 😢

wwilsman:
Oh, I also removed Gulp and instead went with (Rollup)[https://rollupjs.org/] to build the `@bigtest/convergence` package

Robdel12:
Wil, this is exceptional. The PR description tells me exactly what's going on, why the changes, and the reasoning behind them. 

I dig not running the tests for packages that haven't changed. Do any of these packages depend on each other? Like if I make a change in one, could it possibly break in another? If that's possible I'd prefer to run all the tests all the time. Since master runs them all anyways you might find out about a test failure until it's too late.

bin/each-pkg
```diff
@@ -0,0 +1,112 @@
...
+const fs = require('fs');
+const path = require('path');
+const chalk = require('chalk');
+const semver = require('semver');
+const program = require('commander');
+const { spawn, spawnSync } = require('child_process');
+
+/**
+ * Gets the package.json contents of each package located in the
+ * `./packages` subdirectory along with the package's path
```
Robdel12: Ha, super stylistic, but can there be a space between this line and the `@ returns`?

packages/convergence/rollup.config.js
```diff
@@ -0,0 +1,17 @@
...
+    name: 'BigTest.Convergence'
+  },
+  plugins: [
+    babel({
+      babelrc: false,
+      presets: [['env', { modules: false }]],
+      plugins: ['external-helpers', 'remove-comments']
+    })
+  ]
+};
```
Robdel12: Woah, rollup config looks much nicer than webpack

wwilsman:
@Robdel12 They will depend on each other, and the tests _should_ catch bugs first. But even if they make it to master, nothing will be published unless _all_ jobs have successfully finished (test & build for each package). So there would be time to catch it before it might be in production somewhere.

bin/each-pkg
wwilsman: Added a space after all of the descriptions since that's what we're doing in the code so far

cowboyd:
> The master workflow runs all tests, builds all packages, and publishes any new ones.

This is badassery pure and simple. I agree with Rob that we should always run all tests until it becomes prohibitively expensive to do so. Unless it's too burdensome, the sooner we can surface failures the better.

wwilsman:
Okay, I've adjusted the Circle workflows to be a single workflow with the `deploy` job only happening on master. This should be fine for now since we will have so few packages at first, but in the future we'll definitely want to run our tests and builds in parallel to cut down on the time it would take the entire workflow to run

.circleci/config.yml
```diff
@@ -1,22 +1,84 @@
...
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/bigtest
+      - run:
+          name: Setup NPM
+          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
+      - run:
+          name: Publish to NPM
+          command: ./bin/each-pkg -n npm publish
```
cowboyd: Small nit: let's use the full `--newer` command line option for this so that someone dropping into this script will get more information about what this flag is doing.

In general, I prefer long form options inside scripts where they will be read more often than typed, and short form options on the CLI where they are typed more often than read.

wwilsman: Fixed

wwilsman:
Circle says this passed, but it looks like it didn't report the final status to github? Or maybe github is the one acting up. Anyhow I can see that everything is green on Circle, so I'm just going to merge this

cowboyd:
I think this is because the workflow changed maybe?

wwilsman:
That could be it

Robdel12:
[Aweeee yiiissss](http://i0.kym-cdn.com/photos/images/original/000/635/934/77a.gif)

PR #11, opened 2017-12-22
Fix Circle deploy job

wwilsman:
~The `npm publish` command needs `node_modules` from the cache~

The new `each-pkg` script needs `node_modules` from the cache

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,6 +55,7 @@ jobs:
     <<: *defaults
     steps:
       - checkout
+      - *restore_cache
       - attach_workspace:
           at: ~/bigtest
       - run:


wwilsman:
Hmmm. Looks like this one is not reporting the status either. Maybe it's because the last job in the workflow is master only so it thinks it isn't done with the workflow?

Is this a bug we should report, or do we just go back to having two separate workflows with filters?

wwilsman:
I think I found a fix [here](https://circleci.com/docs/2.0/workflows-waiting-status/)

Robdel12:
So this PR probably isn’t needed? Were you able to make the changes to the repo?

wwilsman:
This PR is needed to get deploy working on master

PR #23, opened 2018-01-09
Add @bigtest/mocha README

wwilsman:
[rendered](https://github.com/thefrontside/bigtest/blob/ww/mocha-readme/packages/mocha/README.md)

### DIFF
--- a/packages/mocha/README.md
+++ b/packages/mocha/README.md
@@ -1,1 +1,175 @@
 # @bigtest/mocha
+
+Convergent Mocha functions for testing against asynchronous states
+
+## Synopsis
+
+``` javascript
+import { describe, beforeEach, it } from '@bigtest/mocha';
+import { expect } from 'chai';
+
+describe('clicking my button', () => {
+  beforeEach(() => $button.click());
+
+  // repeatedly asserts until passing
+  it('shows a loading indicator', () => {
+    expect($button.className).to.include('is-loading');
+  });
+
+  // repeatedly asserts it is passing until the timeout
+  it.always('does not navigate away', () => {
+    expect(app.location).to.equal('/')
+  }).timeout(200);
+})
+```
+
+## Convergent Assertions
+
+Typically, when testing asynchronous states (such as rendered content
+in an application) your tests need to run at just the right moment so
+that they are executed in the correct context. If your tests run too
+soon, they will fail; if they run too slow, well then you just have
+slow tests.
+
+Converging on a state means asserting against a state until the
+assertion passes, in which case you have successfully converged on
+that state!
+
+This package uses
+[`@bigtest/convergence`](https://github.com/thefrontside/bigtest/tree/master/packages/convergence)
+to repeatedly run assertions until they pass, or until the timeout has
+almost expired. Performing tests in this way allows them to pass the
+moment the desired state is achieved. This results in very fast tests
+when testing asynchronous things.
+
+_Read the `@bigtest/convergence` [docs on
+convergences](https://github.com/thefrontside/bigtest/tree/master/packages/convergence#why-convergence)
+for more info as to why converging on a desired state is better than
+trying to time it properly._
+
+## How does it work?
+
+For the most part, you write tests in the exact same way that you're
+used to writing tests with Mocha. The only difference is that this
+package wraps Mocha's `it` in a convergence helper so that any
+assertions you write using `it` become convergent assertions that
+allow you to easily test asynchronous states.
+
+### Writing Tests
+
+Because convergent assertions are run repeatedly until they pass, it
+is _highly recommended_ that you **do not** perform any side-effects
+in your assertions. This will result in your side-effect being run
+multiple, perhaps even hundreds of times.
+
+For this reason, you should keep your side-effect producing code in
+hooks and **out of your assertions.** These "pure assertions" also
+help your tests be more readable and explicit.
+
+#### :no_entry_sign: do not do this:
+
+``` javascript
+describe('my button', () => {
+  it('shows a loading indicator after clicking', () => {
+    // this will be called every time the assertion runs
+    $button.click();
+
+    // it might take a few milliseconds for any side-effects to
+    // happen, so this might fail the first time and cause this entire
+    // assertion to run again, thus clicking the button again
+    expect($button.className).to.include('is-loading');
+  });
+});
+```
+
+#### :white_check_mark: do this:
+
+``` javascript
+describe('clicking my button', () => {
+  // keep side-effects inside hooks
+  beforeEach(() => $button.click());
+
+  // a pure assertion has no side-effects; even if it fails, it can
+  // be run again and again without consequence
+  it('shows a loading indicator', () => {
+    expect($button.className).to.include('is-loading');
+  });
+});
+```
+
+Another common scenario is asserting that something **has not**
+happened. If you were to test for this normally (or even with a
+convergent assertion above) the test could potentially pass
+successfully before a side-effect has time to even happen.
+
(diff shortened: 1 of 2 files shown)


packages/mocha/README.md
```diff
@@ -1 +1,102 @@
...
+due to a preconcieved state not being met. For example, you can't
+click a button if it doesn't exist in the DOM. You may use
+`@bigtest/convergence` to converge on these states and return a
+promise in your hooks to wait for async tasks.
+
+``` javascript
+describe('clicking my button', () => {
+  // .run() returns a promise that will cause Mocha to wait for the
+  // convergence to resolve before continuing with the assertions
+  beforeEach(() => new Convergence()
```
wwilsman: This has me thinking, should we also wrap `beforeEach` to provide `this.convergence` which could be created with the current mocha timeout?

We could also make it so if you return a `Convergence` it will automatically call `.run()`.

packages/mocha/README.md
```diff
@@ -1 +1,102 @@
...
+    .do(() => $button.click())
+    .run());
+
+  it('shows a loading indicator', () => {
+    expect($button.className).to.include('is-loading');
+  });
+});
+```
+
+Check out the `@bigtest/convegence` [API
```
cowboyd: `convergence` is mispelled.

packages/mocha/README.md
```diff
@@ -1 +1,102 @@
...
+
+Convergent Mocha functions for testing against asyncronous states
+
+## What is this package?
+
+This package simply wraps Mocha's `it` in a convergence helper so
+that any assertions you write using `it` become convergent assertions
+that allow you to easily test asyncronous states.
+
+### Convergent Assertions?
```
cowboyd: I would lead with this. It explains the problem handily, and then we can talk about how it solves the problem. Also, remove the question mark.

packages/mocha/README.md
```diff
@@ -1 +1,102 @@
 # @bigtest/mocha
+
+Convergent Mocha functions for testing against asyncronous states
+
+## What is this package?
+
+This package simply wraps Mocha's `it` in a convergence helper so
```
cowboyd: ~simply~

packages/mocha/README.md
```diff
@@ -1 +1,102 @@
...
+
+### Convergent Assertions?
+
+Typically, when testing asyncronous states (such as rendered content
+in an application) your tests need to run at just the right moment so
+that they are executed in the correct context. If your tests run too
+soon, they will fail; if they run too slow, well then you just have
+slow tests.
+
+Converging on a state simple means asserting against said state until
```
cowboyd: ~simple~ s/asserting against said/asserting a state/

packages/mocha/README.md
cowboyd: I like the idea of getting your `run()` for free. Is there a way to detect the timeout of the current running test case globally so that you could just:

```js
import { Convergence } from '@bigtest/mocha'
```

And that constructor would detect the currently running testcase and use its timeout to delegate to the constructor of `@bigtest/convergence`

wwilsman:
Updated based on feedback. Also ran spellcheck and fixed a few mistakes.

packages/mocha/README.md
wwilsman: I don't think there is a way to get the current running test's timeout globally. Mocha has `mocha.timeout()` but it's use is for setting the _suite's_ timeout.

The only way I could think of would to update a global `Convergence.timeout` (or some other global) from `it`, which would then be delegated to the constructor of `@bigtest/convergence`.

cowboyd: It just occured to me that we could also just create a new convergence off of the old one, that uses the current timeout.

```js
function beforeEach(setup) {
  return mocha.beforeEach(() => {
    let result = setup.apply(this, arguments);
    if (result instanceof Convergence) {
      return result.timeout(this.timeout).run();
    } else {
      return result;
    }
  });  
}

File: packages/mocha/README.md [hunk 6]
```diff
@@ -1 +1,102 @@
...
+almost expired. Performing tests in this way allows them to pass the
+moment the desired state is achieved. This results in very fast tests
+when testing asynchronous things.
+
+_Read the `@bigtest/convergence` [docs on
+convergences](https://github.com/thefrontside/bigtest/tree/master/packages/convergence#why-convergence)
+for more info as to why converging on a desired state is better than
+trying to time it properly._
+
+## What is this package?
```
cowboyd: /What is this package?/How does it work?/

packages/mocha/README.md
```diff
@@ -1 +1,102 @@
...
+
+This package wraps Mocha's `it` in a convergence helper so that any
+assertions you write using `it` become convergent assertions that
+allow you to easily test asynchronous states.
+
+## Writing Tests
+
+Because convergent assertions are ran repeatedly until they pass, it
```
cowboyd: s/ran/run/

packages/mocha/README.md
```diff
@@ -1 +1,102 @@
...
+
+  // a pure assertion has no side-effects; even if it fails, it can
+  // be run again and again without consequence
+  it('shows a loading indicator', () => {
+    expect($button.className).to.include('is-loading');
+  });
+});
+```
+
+### Convergent Hooks?
```
cowboyd: If we want to phrase this as a question, then it has to connect directly to the reader's experience. E.g.

"What if I want to control the timing of my setup code?"

Otherwise, I'd just remove the question mark.

packages/mocha/README.md
```diff
@@ -1 +1,102 @@
...
+
+This package wraps Mocha's `it` in a convergence helper so that any
+assertions you write using `it` become convergent assertions that
+allow you to easily test asynchronous states.
+
+## Writing Tests
+
+Because convergent assertions are ran repeatedly until they pass, it
+is _highly recommended_ that you **do not** perform any side-effects
+in your assertions. This will result in the side-effect being ran
```
cowboyd: s/ran/run/

packages/mocha/README.md
```diff
@@ -1 +1,102 @@
...
+multiple times in the case when the assertion fails.
```
cowboyd: It isn't just when the assertion fails, if a side-effect takes 500ms to resolve, then we're still running the assertion 50 times. I'd say something like:

> This will result in your side-effect being run multiple, perhaps even hundreds of times.

packages/mocha/README.md
```diff
@@ -1 +1,102 @@
...
+when testing asynchronous things.
+
+_Read the `@bigtest/convergence` [docs on
+convergences](https://github.com/thefrontside/bigtest/tree/master/packages/convergence#why-convergence)
+for more info as to why converging on a desired state is better than
+trying to time it properly._
+
+## What is this package?
+
+This package wraps Mocha's `it` in a convergence helper so that any
```
cowboyd: Let's reiterate that this is very low touch difference from the way you write tests. Maybe start with

> For the most part, you write tests in the exact same way that you're used to writing tests with Mocha. The only difference is that this package wraps.....

wwilsman:
Second round of feedback integrated. I also realized I completely didn't include `it.always`. It's added now, so maybe one more round of feedback on that section?

cowboyd:
Looks great! The only remaining things that I can see to do are:

- [ ] documentation for `it.pause` (or is this too experimental at the moment)
- [ ] linking to `@bigtest/mocha` in the main README.

wwilsman:
I don't think `it.pause` is experimental, but I can't figure out a way to test it without making the suite hang. I added it to this README and added `@bigtest/mocha` to the repo README.

Issue #21, opened 2018-01-09
Autotag releases


cowboyd:
Whenever we merge a new version of a package to master, it automatically publishes that package to npm. This is 🔥 

To aid in reaching the branch points in the event that we have to work with older versions of the code, or that we need to make patch releases, we should _also_ have a `postpublish` script that creates the tag for the new package and pushes it back to the origin repo.

PR #24, opened 2018-01-10
Automatically tag new package versions

wwilsman:
#21 - ~Instead of tagging via each package's`postpublish` hook, tagging is done by the CI. Since the tags also need to be pushed by the CI, tagging can be it's responsibility. This also prevents us from needing to add `postpublish` hooks to every future package we add.~ See below

### TODO
- [x] Add GitHub keys to CircleCI so that `git push --tags` succeeds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,9 @@ jobs:
       - run:
           name: Publish to NPM
           command: ./bin/each-pkg --newer npm publish -- --access=public
+      - run:
+          name: Push Tags
+          command: git push --tags
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,7 @@ jobs:
           command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
       - run:
           name: Publish to NPM
-          command: ./bin/each-pkg --newer npm publish -- --access=public
+          command: ./bin/each-pkg --new npm publish -- --access=public
       - run:
           name: Push Tags
           command: git push --tags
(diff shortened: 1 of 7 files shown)


cowboyd:
What if we need to make a bugfix release off of a non-master branch (the use-case for tagging), would we have to tag the new release manually?

wwilsman:
Yes, since we would need to manually release it, it would need to be manually tagged. Perhaps we can change the `--newer` flag to mean "versions that don't exist"

wwilsman:
☝️ I did just that and changed `--newer` to `--new` to better convey that it means any unpublished
version of a package as opposed to versions newer than the latest.

Can I get a final approval?

wwilsman:
We just need to add those User GitHub keys so Circle can push tags. Do we have a bot account for this? Or shall I just use my own keys since we're just pushing tags?

cowboyd:
My biggest concern here is that if for every single instance where there is an NPM publication, there should also be a tag push, then it doesn't it make sense to make it a postpublish script that CI calls?

wwilsman:
Sure, that does make sense. We can add the `postpublish` script `../../bin/tag-pkg` and it'll work just fine. My concern was that we (or somebody else) will have to remember to add that hook to every package added in the future.

If you think that's fine, I'll remove the tag step from circle and add the `postpublish` scripts to our current packages.

cowboyd:
I think it's a decent tradeoff since new packages shouldn't be that common of an event.

cowboyd:
@wwilsman Are we ready to merge this? Looks like we've got mad approval.

wwilsman:
I was nervous about the github keys, so I was holding off merging until I had time to fix if something went wrong. Looks like all is good though!

Issue #22, opened 2018-01-09
Possibly remove `rollup-commonjs`


cowboyd:
According to @wwilsman, this may no longer be necessary in light of #18 #20

PR #25, opened 2018-01-10
Remove unnecessary commonjs plugin

wwilsman:
#22

### DIFF
--- a/packages/mocha/tests/it-test.js
+++ b/packages/mocha/tests/it-test.js
@@ -16,7 +16,7 @@ describe('BigTest Mocha: it', () => {
 
   it('throws on failure before the timeout', () => {
     expect(tests[1].duration).to.be.within(180, 200);
-    expect(tests[1].err).to.have.property('expected', 200);
+    expect(tests[1].err).to.have.property('expected', '200');
   });
 });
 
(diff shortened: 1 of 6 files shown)


packages/mocha/tests/it-test.js
```diff
@@ -16,7 +16,7 @@ describe('BigTest Mocha: it', () => {
 
   it('throws on failure before the timeout', () => {
     expect(tests[1].duration).to.be.within(180, 200);
-    expect(tests[1].err).to.have.property('expected', 200);
+    expect(tests[1].err).to.have.property('expected', '200');
```
Robdel12: [Hm](https://media.giphy.com/media/U8ZoygMEH3eBW/giphy.gif)

Robdel12:
So why is change needed? Is this so we don't need to use commonjs to consume our other packages? Or does this impact anyone who is consuming our packages?

wwilsman:
@Robdel12 [`rollup-plugin-commonjs`](https://github.com/rollup/rollup-plugin-commonjs) bundles your CommonJS dependencies in `node_modules`. Our only dependencies so far are `@bigtest/convergence` in `@bigtest/mocha` which now has ES module support with #20.

If we later need dependencies outside of this repo that do not support ES modules, then that package will require this plugin.

Robdel12:
Okay, so it's just an internal change and those who consume these packages on NPM will still get the `umd` output, right?

wwilsman:
Word

PR #26, opened 2018-01-10
Fix comment removal

wwilsman:
To remove comments from compiled files, `comments: false` needs to be at the root of babel's config.


Robdel12:
I like having a changelog. Would it be nicer to have something the generates it for you? I usually use https://github.com/skywinder/github-changelog-generator on a lot of FS projects.

It pulls from the GitHub API to build your changes. The tricky part is making sure you get the git history right or things could be missed.

PR #28, opened 2018-01-12
Convergent mocha hooks

wwilsman:
As per some [discussion in #23](https://github.com/thefrontside/bigtest/pull/23#discussion_r160558026), this wraps Mocha's hooks to automatically run returned convergences within the current timeout period.

### DIFF
--- a/packages/mocha/src/index.js
+++ b/packages/mocha/src/index.js
@@ -2,7 +2,7 @@ import * as mocha from 'mocha';
 import Convergence from '@bigtest/convergence';
 
 /**
- * Helper to create a convergent assertion using the current testing
+ * Creates a convergent assertion using the current testing
  * context's timeout.
  *
  * @param {Function} assertion - assertion to converge on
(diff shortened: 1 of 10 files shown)


wwilsman:
Changed to support running `Convergence`-like objects returned from the hooks. That is any returned object that supports both the `.timeout()` and `.run()` methods.

Also added some tests around preserving the existing functionality of the hooks when you return promises.

Finally, changed the README slightly to reflect the wrapped hooks and their ability to auto-run `Convergence` instances.

packages/mocha/src/index.js
```diff
@@ -28,6 +28,37 @@ function convergent(assertion, always) {
   };
 }
 
+/**
+ * Helper to allow automatically running returned objects that implement
```
cowboyd: ~Helper to~

We should also explain here why we want to do this, and provide some examples of what it enables. I.e. with out it, it means 

- annoying boilerplate in every beforeEach
- having to wrap all of the interaction methods with a call to `run()`

wwilsman:
@cowboyd Added an explanation for wrapping the hooks. One final approval?

Issue #27, opened 2018-01-10
Use ESLint


wwilsman:
I'm sure a few stylistic things have already slipped by. To make sure everything remains consistent across packages now and in the future, we should implement ESLint across the entire repo and choose a good starting point for our rules. 

I recommend [eslint-config-standard](https://github.com/standard/eslint-config-standard) as the starting point we can expand on, but I'm welcome to other suggestions as well.

Robdel12:
Would you y'all be interested in prettier too?

wwilsman:
I'd be down with prettier in conjunction with eslint. I'm a stickler for pretty code.

cowboyd:
@SamKeathley Want to give it a shot?

SamKeathley:
@cowboyd Sure!

wwilsman:
ESLint has now been implemented, but without prettier. @Robdel12 want to tackle adding prettier? If not, we can probably close this issue and another one can be opened for prettier if we still want that.

cowboyd:
I'm a prettier skeptic, but would nevertheless be 100% open to seeing what it does with our codebase. And 200% open to it if folks felt more productive with it.

Robdel12:
@cowboyd what has you on the fence?

I can throw a POC together tonight (or while Kaylie is in FL, I'll be living the bachelor life for the next week)

cowboyd:
Mainly it's the way it handles a polymorphic functional programming style (which is where I've been headed lately with funcadelic). We tried it on microstates.js for awhile but then ended up removing it. Here's why.

Essentially it's that this: 

```js
function analyzeStates(values) {
  return map(node => Object.create(node, {
    state: {
      get() {
        let { Type, value } = node;
        // also don't know if this is the way to calculate state.
        if (isPrimitive(Type)) {
          return value;
        } else {
          return Object.create(Type.prototype, propertiesOf(value));
        }
      }
    }
  }), values);
}
```

gets prettified to this:

```js
function analyzeStates(values) {
  return map(
    node =>
      Object.create(node, {
        state: {
          get() {
            let { Type, value } = node;
            // also don't know if this is the way to calculate state.
            if (isPrimitive(Type)) {
              return value;
            } else {
              return Object.create(Type.prototype, propertiesOf(value));
            }
          }
        }
      }),
    values
  );
}
```

In this instance, the original places the emphasis on the mapping operation.

```js
return map(node => Object.create(node, {
```
[...]

PR #29, opened 2018-01-25
Adding ESLint

SamKeathley:
Per issue #27 and some discussion there, add linting to project to fix some current styling problems and for stylistic consistency in the future. Eslint-config-standard has some good starting point configurations for us that we can add onto or change. I have also taken care of the current linting problems after getting all of this set up.

Eslint-config-standard offers some shared configs for us to start out with to get our code looking pretty. I also implemented the rollup-eslint-plugin since we are using rollup already and it claims to verify entry point and all imported files with ESLint.

####Learning
[Eslint-config-standard](https://github.com/standard/eslint-config-standard#eslint-config-standard----)
[Rollup-esling-plugin](https://github.com/TrySound/rollup-plugin-eslint#rollup-plugin-eslint-)

### DIFF
--- a/packages/convergence/package.json
+++ b/packages/convergence/package.json
@@ -19,6 +19,14 @@
     "chai-as-promised": "^7.1.1",
     "mocha": "^4.0.1",
     "rollup": "^0.53.0",
-    "rollup-plugin-babel": "^4.0.0-beta.0"
+    "rollup-plugin-babel": "^4.0.0-beta.0",
+    "rollup-plugin-eslint": "^4.0.0"
+  },
+  "dependencies": {
+    "eslint-config-standard": "^11.0.0-beta.0",
+    "eslint-plugin-import": "^2.8.0",
+    "eslint-plugin-node": "^5.2.1",
+    "eslint-plugin-promise": "^3.6.0",
+    "eslint-plugin-standard": "^3.0.1"
   }
 }
--- a/packages/convergence/src/converge-on.js
+++ b/packages/convergence/src/converge-on.js
@@ -23,7 +23,7 @@
  * @returns {Promise} resolves if the assertion passes at least once;
  * if `always` is true, then rejects at the first error instead
  */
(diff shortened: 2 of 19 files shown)


packages/convergence/package.json
```diff
@@ -19,6 +19,14 @@
     "chai-as-promised": "^7.1.1",
     "mocha": "^4.0.1",
     "rollup": "^0.53.0",
-    "rollup-plugin-babel": "^4.0.0-beta.0"
+    "rollup-plugin-babel": "^4.0.0-beta.0",
+    "rollup-plugin-eslint": "^4.0.0"
+  },
+  "dependencies": {
```
cowboyd: Are these `dependencies` or `devDependencies`?

packages/convergence/src/converge-on.js
```diff
@@ -23,7 +23,7 @@
  * @returns {Promise} resolves if the assertion passes at least once;
  * if `always` is true, then rejects at the first error instead
  */
-export default function convergeOn(assertion, timeout = 2000, always, useStats) {
+export default function convergeOn (assertion, timeout = 2000, always, useStats) {
```
cowboyd: I don't know how I feel about this rule. Seems like not the way most setups work?

packages/convergence/package.json
SamKeathley: ```devDependencies```

packages/convergence/src/converge-on.js
SamKeathley: I will change that.

packages/convergence/package.json
SamKeathley: They are ```devDependencies```

wwilsman:
Thanks for taking care of this!

I was thinking it'd be nice if we could do `yarn lint` and just get linting results without having to do the entire build. Also, we probably want to lint our tests as well, which don't get included with the rollup bundle. So maybe we don't need `rollup-eslint-plugin` after all?

Also, I noticed some qunit things slipped into this branch. 😉

PR #30, opened 2018-01-26
Add ESLint

SamKeathley:
Per issue #27 and some discussion there, add linting to project to fix some current styling problems and for stylistic consistency in the future. Eslint-config-standard has some good starting point configurations for us that we can add onto or change. I have also taken care of the current linting problems after getting all of this set up.

Eslint-config-standard offers some shared configs for us to start out with to get our code looking pretty.

Second iteration of this PR, hopefully all of those silly mistakes are fixed now! I did make a couple of added rule changes for the sake of our testing. I disabled the `no-return-assign` rule because it seems that this is how we really write our tests, and I had to disable `no-unused-expressions` rule because it just does not work with Chai.

#### Learning
[Eslint-config-standard](https://github.com/standard/eslint-config-standard#eslint-config-standard----)
[no-unused-expressions reading](https://github.com/eslint/eslint/issues/2102)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,7 +47,7 @@ jobs:
       - checkout
       - *restore_cache
       - run:
-          name: Link packages
+          name: Lint packages
           command: yarn lint
       - run:
           name: Build packages
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,9 @@ jobs:
       - checkout
       - *restore_cache
       - run:
+        name: Link packages
+        command: yarn lint
+      - run:
           name: Build packages
           command: yarn build
       - persist_to_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,8 +47,8 @@ jobs:
       - checkout
       - *restore_cache
       - run:
-        name: Link packages
-        command: yarn lint
+          name: Link packages
+          command: yarn lint
       - run:
           name: Build packages
           command: yarn build
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,6 @@
+{
+    "extends": "standard",
+    "rules": {
+        "semi": [2, "always"]
+    }
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,9 +1,12 @@
 {
-    "extends": "standard",
-    "rules": {
-        "semi": [2, "always"],
-        "space-before-function-paren": [2, "never"],
-        "no-unused-expressions": 0,
-        "no-return-assign": 0
-    }
+  "extends": "standard",
+  "rules": {
+    "semi": ["error", "always"],
+    "space-before-function-paren": ["error", {
+      "anonymous": "never",
+      "named": "never",
+      "asyncArrow": "always"
+      }],
+    "indent": ["error", 2]
+  }
 }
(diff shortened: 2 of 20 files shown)


packages/convergence/package.json
```diff
@@ -9,7 +9,8 @@
   "scripts": {
     "build": "rollup --config",
     "test": "mocha --opts ./tests/mocha.opts ./tests",
-    "postpublish": "../../bin/tag-pkg"
+    "postpublish": "../../bin/tag-pkg",
+    "lint": "./node_modules/.bin/eslint ./src/**.js ./tests/**.js"
```
wwilsman: We can remove the `./node_modules/.bin/` bit because when we run `yarn <command>`, yarn will look in that directory first for binaries.

We should maybe lint the whole package too, so `eslint ./` is probably what we want.

.eslintrc.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "standard",
```
wwilsman: We should use 2 spaces in this file since that's what we use elsewhere

packages/convergence/tests/convergence-test.js
```diff
@@ -160,7 +160,7 @@ describe('BigTest Convergence', () => {
         assertion = converge.once(() => expect(total).to.equal(5));
       });
 
-      it('resolves after assertions converge', async () => {
+      it('resolves after assertions converge', async() => {
```
wwilsman: If this rule is specific to `async`, we should disable it

packages/mocha/package.json
```diff
@@ -9,7 +9,8 @@
   "scripts": {
     "build": "rollup --config",
     "test": "mocha --opts ./tests/mocha.opts ./tests",
-    "postpublish": "../../bin/tag-pkg"
+    "postpublish": "../../bin/tag-pkg",
+    "lint": "./node_modules/.bin/eslint ./src/**.js ./tests/**.js"
```
wwilsman: Same note here about `./node_modules/.bin/` and linting `./` 😊

package.json
```diff
@@ -5,7 +5,8 @@
   "repository": "https://github.com/thefrontside/bigtest",
   "scripts": {
     "build": "bin/each-pkg yarn build",
-    "test": "bin/each-pkg yarn test"
+    "test": "bin/each-pkg yarn test",
+    "lint": "bin/each-pkg yarn lint"
```
wwilsman: Nice call adding this to the top-level `package.json`!

With this we can add a linting step to our CI. We'll probably want it to come before [this build step](https://github.com/thefrontside/bigtest/blob/master/.circleci/config.yml#L49-L51) of our `build` job.

wwilsman:
Looks awesome! Thanks for doing this!

.eslintrc.json
SamKeathley: Oh, yeah, my Atom tabs are a little messed up. I'll fix that haha.

.circleci/config.yml
```diff
@@ -46,6 +46,9 @@ jobs:
       - *add_ssh_keys
       - checkout
       - *restore_cache
+      - run:
+        name: Link packages
```
cowboyd: looks like we've got a small typo. Also, the indentation is not the same as the other `run` commands. While it looks like this works, let's align it with the other ones. This allowance for variance brings up a profound philosophical question: WHO WILL LINT THE LINTER!

PR #31, opened 2018-01-26
Import mirage

cowboyd:
The extraction of Mirage from ember-cli-mirage has been a stable companion on top of which we have been building production code for almost 8 months. We've been holding off publishing it as an NPM package in the hope that it will get merged upstream, or that we can get the upstream developers on board with supporting this extraction.

It looks like this is not very likely to happen any time soon, and so it's time to move it into bigtest, and release it to NPM so that users can install it from the public registry instead of from git.

This adds what was formerly `mirage-server` into the `packages/mirage` directory as another bigtest package. The history has been preserved from both mirage-server and ember-cli-mirage using the [tomono.sh][1] script so that all credit can be given where credit is due. And the amount of credit due is immense (as the commit history will attest).

I also had to add the `-browsers` image to our CI since mirage needs to be tested in the browser in order to test stubbing out `XMLHttpRequest`

## Followups

- [ ] migrate from building the test suite webpack to building with rollup.

## Further Learning

- Migrating from a single repo to a monorepo using [tomono.sh][1]
- Using [rollup to build your karma test suite][2]

[1]: https://github.com/unravelin/tomono
[2]: https://github.com/jlmakes/karma-rollup-preprocessor


Robdel12:
One of the things I really want to do is merge this upstream. It doesn't have to be the _next thing_ but I feel strongly that we should work with the ember-cli-mirage team to make this happen. Otherwise we're going to be in the same situation we were in with our prelude emacs. Always trying to merge from upstream to stay up to date.

cowboyd:
@Robdel12 I've talked quite a bit about this with @samselikoff, and I think that it's just not a priority for the `ember-cli-mirage` team at the moment, which 100% totally fine.

I still see re-integration as an ultimate goal, but at the same time, we can't continue to build production systems off of private forks. Also, if we're going to promote mirage to the wider JavaScript community, then we have to have it in a public place.

packages/mirage/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "rollup-plugin-babel": "^4.0.0-beta.0",
+    "rollup-plugin-commonjs": "^8.3.0",
+    "rollup-plugin-node-resolve": "^3.0.0",
+    "webpack": "^3.3.0"
+  },
+  "dependencies": {
+    "faker": "^4.1.0",
+    "inflected": "^2.0.2",
+    "lodash": "^4.17.4",
+    "pretender": "http://github.com/cowboyd/pretender#9ffecb8b789c6162e411ee8f023b5d635129a722"
```
wwilsman: What's the reason for this pointing to a sha of a fork?

wwilsman:
I fixed the rollup bundle. We needed the `resolve` plugin to resolve `node_modules` and we needed the `commonjs` plugin to convert commonjs modules to es modules.

Also, there's seems to be a lot of warnings and errors mucking up the console when the tests are run. Not a blocker, but something we might wanna look into since this repo definitely slows down our tests.

Another slow point is when the build process includes `lodash`. This is because there are two instances where we are including the entire `lodash` library instead of specific pieces.

packages/mirage/package.json
cowboyd: It was because pretender was not compatible with Webpack builds https://github.com/pretenderjs/pretender/pull/209

But this has since been fixed https://github.com/pretenderjs/pretender/pull/216

And I'm not even sure if this is going to be an issue now that we're building with rollup, but in either case, it's been fixed upstream, so we should be able to just use Pretender >  2.0  https://github.com/pretenderjs/pretender/blob/v2.0.0/CHANGELOG.md#20

cowboyd:
I wonder if it's something other than `lodash`. I pulled apart all of the monolithic `lodash` dependencies, and replaced them with individual imports, but the rollup is still 1.5M

🤔

wwilsman:
You're right. It is faker causing the large bundle size. 

image

Should we just list faker as a peer dependency to cut down our bundle size? It looks like the only time it's actually used is for adding the `faker.list` and `faker.random.number.range` helpers.

wwilsman:
After making faker a peer dependency:

image

After breaking down the lodash imports:

image

If you agree we should make faker a peer dependency, I can push this commit up

Robdel12:
What are the tradeoffs for peer dependencies? And if they're not bad why not make pretender and its [dependencies peer dependencies?](https://user-images.githubusercontent.com/2072894/35773456-deda7458-0918-11e8-8656-bcb85b65fd8e.png)

wwilsman:
I think the distinction is pretender is required for mirage to work, while faker is not.

cowboyd:
Yeah, it's tough because you almost certainly will end up using faker, and then we haven't really saved them anything because their test bundle will still include faker. I'm trying to think of a case where we've used mirage and also not used faker.

Alternatively, we could just remove it as a dependency completely, and then, if you want to use faker in your factories, you'll have to add it as a dev dependency. We don't do any kind of wrapping of faker, so really it's your app's factories that have the dependency on faker, not mirage.

Although now that I'm thinking about it, what if we take a "if it ain't broke, don't fix it." approach on this one, and leave the dependency as-is?

wwilsman:
If we remove it as a dependency, then we should remove the `list` and `random.number.range` helpers. (I don't think we should do this)

If we leave it as is, then we increase the build times. Not a big deal, but that's where this concern originated from.

While faker works with mirage factories, you can definitely create factories without it. Or you can not use factories at all and just use fixtures. We can list it as an `optionalDependency` and add some conditional logic to the `faker.js` file in this package. This way if you want faker, you can add it to dev dependencies and you'll also get the two helpers this package provides.

In the case of optional dependency, we should add that to the readme concerning factories.

Robdel12:
Wait, it just occurred to me. Why are we splitting ends on test bundle size?

wwilsman:
@Robdel12 

> If we leave it as is, then we increase the build times. Not a big deal, but that's where this concern originated from.

😉

cowboyd:
Looks like it adds about 50s to the overall build https://circleci.com/gh/thefrontside/bigtest/343?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Personally, I'm ok with this. If it's a problem, perhaps we can revisit later.

Robdel12:
🤔 So that's the parse cost we're paying on a 1.5mb lib. Woah.

PR #43, opened 2018-02-05
Update root yarn.lock

cowboyd:
These dependencies were out of date and more importantly, failing to kick the yarn cache on our CI, to let it know that a re-install is necessary.

This updates the global yarn.lock to point to the most recent versions of pretender and fake-xml-http-request, which is technically more correct, but also has desired side-effect of killing the cache


wwilsman:
[Here's the relevant section of our `.circleci/config.yaml`](https://github.com/thefrontside/bigtest/blob/master/.circleci/config.yml#L41)

PR #50, opened 2018-02-20
Default mocha `it.always` timeout to 100ms

wwilsman:
## Purpose

When using `it.always` the mocha default timeout of 2000ms is used. This makes testing with `it.always` much slower.

## Approach

Mocha `Runnable` instances have a chainable method interface. So returning `test.timeout(50)` allows us to have a default timeout of `50ms` and still allow changing the timeout with `it.always(...).timeout(100)`.

### Open Questions

- Is 50ms a good default timeout for `it.always`? Or should we go a little higher to say 100ms? At 50ms, the test will run at least 2-3 times (because it aborts about ~20ms before the timeout)

### DIFF
--- a/packages/mocha/src/index.js
+++ b/packages/mocha/src/index.js
@@ -93,7 +93,7 @@ function it(title, assertion) {
  */
 function itAlways(title, assertion) {
   if (!assertion) return mocha.it.skip(title);
-  return mocha.it(title, convergent(assertion, true));
+  return mocha.it(title, convergent(assertion, true)).timeout(100);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


wwilsman:
I think a single animation frame is about 17ms (1000/60fps). So technically 50ms is within the range. But I like 100ms so it will run the test at least 7-8 times

packages/mocha/src/index.js
```diff
@@ -93,7 +93,7 @@ function it(title, assertion) {
  */
 function itAlways(title, assertion) {
   if (!assertion) return mocha.it.skip(title);
-  return mocha.it(title, convergent(assertion, true));
+  return mocha.it(title, convergent(assertion, true)).timeout(100);
```
Robdel12: Will the timeout ever need to be configured?

cowboyd:
derp. I was getting 60fps confused with 60ms

PR #54, opened 2018-03-02
Redefine convergence timing

wwilsman:
## Purpose

After using `@bigtest/mocha` in  real projects, it seems as though `it.always` [can sometimes be flakey](https://travis-ci.org/folio-org/ui-eholdings/builds/346465220).

After digging into this issue a little bit, I found that the problem is with how we attempt to resolve or reject the converging promise _before the timeout_.

When we created `@bigtest/convergence`, the goal was to help mitigate waiting for the right time to run our tests. What we ended up doing in our convergence is trying to _wait for the right time to settle the promise_. So unknowingly, we ended up creating a very similar problem.

As seen in [these tests](https://circleci.com/gh/thefrontside/bigtest/492), when a convergent assertion has some latency associated with it's execution, it is resolved or rejected _after the timeout_. In `@bigtest/mocha` this results in the convergences not settling in time for mocha's own timeout and causes painful mocha timeout errors.

This issue is not just with `always` convergences. I found that there are situations in which a normal convergence will _successfully_ converge _after_ the timeout. Both situations can be explained by walking through the final loops of a convergence given an assertion with some latency in execution.

Given a `100ms` timeout:
- `79ms` has elapsed within convergence (before the final loop)
- `+6ms` latency for assertion to complete
- `+10ms` wait before looping again (setTimeout)
- `95ms` have elapsed; convergence will settle during this loop
- `+6ms` latency for assertion to complete
- `101ms` - convergence settles, **but not within the timeout**

Even if we just add a rejection after the timeout, that will still cause tests to show cryptic timeout errors when the convergence exceeds the test's timeout.

## Solution

Redefine convergence timing **so that convergences settle as soon as they can**. This means we don't have to attempt to settle one or two intervals before the timeout. We can just settle as soon as the timeout has been exceeded.

For typical convergences, they will still resolve before the timeout, but at any time after the timeout they will reject; _even if the assertion has finally passed_, if it happens after the timeout, the convergence has failed.

For `always` convergences, they will reject before the timeout, and always resolve after the timeout (as long as the assertion passes during every interval).

### Todo / Open Questions

- `@bigtest/convergence` only increments the minor version because this package is not yet at a `1.0.0` release. If we should release this as `1.0.0`, I can amend that commit before merging.

- I locked `@bigtest/mocha` at `@bigtest/convergence: "0.3.0"` so that it still currently works as expected. I also incremented the `@bigtest/mocha` patch version so that dependent projects will not break (as long as they're not locked to a previous version). That does mean that this single PR will release two of our packages at the same time (unless we bump up `@bigtest/convergence` to `1.0.0`)

- After this is released, `@bigtest/mocha` should be updated with this new convergence behavior. I think we could introduce a `latency` to ensure convergence timeouts are at least `x` ms quicker than mocha timeouts. This latency could be customized per project and increased in CI environments where things typically take a bit longer to execute.

- Due to the dependency lock within `@bigtest/mocha`, this is the first time dependencies of other packages within this repo are not linked. As such, it revealed that our module entry points are not set up correctly:

  It is common in a webpack environment when using babel to not transform modules within the `node_modules` directory. When `@bigtest/convergence` is not linked, this causes tests within `@bigtest/interaction` to fail due to transpilation errors. We need to define builds for our packages that compile true es-modules instead of just referencing `src` in our packages' `module` fields.

### DIFF
--- a/packages/convergence/CHANGELOG.md
+++ b/packages/convergence/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Changed
+
+- convergent assertions no longer attempt to resolve or reject before
+  their timeout
+
 ## [0.3.0] - 2018-02-18
 
 ### Added
--- a/packages/convergence/CHANGELOG.md
+++ b/packages/convergence/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+## [0.4.0] - 2018-03-02
+
 ### Changed
 
 - convergent assertions no longer attempt to resolve or reject before
--- a/packages/convergence/README.md
+++ b/packages/convergence/README.md
@@ -47,8 +47,7 @@ longer returns false or throws an error. When the function is finally
 successfully executed, it is considered to be passing and a converging
 promise will resolve. However, if the converging function does not pass
 within the provided timeout period the promise will reject with the
-last error thrown from the function. A converging promise will reject
-right before the timeout so it is never reached during it's execution.
+last error thrown from the function.
 
 ### Using `Convergence`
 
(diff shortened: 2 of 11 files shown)


cowboyd:
This is outstanding work @wwilsman.

I'm sold, and I think the fact that the API surface of `convergeOn` actually shrank means that it's conceptually much simpler.

> After this is released, @bigtest/mocha should be updated with this new convergence behavior. I think we could introduce a latency to ensure convergence timeouts are at least x ms quicker than mocha timeouts. This latency could be customized per project and increased in CI environments where things typically take a bit longer to execute.

Can `@bigtest/mocha` be "smart" about this, and choose a default latency that needn't be tweaked unless necessary?

> When @bigtest/convergence is not linked, this causes tests within @bigtest/interaction to fail due to transpilation errors. 

We should create an issue for this.

PR #57, opened 2018-03-06
Refactor @bigtest/mocha to account for convergence latency

wwilsman:
## Purpose

With the [latest `@bigtest/convergence` changes](https://github.com/thefrontside/bigtest/pull/54), convergences no longer attempt to settle _before_ the timeout. They now settle as soon as the can: when a normal convergence fails, it will fail _just after_ the timeout; when an always convergence succeeds, it will do so _just after_ the timeout. With this, failures in `@bigtest/mocha` (or whenever `it.always` is used) will cause tests to exceed the mocha timeout and throw hard-to-track timeout errors, sometimes wrongly so.

~By introducing a `latency` we can give mocha a little more time before throwing it's dreaded timeout error.~

By setting the mocha timeout to `0` when running convergences, we can allow the convergence to settle just after it's own timeout without experiencing mocha timeout errors.

## Approach

- ~When using `describe` exported by `@bigtest/mocha`, all hooks and tests registered with the suite will have a `.latency()` method added to them. This method works similarly to mocha's own `.timeout()` method in that calling `.latency()` without any arguments will return the current latency, and can also be chained from the timeout method: `test(...).timeout(200).latency(50)`.~

- ~When hooks auto-run convergences, or when convergent assertions are run, the current hook/test will get it's timeout extended by the `latency` amount. So given the test above, the final timeout would be `250ms`.~

  ~**Why not shorten the convergence timeouts by instead?** - Hopefully, we won't get mocha timeout errors anymore. But if we did, and were to increase the latency, we could face an issue where the convergences are only given a _very very_ short time to complete, and they will stop being reliable. By increasing the test's final timeout, we can add as large of a latency as necessary (hopefully not though) without affecting our convergence's ability to be reliable.~

- When a convergence is run, it uses the current mocha timeout. The convergence will always settle, so right before running it the mocha timeout is set to `0` to tell mocha not to worry about timeouts at all.

- After adding a third function to the top of the `index.js`, I decided they should be moved to their own `utils.js` file. I dried up the rest of the index file too by using "creator" functions to bind the variations of `mocha.it`, `mocha.before`, etc. to all of the convergent counterparts.

- All of the previous tests have been adjusted to test for the new convergent timeouts.


Robdel12:
I added `WIP` to the title so this can't accidentally get merged without updating the README

wwilsman:
@Robdel12 We [can't really](https://github.com/yarnpkg/yarn/issues/4725) try it out until it's released. One of the downsides of a monorepo. 

(Should bigtest have it's own org?)

wwilsman:
Can I get another review of the readme addition?

cowboyd:
I'm wondering if there isn't a way to subsume the mocha timeout period completely. In other words. If we are using convergent assertions, then with the recent changes, we _know_ with certainty that a convergent assertion will settle at some point, so can we dynamically fiddle with the mocha timeout so that it just follows the convergent assertions timeout?

cowboyd:
For example: set the mocha timeout to `0` to start (no timeout), and then invoke `done` according to the timing of the convergence.

wwilsman:
@cowboyd you're absolutely right. We could just always set mocha's timeout to `0` when running a convergence. We already return the promise from the convergence, so we wouldn't have to even invoke `done`.

Makes the whole `latency()` stuff not matter, but it's much more simple that way. I'll start making that change  👍

cowboyd:
Awesome! 

Goodbye mocha timeout errors. We won't miss you 👋 
  - hopefully us soon

wwilsman:
Just made that change and updated the PR description (without removing the history).

A final review, please?

PR #60, opened 2018-03-09
Add full support for returning promises in convergence `.do()` blocks

wwilsman:
## Purpose

This was previously, loosely, supported due to the currying nature of `.do()` blocks within a promise. This makes stats for these blocks actually report the true times and return values of the promises.

## Approach

When the result of a `.do()` block is thenable, the stats are calculated when it resolves. Otherwise, they are calculated right away.

If a promise takes longer than the remaining timeout, an error will be thrown after the promise resolves. This same error will now also be thrown before assertions if the elapsed time exceeds the allotted timeout. Before, if a `.do()` block made the convergence exceed the timeout, the following convergent assertion could be called with a negative timeout. That is no longer possible when throwing the new timeout error.

### DIFF
--- a/packages/convergence/CHANGELOG.md
+++ b/packages/convergence/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Added
+
+- full support for returning promises in `.do()` blocks
+- timeout error thrown when convergence exceeds timeout
+
 ## [0.4.0] - 2018-03-02
 
 ### Changed
--- a/packages/convergence/src/convergence.js
+++ b/packages/convergence/src/convergence.js
@@ -196,14 +196,25 @@ export default class Convergence {
       stats.stack.push(newStats);
     };
 
+    // throws when the elapsed time exceeds the timeout
+    let getElapsedSince = (start) => {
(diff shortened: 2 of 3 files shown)


cowboyd:
This feels like we're brushing up against `flatMap`. Two questions:

1. What's the use-case?
2. What do you tihnk _should_  happen if you return a new `Convergence`?

wwilsman:
1. It worked before, this just adds the proper stats tracking. A use case I could see would be performing some other async task in the middle of a convergence / interaction.

2. I meant to handle this as well, but thought that could be another PR. If a convergence is returned, we could run that convergence with the current timeout. Similar to how we return `convergeOn(assertion, timeout, always)`, if a new `Convergence` is returned we can just return `convergence.timeout(timeout).run()`.

wwilsman:
#67 handles returning a convergence

Robdel12:
@wwilsman is this the PR that will help with the test fails I'm seeing in https://github.com/folio-org/ui-eholdings/pull/263?

wwilsman:
@Robdel12 Sort of, yes. This PR adds an error that is throw when a convergence exceeds the timeout. So you wouldn't see negative timeout errors, but you might still see the new exceeded timeout error. But after this error you will be able to see what the actual current timeout is and adjust accordingly.

Maybe we need to think about including information about which convergence was run prior to exceeding the timeout.

PR #61, opened 2018-03-09
Fix @bigtest/interaction isPresent page property

### LINKED ISSUE #58: Using `isPresent` without a selector
When omitting `selector` using any of the page object properties, the property uses the page object's root element instead. `isPresent` without the `selector` will throw an error when the root does not exist. It should catch this error and return `false` if either the root or the selector within the root does not exist.

wwilsman:
## Purpose

Fixes #58.

When the `selector` is not provided, the property should work with the page's root element. For `isPresent`, it throws an error when the root does not exist when it should return `false`.

## Approach

Wrap the return statement in a try-catch block that returns `false` when an error is thrown.

Also upgraded the `@bigtest/convergence` dependency to bring in the new convergence timing.

### DIFF
--- a/packages/interaction/src/properties/is-present.js
+++ b/packages/interaction/src/properties/is-present.js
@@ -9,6 +9,11 @@ import { computed } from './helpers';
  */
 export default function(selector) {
   return computed(function() {
-    return !!this.$$(selector).length;
+    try {
+      // throws if the root element cannot be found
+      return !!this.$$(selector).length;
+    } catch (e) {
+      return false;
+    }
   });
 }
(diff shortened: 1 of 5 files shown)


packages/interaction/src/properties/is-present.js
```diff
@@ -9,6 +9,11 @@ import { computed } from './helpers';
  */
 export default function(selector) {
   return computed(function() {
-    return !!this.$$(selector).length;
+    try {
+      // throws if the root element cannot be found
+      return !!this.$$(selector).length;
+    } catch (e) {
```
Robdel12: Does eslint complain about an unused var?

wwilsman: Not for `catch` because the `error` var is required

Robdel12: TIL. I remember something like that..

Issue #62, opened 2018-03-09
Support async/await syntax


wwilsman:
Based on the popularity of async/await usage in tests, supporting this without `.run()` would reduce the boilerplate required to use this pattern with `@bigtest/interaction` page objects.

**With `.run()`:**
``` javascript
it('does things', async () => {
  await page.fillName('name namerson').run();

  expect(page.name).to.equal('name namerson');

  await page.clickSubmit().run();

  expect(page.hasSuccessMessage).to.be.true;
});
```

**Proposed:**
``` javascript
it('does things', async () => {
  await page.fillName('name namerson');

  expect(page.name).to.equal('name namerson');

  await page.clickSubmit();

  expect(page.hasSuccessMessage).to.be.true;
});
```

wwilsman:
We could also maybe support auto-wrapping getters in a convergence in this case. That would enable us to use convergences with the same async/await syntax.

``` javascript
it('would require chai-as-promised', async () => {
  await page.fillName('name namerson');
  await expect(page.name).to.eventually.equal('name namerson');
  await page.clickSubmit();
  await expect(page.hasSuccessMessage).to.eventually.be.true;
});
```

Robdel12:
Closed with #77

PR #77, opened 2018-03-16
Make Convergences & Interactions thennable

wwilsman:
## Purpose

This will add support for the `async`/`await` syntax with convergences without needing to call `.run()`.

## Approach

A `.then()` method is all that is necessary to support this syntax. By always resolving with the value of the last function in the stack, we can allow convergences to resolve to this value when using `async`/`await`.

``` javascript
async function clickElement(selector) {
  // will resolve to the element when it exists
  let $node = await new Convergence().once(() => {
    let el = document.querySelector('.element');
    return !!el && el;
  });

  $node.click();
}
```

I went ahead and upgraded `@bigtest/interaction` to make use of this as well. So this PR will release two of our packages.

### DIFF
--- a/packages/convergence/CHANGELOG.md
+++ b/packages/convergence/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+## [0.6.0] - 2018-03-16
+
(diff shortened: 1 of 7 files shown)


Robdel12:
Looks like this will take care of #62, right?

wwilsman:
Yep. I updated that issue as well.

Issue #72, opened 2018-03-16
Deprecate `once` in favor of `when`


Robdel12:
We like the API of `when` better than once. So let's deprecate `once` & remove it in v1.0.0

When we tackle this we should answer: does `always ` need to be renamed too?

PR #84, opened 2018-03-23
Change convergence `.once()` to `.when()`

wwilsman:
## Purpose

Addresses #72 

`.when()` reads better in most situations. 

## Approach

Search-and-replaced pretty much every instance of `once` that made sense.

To avoid breaking changes, `.once()` was aliased to `.when()`.

### DIFF
--- a/packages/convergence/src/convergence.js
+++ b/packages/convergence/src/convergence.js
@@ -33,7 +33,7 @@ import {
  *
  * Example:
  *   let first = new Convergence(100)
- *     .once(() => expect(foo).to.equal('bar'))
+ *     .when(() => expect(foo).to.equal('bar'))
  *
  *   let second = first.timeout(200)
  *     .do(() => console.log('foo', foo))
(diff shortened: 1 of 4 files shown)


wwilsman:
This is where a changelog bot would be helpful to remind me to add to the changelog...

packages/convergence/src/convergence.js
```diff
@@ -92,21 +92,31 @@ class Convergence {
...
 
   /**
-   * Similar to `.once()`, creates a new convergence with the given
+   * Alias for `.when()`
+   *
+   * @deprecated
+   * @returns {Convergence} a new convergence instance
+   */
+  once() {
+    return this.when(...arguments);
```
Robdel12: Can we add a `console.warn` here so if anyone is using it they'll know to migrate?

wwilsman: > I didn't add a deprecation warning message to .once() since this is pre-v1 anyway.

Plus, if we log a warning, test runners will get slow when logging a thousand of them. I think not warning is okay since this is all alpha / beta currently

cowboyd: I think if you're an early adopter you're going to want to be made aware of the churn ASAP, because early adopters are more tolerant of churn in order to get the hawt features. It'll make their lives easier. And, after all, you're _upgrading_ the version of the framework. 

Maybe having an easy way to silence the deprecations?

wwilsman: What would be an easy way to silence it? It is just one warn, so I don't think we need to introduce a deprecation mechanism for it.

PR #88, opened 2018-03-31
Convergence Instance Context

wwilsman:
## Purpose

While working on `@bigtest/interaction`, I realized that this is currently not possible with convergences:

``` javascript
new Convergence().do(function() {
  console.log(this.timeout())
})
```

It might not seem useful in the context of convergences, so here's an example using `@bigtest/interaction`:

``` javascript
// scrolls to the bottom of the page when there is room to scroll
await new Interactor('#page')
  .when(function() {
    return this.$root.scrollHeight > this.$root.offsetHeight
  })
  .do(function() {
    return this.scroll({ top: this.$root.scrollHeight })
  })
```

## Approach

First, bind the instance context to any `assertion`s and `callback`s created using `#when()`, `#always()`, `#do()`, etc.

Previously, `convergeOn` curried its own context onto the assertion ~and always overrode any context bound to the assertion. This was removed to prevent it from overriding the Convergence instance's bound context.~ Turns out this isn't true, but since `convergeOn` is meant to be internal and this context currying isn't used by `Convergence` I don't think it is necessary to add back.

Since we now have to pass the context along to the `runAssertion` and `runCallback` helpers, I reduced their API by moving the `last` argument into the `subject` that gets passed along.

### DIFF
--- a/packages/convergence/CHANGELOG.md
+++ b/packages/convergence/CHANGELOG.md
@@ -13,10 +13,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 ### Changed
 
 - updated inline api docs
+- `Convergence` assertions retain the instance context
 
 ### Deprecated
 
 - `.once()` in favor of `.when()`
+
+### Removed
+
+- context currying from `convergeOn`
 
 ## [0.6.0] - 2018-03-16
 
--- a/packages/convergence/src/converge-on.js
+++ b/packages/convergence/src/converge-on.js
@@ -24,7 +24,6 @@
  * if `always` is true, then rejects at the first error instead
  */
 export default function convergeOn(assertion, timeout = 2000, always = false) {
-  let context = this;
   let start = Date.now();
   let interval = 10;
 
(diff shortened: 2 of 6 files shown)


packages/convergence/CHANGELOG.md
```diff
@@ -13,11 +13,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 - updated inline api docs
+- `Convergence` assertions retain the instance context
 
 ### Deprecated
 
 - `.once()` in favor of `.when()`
 
+### Removed
+
+- context currying from `convergeOn`
```
Robdel12: Is this a breaking change of any sort? I'm trying to think of a way you'd use this while testing..

wwilsman: It used to be used by `@bigtest/mocha` via `convergeOn.call(this, assertion)`. 

`convergeOn`, however, is meant to be internal. I think by v1 we should stop exporting it.

Robdel12: Yea, I agree. We should open an issue to track that.

I bet we could get away with this breaking change. _oof famous last words_

Issue #81, opened 2018-03-23
Let's remove page objects from @bigtest/interaction...


wwilsman:
`@bigtest/interaction` is named after the `Interaction` class, but it's primary usage is with page objects. The page objects themselves are just thin wrappers around interactions. They only add property getters that may throw on certain conditions, otherwise their methods just get forwarded to the interaction. The term "page object" is actually a misnomer within our library because we can make much smaller, composable, component-based, interaction objects... in fact we should just call them that: _**Interaction Objects.**_ 

Although, do we need the thin layer objects to begin with? `@interaction` (formerly `@page`) already  builds custom `Interaction` subclasses. These custom subclasses are returned when page-object property methods are called. We would just need to move the other getter properties down to the custom interaction subclasses, and page-objects can be removed completely.

If we do add the property getters to interactions, that adds a potential pitfall:
```javascript
new CustomInteraction()
  .fillName('name')
  .click('.submit')
  .someProperty 
// ^ this is just a getter that returns a value;
// the above interactions never actually happen.
```

However, this actually helps when composing interactions, and can be converged on using the interaction's own convergence methods
```javascript
// when composing interactions ...
let interaction = new CustomInteraction()
  .fillName('name')
  .click('.submit')

// ... this works ...
expect(interaction.someProperty).to.equal('something')

// ... and you can converge on values via convergence methods
let afterProp = await interaction
  .do(() => interaction.someProperty)

// or write the assertions within the interaction convergence
new CustomInteraction()
  .fillName('name')
  .click('.submit')
  .once(function() {
    expect(this.someProperty).to.equal('something')
  })
```

#### What does this mean for existing page objects?
Well, _nothing_. Page objects are such a thin wrapper that by just moving the getters down to the underlying interaction, the API stays exactly the same. We would just alias `@page` to `@interaction`, deprecate `@page`, and remove it for the v1 release.

_Technically_, any use of `page.interaction` would need to be replaced with just `page` since the interaction would no longer be wrapped. But that usage is pretty minimal.

cowboyd:
I agree 100% that PageObject is a complete misnomer.

Although I'm still confused on terminology. Who is responsible for scoping to the DOM? Because to me, the word "interaction" feels both atomic and abstract. Like a click, or touch, or something composed of touches like a swipe. 

Although when I actually read it, I like the sound of 

```js
@interaction class CardStackInteraction {
  cards = collection(class CardInteraction {
    title = focusable('input[name="title"]');
[...]

Issue #80, opened 2018-03-21
Group common property helpers


wwilsman:
While working with interaction page objects, I'm seeing stuff similar to this a lot:

```javascript
hasField = isPresent('input[data-test-field]');
focusField = focusable('input[data-test-field]');
fillField = fillable('input[data-test-field]');
blurField = blurable('input[data-test-field]');
```

What if we simplified that with a single property helper?

```javascript
field = input('input[data-test-field]');
```

That has a predetermined set of properties...

```javascript
page
  .field.focus()
  .field.fill('value')
  .field.blur()
```

We could do something similar with other properties as well.

```javascript
error = element('[data-test-error]');
```

...

```javascript
expect(page.error.exists).to.be.true
expect(page.error.text).to.equal('some error')
```

And maybe even a way to compose these common properties?

```javascript
field = element('[data-some-thing]').focusable().blurable().fillable()
```

Robdel12:
Just sayin' `field = input('input[data-test-field]');` is an excellent place to do #73

wwilsman:
I thought of that too. We could make use of composable convergences for these. An `input` convergence would do proper input validations, then the `fill`, `focus`, `blur`, etc, convergences are just chained from that initial one.

PR #89, opened 2018-04-02
Remove @bigtest/interaction page-object

wwilsman:
## Purpose

This is meant to address #80 & #81. 

TL;DR: 

Page objects are misnomers within `@bigtest/interaction` and aren't really needed since they are just thin wrappers around the `Interaction` class.

- `Interaction` -> `Interactor`
- `@page` -> `@interactor`
- Added nested interactors
- Made `collection` lazy
- Reordered property creator arguments

## Approach

This actually started out as an implementation of #80. To achieve a syntax where nested interaction methods return an instance of the top-most interaction, a `parent` option was added to the interaction constructor. When a new interaction inherits from a previous instance with a parent, it appends to the parent instance and returns the resulting interaction.

``` javascript
page.input.fill('value')
//=> page.append(input.fill('value'))
``` 

This allows chaining nested _interactions_ like so:

``` javascript
page
  .input.fill('value')
  .input.blur()
  .submit()
```

While adding this parent reference to the decorator, I realized that I would be dealing with nested _page-objects_ that had references to their _own interactions_. And that was a lot more complicated than just nesting interactions. Since page-objects were on the verge of being removed, I lumped #80 & #81 together.

---

To continue with the work on #80, the `collection` property could also use this new parent option:

``` javascript
checkboxes
  .items(3).check()
  .items(6).check()
```

However, the `collection` method would throw an error if the element didn't exist at the index during invokation

``` javascript
// throws because it cannot find the 99th item
list.items(99)
```

To prevent an error from being thrown immediately, the `scope` was changed to allow a function that will be evaluated whenever the `$scope` getter is accessed. This allowed the collection method to be lazy and not look for the item until a getter, or `run` was called.

``` javascript
// converges on the 99th item existing
list.items(99).click()

// throws because it cannot find the 99th item
list.items(99).text
```

---

Coming back around to #81: 

The page-object helper methods were moved to the interaction class, and that was renamed to `Interactor`. 

The `$scope` getter was renamed to `$root`, but `scope` is still the name of the option given to the constructor which can be a selector, element, or function that returns a selector or element.

The `properties` directory was renamed to `interactions` to better convey that an interactor is comprised of interactions (and other interactors).

While refactoring some of the property helpers to reference `this.$root` instead of `this.$scope`, I realized that they are now no longer page-object properties, but interactor properties. So the arguments given to the respective methods should match the arguments given to the property creators. **This is a major breaking change**

``` javascript
// before
page.trigger([selector], eventName, [options])
triggerable(eventName, [selector], [options])

// after
page.trigger([selector], eventName, [options])
triggerable([selector], eventName, [options])
```

Previously, interaction methods already had an optional selector as the first argument, so property creators were changed to meet this requirement as well

``` javascript
// before
attribute(attr, [selector])
property(prop, [selector])

// after
attribute([selector], attr)
property([selector], prop)
```

Some property getters were added as default interactor properties just like there are already default methods.

``` javascript
// works out-of-the-box
new Interactor('input').value
new Interactor('p').text
new Interactor('div').isPresent
new Interactor('div').isVisible
new Interactor('div').isHidden
```

---

Of course, tests were updated / added for everything as well.

While I was at it, I added proper documentation comments for everything in `@bigtest/interaction`.

I'm not sure why, but the diff didn't catch the file renames, so every file I renamed shows up as a delete and addition. That explains the "101 files changed." Might be best just to view this one without the diff in the way since it probably isn't useful anyhow.


Robdel12:
Holy crap, lots to take in here. 

First reaction is this is dope. 🎉  Second reaction is how do we handle this? Go v1 here? Seems like there's a lot of breaking changes in here.

wwilsman:
I'd say the next release is a good candidate for v1  😃 

However, `@bigtest/convergence` would still be pre-v1, and I'm not sure how I feel about this package being v1 with it's only dependency pre-v1. Would we want to release a v1 of that package as well?

PR #92, opened 2018-04-07
@bigtest/convergence v0.7.0 & @bigtest/interaction v0.4.0

wwilsman:
This releases `@bigtest/convergence` 0.7.0 and `@bigtest/interaction` 0.4.0 at the same time due to test failures when trying to release `@bigtest/convergence` 0.7.0 alone (#91).

This highlights some potential pitfalls of a monorepo. Since the dependency in `@bigtest/interaction` was automatically linked, recent changes utilized an unreleased feature of `@bigtes/convergence`. When changing the version number of `@bigtest/convergence`, the link was broken and `@bigtest/interaction` used the version from npm instead, which did not support `#when()`. Failures can be seen for that release [here](https://circleci.com/gh/thefrontside/bigtest/773).

Released packages are not effected since the version numbers are correct, but this requires us to release these two packages at once to avoid CI/CD failures.

### DIFF
--- a/packages/convergence/CHANGELOG.md
+++ b/packages/convergence/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+## [0.7.0] - 2018-04-07
+
 ### Added
 
 - `docs` script to generate documentation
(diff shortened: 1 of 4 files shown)


cowboyd:
Yeah, I'm glad we got to try the monorepo thing, and now we can have an intelligent opinion about the tradeoffs.

PR #109, opened 2020-02-04
RFC: New Interactor API

pittst3r:
[Rendered](https://github.com/thefrontside/bigtest/blob/interactor-rfc/rfcs/0001-new-interactor-api.md)

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "proseWrap": "always"
+}
--- a/rfcs/0000-template.md
+++ b/rfcs/0000-template.md
@@ -1,0 +1,69 @@
+---
+Start Date: (fill me in with today's date, YYYY-MM-DD)
+RFC PR: (leave this empty)
+GitHub Issue: (leave this empty)
+---
+
+## Summary
+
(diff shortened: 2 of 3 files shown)


jnicklas:
So I was reading this again, to see if we can close it out. Obviously the API we settled on in #301 is pretty different from what was proposed in this RFC, but I think most of the differences aren't actually that significant, however, two things did stick out to me:

1. Chaining of actions. This was proposed in this RFC. It would be totally possible to implement with the new API design, we just haven't done it. Do we want this? I think it could possibly make errors harder to track down properly, so there is a bit of a downside to it. It would make some cases a bit shorter, I am not sure how common those cases are though.

2. Decoupling from the DOM. This is really the biggest difference. The API proposed here can be used with something like Selenium/Playwright, where commands are sent over the wire, whereas the API in #301 is tied to running directly against a DOM. I think that maybe the solution here might be to have a completely separate implementation of interactors for something like Selenium, but with a similar or even identical API. Just like with native interactors, it will be very difficult to have a shared set of interactors for different platforms, and it would effectively limit us to a subset of the functionality of either option.

PR #115, opened 2020-02-11
Welcome message from old Bigtest redirects

minkimcello:
## Approach
Added new components for the landing page for users that were directed there from the old bigtest website.

The old website will dynamically use ~~`https://new-bigtest.io/?from-old-website=true&archived-page={last-page-they-were-at}`~~ `https://new.bt.io/?archived-page={last-page-they-were-at}` to redirect users to the new landing page.

The changes necessary to be made on the old website to accommodate the changes of _this_ pull request for the new website can be found [here](https://github.com/bigtestjs/bigtestjs.io/pull/74).

## Todo
- [x] Styling needs adjusting.
- [x] Are ~~`from-old-website`~~ and `archived-page` params appropriate? Reduced to one param.
- [x] ~~Ran prettier but not sure why it's using double quotes contrary to its prettier config file.~~ Not going to worry about it for now.


packages/website/src/components/Announcement.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const redirected = (props): boolean => {
+  const params = new URLSearchParams(props.search);
+  const from_old_website: string = params.get('from-old-website')!;
+  return from_old_website == 'true' ? true : false;
+};
+
+const from_where = props => {
+  const params = new URLSearchParams(props.search);
+  const from_which_page: string = params.get('archived-page')!;
```
jorgelainfiesta: We'd need to prepend the archive site url to the param (like `https://v1.bigtest.io`). It could be argued that we should store that URL in some variable at the project-level, but this announcement is a very unique thing so I think it's fine to just hard-code it here.

packages/website/src/components/Announcement.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <Text>
+      You can still find the page you were trying to visit in our archive:{' '}
+      <a href={from_which_page}>{from_which_page}</a>
```
jorgelainfiesta: Link do have a special style, so you have to use something like:

```
const AnnouncementLink = styled.a`
  border-bottom: 1px dotted ${({ theme }) => theme.colors.bodyCopy};
`;
```

jorgelainfiesta:
Looks great Min!

The only tiny style missing is for the links, the rest is perfect.
I'm ok with those params. 
Let's just do as Prettier says, uniform code is what matters at this point. We can change the rules later on and the tool will take care of it.

packages/website/src/components/Announcement.tsx
```diff
@@ -0,0 +1,58 @@
...
+import Section from './Section';
+
+interface AnnouncementProps {
+  location: Object;
+}
+
+const AnnouncementBox = styled(Box)`
+  background: #ffe9bf;
+  width: 60%;
+  padding: ${({ theme }) => theme.space.medium} ${({ theme }) => theme.space.large};
```
jorgelainfiesta: Can you please change this padding to the following? (It's to adjust for the paragraph margin)

```
padding: ${({ theme }) => theme.space.large} ${({ theme }) => theme.space.large} ${({ theme }) => theme.space.xSmall} ${({ theme }) => theme.space.large};
1.6103rem 1.6103rem 0.3818rem 1.6103rem
```

packages/website/src/components/Announcement.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <Text>
+      You can still find the page you were trying to visit in our archive:{' '}
+      <a href={from_which_page}>{from_which_page}</a>
+    </Text>
+  );
+};
+
+const Announcement: React.FC<AnnouncementProps> = props => {
+  if (redirected(props.location)) {
```
jorgelainfiesta: I'm not convinced by this pattern. I'd think it's more intuitive if the parent determines whether this component is visible, rather than the component determining itself whether to show up or not. But perhaps I'm way too used to thinking in ember terms? cc. @taras 

If we're to leave this logic inside the component, we'd have to rename this component to something more specific, as `Announcement` reads more like a general-use piece.

cowboyd: I agree that we should probably rename this component to `<MaybeGreetLegacySiteUsers>`

packages/website/src/components/Announcement.tsx
minkimcello: okay will do. and taras helped me with ruby last night to make this much easier.

packages/website/src/components/Announcement.tsx
minkimcello: oh okay 👌

packages/website/src/components/Announcement.tsx
minkimcello: Will do

packages/website/src/components/Announcement.tsx
minkimcello: oh yeah, i put the logic in index at first but moved it into components to make it look nicer. so should i rename the component to `<Maybe...>` or move the logic onto the index page? @cowboyd

cowboyd: Try both and ask yourself which one you like best and why?

packages/website/src/components/Announcement.tsx
minkimcello: i prepended the url from the old website so it's not necessary to hard-coded it from the new website.

packages/website/src/components/Announcement.tsx
minkimcello: Did some refactoring to simplify the logic and renamed the component to be more explicit. 

I like the idea of keeping components modular so for now I decided to keep the conditional inside the component as this logic only applies for this component only at the moment.

packages/website/src/components/GreetLegacyUsers.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <Section>
+        <GreetLegacyBox>
+          <H4 color={'bodyCopy'}>Coming from the old BigTest Site?</H4>
+          <Text>
+            Welcome to the new BigTest! We have made major changes in the project to make BigTest the ultimate
+              testing framework. Take a look around, and reach out to us if you have any questions{' '}
+            <GreetLegacyLink href="mailto:bigtest@frontside.io">bigtest@frontside.io</GreetLegacyLink>.
+            </Text>
+          <Text>
+            You can still access the old website here: <GreetLegacyLink href="https://hello.com">https://v1.bigtestjs.io</GreetLegacyLink>
```
jorgelainfiesta: This link should point to `https://v1.bigtestjs.io`

packages/website/src/pages/index.tsx
```diff
@@ -44,6 +45,8 @@ const IndexPage = () => (
         />
       </Flex>
     </Hero>
+    {/* <GreetLegacyUsers /> */}
```
jorgelainfiesta: We probably don't need this comment.

minkimcello:
@jorgelainfiesta I'll squash and merge this time :P

packages/website/src/components/GreetLegacyUsers.tsx
```diff
@@ -24,21 +24,19 @@ const GreetLegacyLink = styled.a`
...
+const GreetLegacyUsers: React.FC<GreetLegacyProps> = () => {
+  const archived_param = new URLSearchParams(window.location.search);
+  const from_legacy = archived_param.get('archived-page');
+  if (from_legacy) {
     return (
       <Section>
         <GreetLegacyBox>
           <H4 color={'bodyCopy'}>Coming from the old BigTest Site?</H4>
           <Text>
             Welcome to the new BigTest! We have made major changes in the project to make BigTest the ultimate
```
cowboyd: Should tis be "we're making major changes" vs "we have made changes"

PR #116, opened 2020-02-11
Initial commit for @bigtest/project

cowboyd:
Motivation
----------
See 

- https://github.com/bigtestjs/server/issues/80
- https://github.com/bigtestjs/server/issues/75

We need a way to store common working files between different processes that will be operating on a BigTest project. For example, when the CLI tries to connect to a server, it needs to know which websocket port to connect to. This type of runtime data has to live in a well-known place, and we need a well-known way of discovering it and accessing it that can be shared between different codebases.

Approach
----------

We'll follow the time honored tradition of looking for a bigtest specific configuration file inside up the directory tree until we find it, and then having a way writing and reading from it.

This is just a mostly empty shell of a library where this method for accessing project configuration will live. The main thing here is to get it placed, and get it building, testing, and deploying to NPM properly.

### TODOS

- [x] How do we build our index.js with typescript
- [ ] Make sure integrates with github preview actions
- [ ] How do we run our tests


cowboyd:
Closing as deploy preview doesn't seem to be working.

PR #117, opened 2020-02-12
Website: Adds typekit fonts 

jorgelainfiesta:
## Motivation

The institutional fonts are served through Typescript because they're not free. Before, users would only see the font if they had it installed, and the site would default to another option.

## Solution

Jeffrey created a typescript webkit for us with the fonts we need. Now we serve typescript fonts to the user so everybody sees the same font.

### DIFF
--- a/packages/website/gatsby-config.js
+++ b/packages/website/gatsby-config.js
@@ -15,7 +15,7 @@ module.exports = {
     {
       resolve: `gatsby-plugin-mailchimp`,
       options: {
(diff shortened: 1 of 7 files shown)


cowboyd:
Front page a little bit whack at the bottom on mobile. 
https://share.icloud.com/photos/0Jj0LbalmMY8bwVjscp0H_TCA

jorgelainfiesta:
Oh, thanks Charles! It's fixed now.

PR #118, opened 2020-02-12
Website: Applies prettier and introduces stylelint

jorgelainfiesta:
## Motivation

We want to have a consistent code for the website. Tooling can contribute to achieve that goal.

## Approach

Min had introduced `prettier` before. This PR applies the checks by prettier and adds `stylelint` configured to work with `styled-components`. 

New commands are introduced. 

- `lint:ts`: runs prettier to check the format of the typescript files. 
- `lint:styles`: runs `stylelint` to check the format of `styled-components` inside typescript files.
- `lint`: runs both lint checks. I decided to us `;` instead of `&&` to provide the full picture to a developer when they run `lint`. If we used `&&`, the developer could think "oh great, I just have this ts errors" and then when they run it again to check what they fixed, new style errors pop up.

`format:ts`: was renamed because styles format cannot be automatically fixed when using `styled-components`. 

Also, basic rules for `.editorconfig` were introduced.

### DIFF
--- a/packages/website/.editorconfig
+++ b/packages/website/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+indent_style = spaces
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+end_of_line = lf
--- a/packages/website/.stylelintrc
+++ b/packages/website/.stylelintrc
@@ -1,0 +1,10 @@
+{
+  "processors": ["stylelint-processor-styled-components"],
+  "extends": [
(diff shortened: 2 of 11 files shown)


cowboyd:
How can we introduces style rules for all of the packages in this repo, not just the `website`? It will soon become tedious to maintain this across 7 different packages.

jorgelainfiesta:
After reading a few articles, it seems that the general recommendation is to treat each package independently. In the case of linting, we can publish a `frontside-eslint-rules` package and use it on every package. That way linting is installing a regular dependency as we would do anyways in each package, instead of setting up rules for each package.

PR #126, opened 2020-02-14
Add @bigtest/agent

cowboyd:
Motivation
----------

There are a couple of problems with our current setup of having the agent living inside the bigtest/server.  First and foremost, we're building the agent application code every single time and that causes a lot of unwanted startup time. We're willing to deal with this in development, but this is really going to add significant time to work with bigtest applications in real life, and also consume lots of unecessary computational resources in CI.

Approach
--------

This extracts the agent server app into a pre-built web application that uses parcel and microbundle to compile the HTML and typescript respectively.

In addition, an enhancement has been made to have the agent server bind to optionally bind to a different port every time.

The actual agent application lives in the `/app` directory, and is served from `/app/index.html`

`index.ts` is the code meant to be called from Node.js for starting and stopping an agent server.

### DIFF
--- a/packages/agent/bin/helpers.ts
+++ b/packages/agent/bin/helpers.ts
@@ -1,0 +1,31 @@
+import { main as effectionMain, Operation } from "effection";
+import { EventEmitter } from "events";
+
+export const self: Operation = ({ resume, context: { parent }}) => resume(parent);
+export const parent: Operation = ({ resume, context: { parent }}) => resume(parent.parent);
+
+export function main(operation: Operation) {
+  return effectionMain(function*() {
+    let mainContext = yield self;
+    let interrupt = () => { mainContext.halt(); };
+    try {
+      process.on('SIGINT', interrupt);
+
+      yield operation;
+
+    } catch (error) {
+      console.log(error);
+      process.exit(-1);
+    } finally {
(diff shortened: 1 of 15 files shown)


packages/agent/bin/helpers.ts
```diff
@@ -0,0 +1,31 @@
+import { main as effectionMain, Operation } from "effection";
+import { EventEmitter } from "events";
+
+export const self: Operation = ({ resume, context: { parent }}) => resume(parent);
+export const parent: Operation = ({ resume, context: { parent }}) => resume(parent.parent);
+
+export function main(operation: Operation) {
```
cowboyd: Once we make `effection` into a monorepo, this function will live in `@effection/node`

packages/agent/bin/helpers.ts
```diff
@@ -0,0 +1,31 @@
...
+    } catch (error) {
+      console.log(error);
+      process.exit(-1);
+    } finally {
+      process.on('SIGINT', interrupt);
+    }
+  })
+};
+
+export function once(emitter: EventEmitter, event: string | symbol): Operation {
```
cowboyd: Once we make effection into a monorepo, this will live in package `@effection/events`

PR #139, opened 2020-02-21
Import `@bigtest/server` package

cowboyd:
This brings over the repo at https://github.com/bigtestjs/server along with all of its history.

Currently preview and release are disabled since it is not yet ready to be made into an npm package.


packages/server/.babelrc
```diff
@@ -0,0 +1,4 @@
+{
```
cowboyd: These are the presets needed to compile the manifest correctly. Very odd indeed, and something that we'll have to work out since it doesn't apply to the actual _server_ code, but the manifest code that the server compiles with parcel.

PR #162, opened 2020-02-24
Strip code from manifest

jnicklas:
Frustratingly this doesn't typecheck, though I've verified that it does actually work as intended.


packages/server/src/manifest-builder.ts
```diff
@@ -18,6 +21,12 @@ interface ManifestBuilderOptions {
   distDir: string;
 };
 
+const stripCode = compose(
```
cowboyd: Where are we actually using this?

cowboyd:
I'm guessing that we don't need this since we discovered that `JSON.stringify()` strips code by default.

jnicklas:
Yupp

PR #165, opened 2020-02-24
Process commands

jnicklas:
This is working towards actually running the tests. Once the command server receives a run mutation it will enqueue this in the orchestrator which will then pick it up and forward it to the agent. The agent will then proceed to run the supplied tests. We're handing the agent a subset of the manifest (or the entire manifest), which it will recursively descend through.


packages/agent/app/agent.ts
```diff
@@ -1,59 +1,44 @@
...
-      scriptElement.src = message.manifest;
-      scriptElement.addEventListener('load', () => {
-        console.log('[agent] loaded test manifest', __bigtestManifest);
-      });
-      document.body.appendChild(scriptElement);
+  while(true) {
+    let message = yield connection.receive();
 
+    if(message.type === "run") {
+      fork(run(testFrame, message));
```
cowboyd: Looks like the `fork` expression won't ever run.

packages/agent/app/agent.ts
```diff
@@ -1,59 +1,44 @@
...
-  console.log('[agent] connecting to orchestrator at', connectTo);
-  let socket = new WebSocket(connectTo);
-
-  socket.addEventListener('open', () => {
-    socket.send(JSON.stringify({
-      type: 'connected',
-      data: Bowser.parse(navigator.userAgent)
-    }));
-    console.log('[agent] socket connection established');
+  fork(function*() {
```
cowboyd: Here's another "lone" fork?

packages/agent/app/index.html
```diff
@@ -19,6 +42,9 @@
   </head>
 
   <body>
+    <div id="splash">
+      <img src="./assets/bigtest-logo.svg">
```
cowboyd: 😍 😍 😍 

If the agent is going to just be standing around, might as well be wearing nice clothes, right? love to see it...

packages/agent/app/socket-connection.ts
```diff
@@ -0,0 +1,24 @@
+import { Operation } from 'effection';
+import { Mailbox } from './effection/mailbox';
+
+export class SocketConnection {
+  private connection: WebSocket;
+
+  static *start(connectTo: string) {
+    let socket = new WebSocket(connectTo);
+    let mailbox = yield Mailbox.watch(socket, ['open', 'message', 'close']);
```
cowboyd: Should this mailbox be watching in the parent scope?

jnicklas:
@cowboyd one thing in this PR is that it basically copy-pastes the mailbox implementation into this package, that's not ideal, of course. Can we extract this into its own package, maybe? My packaging-fu is not strong enough to work this out myself.

cowboyd:
Should we make it `@bigtest/events`? or go through the ordeal of setting up `effection` as a monerepo. Alternatively, we could actually set up `@effection/events` inside the bigtest monorepo since @frontsidejack is the one with the credentials to publish both of those NPM packages.

jnicklas:
This has now all been extracted to other PRs, the last bit that hasn't been merged yet is in #205

Issue #168, opened 2020-02-24
Responsive design tweak for GreetLegacyUser component


minkimcello:
## Motivation
The legacy announcement that welcomes users from the old website `<GreetLegacyUsers>` needs to be adjusted so that it spreads out more on smaller screens.

![Screen Shot 2020-02-24 at 6 53 55 PM](https://user-images.githubusercontent.com/29791650/75201746-4cb57680-5737-11ea-81f7-c7b881513362.png)

Issue #164, opened 2020-02-24
Overlapping component styling issue


minkimcello:
## Motivation
The styling of the last section of the landing page needs to be adjusted as it's being overlapped at the moment.

<img width="724" alt="Screen Shot 2020-02-24 at 9 57 21 AM" src="https://user-images.githubusercontent.com/29791650/75163123-b3617280-56ec-11ea-9dad-912bd5ee1ec0.png">

PR #170, opened 2020-02-25
Fix mobile styles

jorgelainfiesta:
## Motivation

Two mobile style problems: #168 #164 

## Approach

Rely on `styled-system` to manage widths for responsive styles.


minkimcello:
The subscribe button looks like it's still overlapping for me (on Firefox). 🤷‍♂

Screen Shot 2020-02-25 at 9 20 57 AM

jorgelainfiesta:
Ah I see, thanks! I made the column break earlier to make sure it works everywhere.

PR #172, opened 2020-02-25
Use atomic slices to improve type-checking of lens operations

cowboyd:
Handy as they are, using TypeScript with Ramda lenses can be annoying at best, and perilous at worst. There are two basic problems:

1. lens operations are cumbersome in that you have to store a reference to not only the child and the root value, but also the lens.
2. the lenses erase type information. Ideally they shouldn't since they don't in other statically typed languages, but they don't type easily since TS does not support higher kinded types.

This solves both of those problem by capturing a lens into a `Slice` object. That way, there's no need to construct an extra referenc, and all of the operations on that slice are implicity happening on the focus of that lens. It solves the typing problem by the fact that `Slice`s are parameterized with the type of the lens's focus. While it isn't 100% correct, it does side-step the issue of typing lenses entirely by essentially declaring at the point you take the slice "this slice has this type" and while that is not strictly type-safe, if you do get that right, then every subsequent usage of the slice thereafter _is_.

Before:

```ts
import { lensPath } from 'ramda';

const agentsLens = lensPath(['agents']);
// add  an agent
options.atom.over(agentsLens, assoc(identifier, assoc("identifier", identifier, data)));

// remove an agent 
options.atom.over(agentsLens, dissoc(identifier));
```

After:

```ts
let agent = options.atom.slice<AgentState>(['agents', identifier]);

// add the agent 
agent.set(agentData); // must type check against `AgentState`

// remove the agent
agent.remove();
```

### DIFF
--- a/packages/server/src/orchestrator/atom.ts
+++ b/packages/server/src/orchestrator/atom.ts
@@ -27,28 +27,68 @@ export class Atom {
     this.subscriptions.emit('state', this.state);
   }
 
-  // Ramda Lens types. How do they work?
-  //eslint-disable-next-line @typescript-eslint/no-explicit-any
-  view(lens: any): unknown {
-    return R.view(lens, this.get());
-  }
-
-  // Ramda Lens types. How do they work?
-  //eslint-disable-next-line @typescript-eslint/no-explicit-any
-  over(lens: any, fn: any) {
-    //eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.update(R.over(lens, fn) as any as (state: OrchestratorState) => OrchestratorState);
-  }
-
-  // Ramda Lens types. How do they work?
-  //eslint-disable-next-line @typescript-eslint/no-explicit-any
-  set(lens: any, value: unknown) {
-    //eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.update(R.set(lens, value) as any as (state: OrchestratorState) => OrchestratorState);
+  slice<T>(path: string[]): Slice<T> {
+    return new Slice(this, path);
   }
 
   *next(): Operation {
     let [state]: [OrchestratorState] = yield on(this.subscriptions, 'state');
     return state;
+  }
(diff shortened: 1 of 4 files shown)


packages/server/src/orchestrator/atom.ts
```diff
@@ -27,28 +27,68 @@ export class Atom {
...
+      this.atom.update(state => {
+        let array = parent as unknown[];
+        return R.set(parentLens, array.filter(el => el !== this.get()), state) as unknown as OrchestratorState;
+      })
+    } else {
+      let [property] = this.path.slice(-1);
+      this.atom.update(state => {
+        return R.set(parentLens, R.dissoc(property, parent), state) as unknown as OrchestratorState;
+      })
+    }
```
jnicklas: It's a bit of a smell that this ends up being so complicated.

taras: @cowboyd could we use microstates like api or microstates here? it would make this entire block a one liner

cowboyd: We could, but microstates does not currently have any TypeScript typings. How much effort would it be to fix that?

cowboyd: > It's a bit of a smell that this ends up being so complicated.

Yeah, the problem is that you need to capture the relationship between the objects when you are establishing the lens. The alternative (apart from using microstates, which abstracts over lenses, but does not currently have any type information) would be to manually walk the `path` given to `slice` and if it's a string, use `lensProp`, or if it's a number, use `lensIndex`

taras: > How much effort would it be to fix that?

It can be very doable if we scrap the DSL (potentially use something similar to Robbie's interactors) and drop the context funny business.

cowboyd:
I'm going to go ahead and merge this since it's more TypeSafe than what we have currently, and is a pre-requisite for some other stateful effects

PR #173, opened 2020-02-25
Lint all the things

cowboyd:
Now that we have everything in a monorepo, it's time to get our listing story in line.

This adds a lint workflow to run on each PR that delegates to each package (and also fixes the current crop of listing errors)


jorgelainfiesta:
Should the same rules be applied to `packages/website`?

cowboyd:
Possibly. I agree that we should unify them at some point, but this this just runs them on the build.

PR #175, opened 2020-02-26
Add link to v1 docs to header

jorgelainfiesta:
## Motivation

Users of v1 did not have a way to access the archived guides if they visited `https://bigtestjs.io`. 

## Approach

Added a link on the header pointing to `v1` documentation.

### DIFF
--- a/packages/website/src/components/Header.tsx
+++ b/packages/website/src/components/Header.tsx
@@ -39,6 +39,16 @@ const Nav = styled.nav`
 //   margin: 0 ${({ theme }) => theme.space.medium};
 // `;
 
+const LegacyLink = styled.a`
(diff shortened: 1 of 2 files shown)


jorgelainfiesta:
`@bigtest/convergence` release has been on `1.1.2` for 4 months now. In general the versions seems chaotic. I could not find any `@bigtest` package with version `0.7`. I'm very confused.

Also, we set the old website name alias as `v1.bigtest.io` together. I'm surprised that you bring that up now. I don't know what to do. Do we rename and redirect everything again?

jorgelainfiesta:
> Although when I follow the link, the messaging here looks a bit brusque.

I'll prepare another PR on the old website to change the wording.

PR #178, opened 2020-02-26
Add Test interfaces for use across components

cowboyd:
Motivation
----------

There are several interfaces that we use across different components to work with tests. Most of them don't need to know about the specifics of the implementation of the test, but instead just the structure of them. There are some systems like the harness and any DSL that will need to know about the test implementation interface however.

Approach
----------
This adds a `@bigtest/suite` package that will serve as a home for the shared interfaces, some of which will need to be bundled on both the browser as well as in `node`. There are two separate interfaces for `Test` and `TestImplemenation` that represent the structure as well as the "live functions" respectively.

In the declaration of the interfaces themselves, there is some duplication for the sake of readability that has no effect on correctness.

Currently, this package only contains shared interface definitions, but is the likely home for the eventual DSL.

### DIFF
--- a/packages/suite/.eslintrc.json
+++ b/packages/suite/.eslintrc.json
@@ -1,0 +1,10 @@
+{
+  "extends": "plugin:@typescript-eslint/recommended",
+  "plugins": ["prefer-let"],
+  "rules": {
+    "prefer-const": 0,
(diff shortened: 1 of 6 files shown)


jnicklas:
Looks great!

PR #179, opened 2020-02-27
Consume @bigtest/suite from within @bigtest/server

cowboyd:
Now that we have a @bigtest/suite package that contains the interfaces of what comprises a tests that can be shared between the harness, server, and manifest, this points the server at those interfaces and removes those local copies.

Note that the GraphQL schema has been updated to match these interfaces as well. Specifically, the `path` element has been removed since it is not present in the generated manifest.


jnicklas:
The generated manifest does actually contain `path`, it is the path of the test file that this part of the tree originated from, see here: https://github.com/thefrontside/bigtest/blob/7cc29f6b39424aa9d8b0d339d878877b4f7376a8/packages/server/src/manifest-generator.ts#L25

However, `steps` and `assertions` should not contain `path`, this is a relic we forgot to remove.

cowboyd:
Ah, I see. Perhaps we should find a different name for this value since everywhere else in our system, `path` means an array of strings that help you navigate a tree. `filePath`? It seems weird that we would have both a `filePath` and a `fileName` property on manifest.

jnicklas:
Yeah, totally agree that the naming isn't great. Maybe the problem is including `fileName` in the manifest in the first place, which maybe doesn't *really* make sense. Then `filePath` would be a more natural name.

PR #181, opened 2020-02-27
Simplify and reorganize Agent build for multi-file

### LINKED ISSUE #171: Replace agent start script with parcel
The agent package has a start script in `bin/start` which in turn delegates some work to `bin/run-parcel`, this seems a bit unnecessary to me. We can just modify the run script to use the parcel binary and we can remove both of these files.

The one downside is that the built in server in parcel doesn't seem to set the content type correctly for the root path, so we have to use: `http://localhost:24400/index.html` rather than just `http://localhost:24400` but this is a pretty minor issue compared to getting rid of two rather complicated scripts imo.

cowboyd:
resolves #171 

We need a way to test the agent, and that means adding a websocket client that can connect to the agent, which means adding a new class, which means that we're outgrowing our rinky-dink index.ts

This moves the source files into the canonical `src/` directory, and splits the server out into its own file. That way, we can move new source files in as we will.

Also, this migrates away from `microbundle`, since the control classes are not used anywhere except for in node and so there's no need to build separate `umd` and `ejs` bundles. In fact, there's no need for bundles at all.


jnicklas:
Did you see this: https://github.com/thefrontside/bigtest/issues/171

I think just shelling out to parcel is fine for the agent build, which would remove quite a bit of complexity here. We don't have the same needs here as in the manifest builder.

cowboyd:
@jnicklas I did, but didn't see just how simple it could be. This just invokes the parcel directly. I've tried it out with the workflow, and it does work with just the port name, and not appending `index.html` to the url, so I think we're good to go?

PR #187, opened 2020-03-02
Allow sharing of effection utilities between bigtest packages.

cowboyd:
Motivation
----------

While effection makes working with async processes a lot eaiser, it is still a primitive, and to be truly useful, most of the time what you really want is to be working with higher level constructs. We've built a lot of these constructs for working in a variety of environments: nodejs, event emitters, etc... Now that we a bunch of packages, we need to have these utitilies on-hand so that they can be used everywhere rather than be copy-pasted around like they are now.

Approach
--------

Ideally, we'd like to publish these package in the `@effection` namespace. However, we don't currently have monorepo support set up for effection, and so rather than go through the hassle of that, this just creates a single package that can be shared within the bigtest project itself. Once we start using effection outside of bigtest, then we can take up the work of migrating these utilities to their permanent home.

To start with, this extracts perhaps the most-used utility for async programming: the `once()` method which allows synchronization on the `EventEmitter.once()` interface. Other places in the codebase that were using their own copies have been updated to use the shared version.

After this, we can introduce one utility at a time along with the changes that replace local copies to use the central version.

### DIFF
--- a/packages/effection/.eslintrc.json
+++ b/packages/effection/.eslintrc.json
@@ -1,0 +1,10 @@
+{
+  "extends": "plugin:@typescript-eslint/recommended",
+  "plugins": ["prefer-let"],
+  "rules": {
+    "prefer-const": 0,
+    "prefer-let/prefer-let": 2,
+    "@typescript-eslint/no-use-before-define": 0,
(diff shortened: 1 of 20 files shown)


jnicklas:
Looks great! 🚢

PR #188, opened 2020-03-03
Move implementation of Mailbox into effection package

jnicklas:
So it can be reused from different places


packages/effection/src/mailbox.ts
```diff
@@ -0,0 +1,104 @@
+import { monitor, Operation } from 'effection';
+import { EventEmitter } from 'events';
+
+import { compile } from './pattern';
+export { any } from './pattern';
+
+type EventEmitterName = string | symbol;
```
cowboyd: I think this symbol is already defined in the `events` module as `EventName`

jnicklas: Yes, looks like!

PR #191, opened 2020-03-04
Update landing page copy

jorgelainfiesta:
## Motivation

The previous copy was not specific enough about how BigTest is bringing new and exciting stuff to the testing ecosystem. 

## Solution

A new copy was produced and edited. This time it includes many more technical details on what makes BigTest different from existing alternatives. 

## Next steps

Decide on a new slogan.

### DIFF
--- a/packages/website/src/components/WhyBigTest.tsx
+++ b/packages/website/src/components/WhyBigTest.tsx
@@ -37,28 +37,28 @@ const WhyBigTest: React.FC<{ title?: string }> = ({ title }) => (
     <H2 color="secondary">{title ? title : 'Why BigTest?'}</H2>
     <List>
       <Item largeOrder="0">
-        <H4>Seriously fast</H4>
-        <p>Designed and implemented to speed up test writing, execution, and maintenance.</p>
+        <H4>Helpful test results</H4>
+        <p>Designed to diagnose the cause of test failure at a glance.</p>
       </Item>
       <Item largeOrder="3">
-        <H4>Experience-centered</H4>
-        <p>Test what matters to users through a refined developer experience.</p>
-      </Item>
-      <Item largeOrder="1">
-        <H4>Revolutionary architecture</H4>
+        <H4>Revolutionary Architecture </H4>
         <p>Built from the ground up with new technologies for modern apps.</p>
       </Item>
+      <Item largeOrder="1">
+        <H4>Easy To Use</H4>
+        <p>Flexibly built to facilitate test writing, running, and maintenance.</p>
+      </Item>
       <Item largeOrder="4">
-        <H4>Cross-Browser/Device</H4>
-        <p>If your user can see it, you can test it. Mix and match any browser and device or OS.</p>
+        <H4>Cross-browser, cross-device</H4>
+        <p>Tests your apps in user-relevant contexts and scenarios.</p>
       </Item>
       <Item largeOrder="2">
-        <H4>Cross-Framework</H4>
-        <p>Test no matter what you build your UI with: React, Vue, Ember, Angular, bring it on!</p>
+        <H4>Framework agnostic</H4>
+        <p>Works with everyone's preferred JavaScript framework.</p>
       </Item>
       <Item largeOrder="5">
-        <H4>Extensible</H4>
-        <p>We provide powerful APIs to bring testing to the new decade.</p>
+        <H4>Dynamically Extensible</H4>
+        <p>Conceived to create a community-based robust testing ecosystem.</p>
       </Item>
     </List>
   </Section>
--- a/packages/website/src/pages/index.tsx
+++ b/packages/website/src/pages/index.tsx
@@ -27,9 +27,9 @@ const IndexPage = props => (
             <br />
             not the other way around
           </H1>
-          <Text>
-            BigTest is a framework that enables you to test real-world experiences on your app rapidly, in every
-            browser and device, regardless of which framework you use.
+          <Text fontSize="large">
+            BigTest lets developers test their applications across browsers and devices and provides them with
+            swift and actionable feedback.
           </Text>
           <Box marginTop={['small', 'small', 'small', 'xxLarge']}>
             <H3>Get notified about BigTest progress and know when it's ready:</H3>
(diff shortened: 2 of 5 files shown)


packages/website/src/pages/index.tsx
```diff
@@ -75,16 +85,22 @@ const IndexPage = props => (
...
         <Box width={[1, 1 / 2, 2 / 3]} paddingRight={[0, 'medium', 'xxLarge']}>
-          <H2>Seriously fast testing</H2>
+          <H2>Tests that are quick to write, run, and maintain</H2>
           <Text>
-            When you're trying to test from the perspective of your user, there's a lot of information coming in.
-            All the clicks, streaming updates from the server, and the hundreds (or thousands) of background
-            processes create a highly asynchronous environment. As a result, UI testing has historically been
-            time-intensive in both setup and run time.
+            Tests are tools to help us see what we don’t know; they alert us to what is broken when we thought
+            something was unaffected. The best tests swiftly and precisely diagnose the issue so teams can iterate
```
minkimcello: `The best tests swiftly and precisely diagnose its issues so teams can iterate`

packages/website/src/pages/index.tsx
```diff
@@ -75,16 +85,22 @@ const IndexPage = props => (
...
+            and problem-solve on the spot.
           </Text>
           <Text>
-            BigTest brings a new stateful architecture, along with several paradigm shifts that allow tests to be
-            faster to write, execute, and maintain.
+            BigTest facilitates the creation and maintenance of tests by employing Interactors. Instead of relying
+            on class-names or HTML-attributes which continuously change, Interactors are easy to use stable
+            interfaces that developers can use (and reuse) throughout their tests.
```
minkimcello: `BigTest facilitates the creation and maintenance of tests by employing Interactors; Interactors are easy to use stable interfaces that developers can use (and reuse) throughout their tests as opposed to relying on class-names or HTML-attributes which continuously change.`

packages/website/src/pages/index.tsx
```diff
@@ -95,16 +111,22 @@ const IndexPage = props => (
     <Section>
       <Flex flexWrap="wrap">
         <Box width={[1, 1 / 2, 2 / 3]} paddingRight={[0, 'medium', 'xxLarge']}>
-          <H2 color="secondary">BigTest makes your tests reusable</H2>
+          <H2 color="secondary">Innovative architecture</H2>
+          <Text>
+            BigTest’s architecture represents a departure from traditional test runners. Instead of being a
+            stateless tool, BigTest relies on data storage and sharing to allow users to interact with the test in
```
minkimcello: `tests`?

packages/website/src/components/WhyBigTest.tsx
```diff
@@ -37,28 +37,28 @@ const WhyBigTest: React.FC<{ title?: string }> = ({ title }) => (
     <H2 color="secondary">{title ? title : 'Why BigTest?'}</H2>
     <List>
       <Item largeOrder="0">
-        <H4>Seriously fast</H4>
-        <p>Designed and implemented to speed up test writing, execution, and maintenance.</p>
+        <H4>Helpful test results</H4>
+        <p>Designed to diagnose the cause of test failure at a glance.</p>
```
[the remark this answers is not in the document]
jorgelainfiesta: From Dr. Olson

> This is really a matter of style versus grammar. If you wanted you could say causes versus cause, but then you’re implying that the user writes really crappy code that’s full of glitches (which might be true, but I don’t think you make friends by saying so). That’s why I kept it singular.

packages/website/src/components/WhyBigTest.tsx
```diff
@@ -37,28 +37,28 @@ const WhyBigTest: React.FC<{ title?: string }> = ({ title }) => (
...
-        <H4>Cross-Framework</H4>
-        <p>Test no matter what you build your UI with: React, Vue, Ember, Angular, bring it on!</p>
+        <H4>Framework agnostic</H4>
+        <p>Works with everyone's preferred JavaScript framework.</p>
       </Item>
       <Item largeOrder="5">
-        <H4>Extensible</H4>
-        <p>We provide powerful APIs to bring testing to the new decade.</p>
+        <H4>Dynamically Extensible</H4>
+        <p>Conceived to create a community-based robust testing ecosystem.</p>
```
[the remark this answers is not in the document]
jorgelainfiesta: > Well, the question here is whether these are co-ordinate adjectives or not (sorry to get in the grammar weeds). Long story short, if you can reverse the order of the adjectives without losing the sense of the original then you can put a comma (or “and”) between them.

packages/website/src/pages/index.tsx
jorgelainfiesta: > So the repetition of Interactors (even separated by a colon) will look like a typo (and again, stylistically is just horrible to encounter).  You could get around that by saying “Interactors; these are...”
> 
> More to the point, what is it that you want to emphasize in the second sentence – What Interactors can do, or what you avoid by using them?  Which ever that is, put that last in the sentence. 
> 
> I took it to be the ease of use of Interactors – the selling point here – and your colleague seemingly confirms that with his version’s “opposed to” – in fact, we can incorporate that stronger sense by slightly changing the text by adding “having to” (see sentence).
> 
> So again this is a matter of style and meaning, and I thought you got the big picture right the order in which you wrote it.

packages/website/src/components/WhyBigTest.tsx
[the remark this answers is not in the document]
minkimcello: Then still it should be `Designed to diagnose the cause of *a* test failure at a glance` if we want to imply that there aren't many bugs.

packages/website/src/components/WhyBigTest.tsx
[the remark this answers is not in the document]
minkimcello: So we should go with either of the suggestion I made because those adjectives *can* be reversed?

packages/website/src/pages/index.tsx
minkimcello: I think for this one, it's safe to use plural because it doesn't necessarily imply there will always be multiple (because of bad code) but just a grammatical issue.

packages/website/src/pages/index.tsx
minkimcello: 👌. My primary concern, aside from possible grammatical errors, was that the cadence of most of the content on the landing page is too repetitive. Like reading a 20-line poem with the same rhythm. So I often prefer `{ A } + { B }` over `{ B }, { A }`.

taras:
The biggest challenge here is that @jorgelainfiesta is in the middle between the editor and our feedback. This is frustrating for Jorge, adds work for the editor and doesn't provide the necessary dialog to help resolve the feedback. 

We're going to punt on editing editor's work at the moment and accept his as an authority on the subject of grammar. Let's get it out there and ask people for feedback on our copy. If we get negative feedback then we'll re-evaluate this.

taras:
> Sorry

@minkimcello there is nothing to be sorry about. You didn't do anything wrong. Frontside is pedantic. That's just how we are. You want to make sure that we put the best foot forward. That is absolutely the right thing to do. We just have to make sure that we're setup to do that efficiently. Your feedback highlighted the fact that we're not setup correctly to allow this kind of discussion. We'll address this issue in the future but unblocking this PR is a work around for the current situation. Again, you have nothing to apologize here. You absolutely did the right thing.

PR #193, opened 2020-03-05
Add command processor to orchestrator

jnicklas:
The command processor is basically a run-loop for the orchestrator. It picks off commands from its inbox and runs them.

Extracted from #165


jnicklas:
@cowboyd sounds reasonable to me, but it might be better to do that as a separate PR?

PR #194, opened 2020-03-05
Extract nodejs main() method into shared @bigtest/effection

cowboyd:
Motivation
----------

When you're managing a node process, there are things that you almost always want to do: setup a SIGINT handler to interrupt the process, catch errors and log them to the console, and exit when appropriate with the proper code. The server package had a very handy main method that was used in several places via copy paste.

Approach
----------
This consolidates all of those main methods down into a single location that can be shared by all.


packages/effection/src/index.ts
```diff
@@ -3,3 +3,4 @@ export { Mailbox } from './mailbox';
 export { any } from './pattern';
 export { suspend } from './suspend';
 export { ensure } from './ensure';
+export { main } from './node';
```
jnicklas: Should this be exported from `index`? Would this cause the agent to bundle a bunch of node-specific code we don't want?

cowboyd: Ugh. You're correct... JavaScript bundling is the worst.

cowboyd:
@jnicklas agree. I'm going to go ahead and shut this in favor of that.

PR #197, opened 2020-03-06
Use playwright to test agent code

jnicklas:
This uses microsoft's playwright tool for browser automation in the agent testsuite.

### DIFF
--- a/packages/agent/src/client.ts
+++ b/packages/agent/src/client.ts
@@ -1,0 +1,103 @@
+import { fork, monitor, Operation } from 'effection';
+import { createServer } from 'http';
+import { promisify } from 'util';
+
+import {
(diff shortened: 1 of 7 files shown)


packages/agent/src/client.ts
```diff
@@ -0,0 +1,108 @@
...
+        let [error]: [Error] = yield once(httpServer, "error");
+        throw error;
+      });
+
+      try {
+        yield acceptConnections(socketServer, inbox, delegate, forEach);
+      } finally {
+        socketServer.unmount();
+      }
+    }));
```
cowboyd: We should be able to use our new effection primitives here, no?

jnicklas: Yes indeed!

PR #200, opened 2020-03-09
Lint bigtest using shared frontside eslint config

cowboyd:
Motivation
----------

We had a preponderance of eslint configs multiplying inside the monorepo, and it turned out that we weren't linting the website *at all*

Approach
----------

This updates repo to use the [shared frontside][1] config for JavaScript and TypeScript. The website which uses JSX has its own local config which extends this to make it react-aware. This could be extracted into its own package once necessary.

All of the invididual eslint configs have been removed, and all dependencies hoisted up to the top of the repo so that they all use the same version of eslint, typescript, etc...

[1]: https://github.com/thefrontside/eslint-config


packages/website/.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": ["plugin:react/recommended"],
+  "rules": {
+    "react/prop-types": "off",
```
cowboyd: This should not be necessary since we're using TypeScript. The compiler can type-check properties

packages/website/.eslintrc.json
```diff
@@ -0,0 +1,12 @@
...
+    "react/no-unescaped-entities": "off"
```
cowboyd: Since this is a Gatsby site, it will be common to have text, and so having to escape double quotes and single quotes etc. is a big pita. 

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md

My theory is that it makes a lot of sense for code, but not so much for content sites?

packages/website/src/images/index.ts
```diff
@@ -1,32 +1,34 @@
+/* eslint @typescript-eslint/no-var-requires: 0 */
```
cowboyd: eslint wants us to use `import` here, but I couldn't get it to work. If we can, then we should go ahead and re-enable this rule, and change the `require`'s to `import`'s

packages/website/src/theme/base.ts
```diff
@@ -52,11 +52,11 @@ const breakpointAliases: string[] = ['small', 'medium', 'large', 'xLarge'];
 
 const sizeAliases: string[] = ['xxSmall', 'xSmall', 'small', 'medium', 'large', 'xLarge', 'xxLarge', 'xxxLarge'];
 
-base.fontSizes.forEach((_: any, index: number): void => {
```
cowboyd: We don't need a type-specifier on a field that is never used.

PR #203, opened 2020-03-10
Remove type declaration packages from runtime dependencies

cowboyd:
depends on #202 (re-target this PR on to master that PR is merged)

Motivation
---------- 

We use TypeScript as a fundamentally develop-time tool, and have no need to distribute interface files from 3rd party packages with our own code. For example, bigtest server compiles against `@types/node`, but once it is compiled to JavaScript, there's no need for anybody downloading `@bigtest/server` to also get a copy of `@types/node` pooped into their `node_modules/` directory.

Approach
----------

This moves all `@types` packages into `devDependencies` instead of `dependencies` across the board.


cowboyd:
Hmm, type checks seem stuck, let me close and re-open

PR #204, opened 2020-03-10
♻️Create `@bigtest/parcel` package to control parcel processes

cowboyd:
Motivation
----------

We can't ship bigtest with a runtime that is dependent on ts-node. Not only is it way too gigantic, it takes too long to start up if first we have to compile all the TypeScript on every start up.

Approach
--------

This takes a step towards ts-node independence by extracting the parcel process into its own package that can be pre-compiled and imported by the `@bigtest/server`

The ParcelProcess is an object that, when allocated with the `create()` operation, will make sure that as soon as the current operation passes out of scope, the entire parcel worker farm is torn down, no matter what stage of readiness it is in.

Also, the parcel process implements a `receive()` method so that you can await updates from it.

> Note: This really speeds up the orchestrator tests and general startup time since it isn't re-compiling the parcel server code on startup 🐎 🐎 🐎


packages/parcel/src/parcel-process.ts
```diff
@@ -0,0 +1,86 @@
...
+  static *create(options: ParcelOptions): Operation {
+    let stdioMode = options.stdio || 'pipe';
+    let parcelProcess = new ParcelProcess();
+
+    let runParcel = Path.join(__dirname, 'parcel-run');
+    let child: ChildProcess = forkProcess(
+      runParcel,
+      ['--out-dir', options.buildDir, '--out-file', 'manifest.js', '--global', '__bigtestManifest', options.srcPath],
+      {
+        execPath: options.execPath || undefined,
```
jnicklas: I think we probably don't need this line?

packages/parcel/src/parcel-process.ts
```diff
@@ -0,0 +1,86 @@
...
+  mailbox: Mailbox = new Mailbox();
+
+  static *create(options: ParcelOptions): Operation {
+    let stdioMode = options.stdio || 'pipe';
+    let parcelProcess = new ParcelProcess();
+
+    let runParcel = Path.join(__dirname, 'parcel-run');
+    let child: ChildProcess = forkProcess(
+      runParcel,
+      ['--out-dir', options.buildDir, '--out-file', 'manifest.js', '--global', '__bigtestManifest', options.srcPath],
```
jnicklas: Seems a bit odd to hard-code `__bigtestManifest` here?

jnicklas: We should probably extract this into an option?

jnicklas:
This is great! Brings us one step closer towards having a fully functioning packaging solution!

packages/parcel/src/parcel-process.ts
cowboyd: This was for the case where we are running inside the tests which Stilll happen in ts-node. I couldn’t get it to work without the exec path. Maybe we could run against the preconoiled code in the tests.... in any event, The `|| undefined` is a bit silly

cowboyd: This was for the case where we are running inside the tests which Stilll happen in ts-node. I couldn’t get it to work without the exec path. Maybe we could run against the preconoiled code in the tests.... in any event, The `|| undefined` is a bit silly

jnicklas: Just realized we're using this for the tests :)

packages/parcel/test/parcel-process.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    let message: { type: string };
+
+    beforeEach(async function() {
+      await fs.writeFile("./build/test/sources/input.ts", "export default {hello: 'world'}\n");
+      message = await spawn(parcel.receive());
+    });
+
+    it.skip('notifies that a new build is available', () => {
+      expect(message).toEqual({ type: "update" });
+    });
```
jnicklas: Is it possible to only skip this on CI?

cowboyd: Excellent idea!

PR #206, opened 2020-03-12
Generic mailbox

jnicklas:
Make the Mailbox generic over the type of message, this allows us to have more typesafe mailboxes. For now we default this type to `any` so we don't need to specify the type yet, this keeps the diff more manageable.

### DIFF
--- a/packages/effection/src/mailbox.ts
+++ b/packages/effection/src/mailbox.ts
@@ -9,10 +9,11 @@ import { ensure } from './ensure';
 function isEventTarget(target): target is EventTarget { return typeof target.addEventListener === 'function'; }
 
 export interface SubscriptionMessage {
-  event: string,
-  args: unknown[],
+  event: string;
+  args: unknown[];
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export class Mailbox<T = any> {
   private subscriptions = new EventEmitter();
   private messages = new Set<T>();
@@ -57,7 +58,7 @@ export class Mailbox<T = any> {
   }
 
   *pipe(other: Mailbox<T>) {
-    let that = this;
+    let that = this; // eslint-disable-line @typescript-eslint/no-this-alias
     yield suspend(monitor(function*() {
       while(true) {
         let message = yield that.receive();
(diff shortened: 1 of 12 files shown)


packages/effection/src/mailbox.ts
```diff
@@ -52,23 +57,45 @@ export class Mailbox {
...
-  *subscribe(
-    emitter: EventEmitter | EventTarget,
-    events: string | string[],
-    prepare: (event: { event: string; args: unknown[] }) => unknown = x => x
-  ): Operation {
-    for (let name of [].concat(events)) {
-      let listener = (...args) => {
-        this.send(prepare({ event: name, args }));
+  *pipe(other: Mailbox<T>) {
+    let that = this; // eslint-disable-line @typescript-eslint/no-this-alias
```
jnicklas: Typescript doesn't like this pattern, but I'm not sure how else to get around the scope change introduced by `function*`.

packages/effection/src/mailbox.ts
```diff
@@ -52,23 +57,45 @@ export class Mailbox {
...
+    }));
+  }
 
-      if(isEventTarget(emitter)) {
-        emitter.addEventListener(name, listener);
-        yield suspend(ensure(() => emitter.removeEventListener(name, listener)));
-      } else {
-        emitter.on(name, listener);
-        yield suspend(ensure(() => emitter.off(name, listener)));
+  *map<R>(fn: (from: T) => R): Operation<Mailbox<R>> {
```
cowboyd: Ah yes of course! `map` needs to be an operation nice how that solves it.

jnicklas: Yeah, there isn't really a good way to remove the subscription otherwise, which I think will eventually lead to memory leaks, so this is probably the best solution. Also, it makes implementing this way easier.

packages/server/src/connection-server.ts
```diff
@@ -20,10 +20,12 @@ export function* createConnectionServer(options: ConnectionServerOptions): Opera
   function* handleConnection(connection: Connection): Operation {
     console.debug('[connection] connected');
 
-    let messages = yield Mailbox.subscribe(connection, "message", ({ args }) => {
+    let messages = yield Mailbox.subscribe(connection, "message")
+
+    messages = messages.map(({ args }) => {
```
cowboyd: I must say that having to re-assign just feels wrong though. what about:

```ts
let messages = (yield Mailbox.subscribe(connection, "message"))
  .map(({ args }) => {
    let [message] = args as IMessage[];
    return { message: JSON.parse(message.utf8Data) };
  });

Somehow that feels much better to me.

PR #216, opened 2020-03-26
Add caching and running tests in parallel to the test workflow

### LINKED ISSUE #119: Paralellize package tests
Right now, tee have an action that runs each test suite one right after the other. However, the test suites are completely independent from each other and so to allow for maximum dev cycles, we should run them in parallel.

### LINKED ISSUE #201: Add caching to yarn install on CI
The runtime of GitHub actions CI is longer than it needs to be, due to always reinstalling from scratch. We should use github action's caching infrastructure to cache the node_modules.

minkimcello:
## Motivation
To resolve #119: Currently, BigTest's test workflow is running the test suites one after the other and making the whole process take very long. This pull request is to modify that workflow to run the test suites in parallel. 

And in the spirit of streamlining the testing workflow, this pull request will also resolve #201.

## Approach
```
name: Test

jobs:
  prepack-upload:
    cache
    yarn install
    yarn prepack
    tar packages/*/dist
    upload tar

  download-test:
    need: prepack-upload
    matrix: [ packages ]
      cache
      download tar && unzip
      yarn install
      cd packages/${{ matrix.package }}
      yarn test
```
- Split the workflow into two jobs (we need to `yarn install` in both jobs because we need it for both `prepack` and `test`):
  - The first job builds and uploads the `dist` directories as an artifact. 
  - The second job downloads that artifact and runs tests for all the individual packages in parallel.
- The workflow uses caching at the beginning of each job right after checkout. If the dependencies listed in `yarn.lock` are the same, it'll be able to load up the previous cache.
- I named all of the workflows (Test, Lint, Preview, Release) so that in the Github checks you'll see `Test / Prepack and Upload` as opposed to `./github/workflows/test / Prepack and Upload`; it'll be much easier on the eyes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,25 +1,54 @@
+name: Test
+
 on:
   pull_request:
     branches:
       - master
 
 jobs:
-  build:
+  prepack-upload:
     runs-on: ubuntu-18.04
-    name: Tests
+    name: Prepack and Upload
     steps:
       - uses: actions/checkout@v1
-      - name: Setup node
-        uses: actions/setup-node@v1
+      - uses: actions/cache@v1
         with:
-          node-version: '12.x'
+          path: /home/runner/.cache/yarn/v6
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+      - name: Install dependencies
+        run: yarn
+      - name: Run prepack
+        run: yarn prepack
+      - name: Tarball prepacked dist
+        run: tar czvf bigtest.dist.tgz ./packages/*/dist
+      - uses: actions/upload-artifact@v1
+        with:
+          name: bigtest.dist.tgz
+          path: bigtest.dist.tgz
+
+  download-test:
+    runs-on: ubuntu-18.04
+    name: ${{ matrix.package }}
+    needs: prepack-upload
+    strategy:
+      matrix:
+        package: [agent, cli, convergence, effection, interactor, logging, parcel, project, server, suite, todomvc]
+    steps:
+      - uses: actions/checkout@v1 
+      - uses: actions/cache@v1
+        with:
+          path: /home/runner/.cache/yarn/v6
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -1,25 +1,53 @@
...
         run: |
           sudo apt-get update
           sudo apt-get install xvfb
       - name: Install dependencies
         run: yarn
-      - name: Prepackage
-        run: yarn prepack
       - name: Run tests
-        run: xvfb-run --auto-servernum -- yarn test --browsers=ChromeHeadless
+        run: cd packages/${{ matrix.package }} && xvfb-run --auto-servernum -- yarn test --browsers=ChromeHeadless
```
taras: use `xvfb-run --auto-servernum -- yarn workspace ${{ matrix.package }} test --browsers=ChromeHeadless` here

.github/workflows/test.yml
```diff
@@ -1,25 +1,53 @@
...
   pull_request:
     branches:
       - master
 
 jobs:
-  build:
+  prepack-upload:
     runs-on: ubuntu-18.04
-    name: Tests
+    name: Prepack and Upload
```
cowboyd: Let's just name this `Prepack` the upload / download is an implementation detail, and may change if GitHub rolls out persistent workspaces

.github/workflows/test.yml
```diff
@@ -1,25 +1,53 @@
...
+        run: yarn prepack
+      - name: Tarball prepacked dist
+        run: tar czvf bigtest.dist.tgz ./packages/*/dist
+      - uses: actions/upload-artifact@v1
         with:
-          node-version: '12.x'
+          name: bigtest.dist.tgz
+          path: bigtest.dist.tgz
+
+  download-test:
```
cowboyd: for the same reason, let's just call this "test" the upload/download portion is an implementation detail, and is explicitly declared in the `needs:` clause.

.github/workflows/test.yml
minkimcello: Will create a separate PR for it.

PR #218, opened 2020-03-30
Optimize agent for orchestrator-controlled result collection

cowboyd:
Motivation
----------

When we began to implement the result collection for tests in the orchestrator, the most optimal way to do it was to immediately create a tree of "result listeners" for each test, step and result in the entire hiearchy. Because of this, events like 'test:running' are purely informational, and events like 'test:done' are completely unecessary.

Approach
----------
This PR removes `test:done` entirely, and separates out the `run:begin` and `run:end` to fire a `lane:begin` and `lane:end` event from the harness. These are purely informational and not needed except for perhaps some debugging purposes.

### DIFF
--- a/packages/agent/app/create-harness.ts
+++ b/packages/agent/app/create-harness.ts
@@ -16,7 +16,12 @@ export function* createHarness() {
     if(message.type === 'run') {
       let manifest: TestImplementation = yield loadManifest(message.manifestUrl);
 
-      yield runTest(parentFrame, manifest, message.path.slice(1))
+      try {
+        parentFrame.send({ type: 'lane:begin' });
+        yield runTest(parentFrame, manifest, message.path.slice(1))
+      } finally {
+        parentFrame.send({ type: 'lane:end' });
+      }
     }
   }
 }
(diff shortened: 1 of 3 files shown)


packages/agent/app/create-harness.ts
```diff
@@ -16,7 +16,12 @@ export function* createHarness() {
     if(message.type === 'run') {
       let manifest: TestImplementation = yield loadManifest(message.manifestUrl);
 
-      yield runTest(parentFrame, manifest, message.path.slice(1))
+      try {
+        parentFrame.send({ type: 'lane:begin' });
```
jnicklas: Might be worth sending along the path of the lane we are running here?

cowboyd: Added.... indeed, it wouldn't be that useful of an event without it, would it 😆

packages/agent/app/create-harness.ts
```diff
@@ -15,12 +15,12 @@ export function* createHarness() {
 
     if(message.type === 'run') {
       let manifest: TestImplementation = yield loadManifest(message.manifestUrl);
-
+      let path = message.path.slice(1);
```
jnicklas, line 18: Why the slice?

PR #220, opened 2020-03-31
New Interactor API spike

pittst3r:
Documented in [this RFC](https://github.com/thefrontside/bigtest/pull/109), this new Interactor API is a spike to flesh out new ideas.


packages/interactor/src/when.ts
```diff
@@ -0,0 +1,51 @@
...
+      result = await proc();
+      success = true;
+    } catch (e) {
+      error = e;
+      await new Promise(resolve => {
+        setTimeout(async () => {
+          await attempt();
+          resolve();
+        }, 10);
+      });
```
jnicklas: I'm wondering if an iterative solution wouldn't be easier to understand, rather than using recursion? I think if we have some helper function `sleep(x)`, which just returns a promise which resolves in x milliseconds, then we can write it like this:

``` javascript
while(true) {
  await sleep(10);
  if (elapsedSince(start) >= timeout) { break }
  // ...
}
```

packages/interactor/src/selector.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type Selector<Container, Result> = (locator: string, container: Container) => Promise<Result[]>;
+
+export function selector<Container, Result>(query: Query<Container, Result>): Selector<Container, Result> {
+  return (locator, container) =>
+    when(async () =>
+      throwIfEmpty(
+        Array.from(await query(locator, container)),
+        `Did not find any matches with locator "${locator}"`
+      )
```
jnicklas: This is fine as a first implementation, but based on my experience from Capybara, I'm thinking that this will probably be too lax, eventually. Capybara gives you the option of specifying the conditions [when a selector is allowed to proceed](https://github.com/teamcapybara/capybara#strategy). And by default it is not allowed to have more than one matching element. This is kind of important for making tests more solid, since sometimes you might have several matching elements on the page, and it should be deterministic which of those you click on.

jnicklas:
So I'm not entirely sure how the implementation of interactor works at the moment, since it's a bit complicated, and I'm not entirely sure that I understand all of it, but one possible problem area that I'd like to flag is that it looks to me like the resolution of elements is eager, and I think this could lead to problems. Ideally, resolution of elements should be deferred up until the very last second, right before we are interacting with the element.

Maybe I'm just misunderstanding the implementation, so if you've already considered this, just ignore this comment ;)

Given that we are running in the page's run-loop, we have the advantage compared to Capybara that the element cannot disappear from the page while we are interacting with it, but this only works if we do the resolution of the element (and all of its ancestors!) directly before the interaction. That is they must happen during the same tick of the event loop!

Imagine some pseudo test code like this:

``` javascript
let someAreaOfPage = Element("div#some-area");
await sleep(10) // #some-area disappears and reappears here
let button = Button("My Button", someAreaOfPage); // uses old #some-area
```

But this is also a problem for the actions themselves, for example one of the tests does something like this:

``` javascript
        async fill(val: string) {
          const elem = await subject.first;
          await sleep(200);
          elem.value = val;
        }
```

This is problematic, because `elem` could actually disappear while the `sleep` is happening.

One possible workaround to this is to not allow actions to be async functions, so that they cannot introduce any race conditions like this.

packages/interactor/src/when.ts
cowboyd: Awhile back, I made a prototype based on async iterators that provided a pretty elegant approach https://github.com/cowboyd/convergence/blob/master/src/converge.js

pittst3r:
@jnicklas Thanks for the great feedback! I will need to get my head back into it to properly answer your questions.

I agree that the current implementation is complicated and dense. In addition to ensuring we're lazily selecting elements we should make it easier to understand.

jnicklas:
This has been superseded by #301

PR #221, opened 2020-03-31
Report test results through the GraphQL API.

cowboyd:
> depends on #223 #224 

Motivation
----------

While the agent supports running tests and streaming results back to the server, we don't currently have a way to interface with those results via the GraphQL API. In our planned scenario, you will issue a `run` mutation and then subscribe to the results of that run.

Approach
----------

Whenever a run command is issued to the command processor, we fork an _entire effection tree_ corresponding to the test hierarchy. This allows us to set up a process to track _each_ result, be it step, test, or assertion. If a step fails, we throw an exception which causes every subsequent step and assertion to be disregarded.

Screenshots
------------

### DIFF
--- a/packages/server/test/run-tests.test.ts
+++ b/packages/server/test/run-tests.test.ts
@@ -1,0 +1,148 @@
+import * as expect from 'expect';
+import { Agent, Command } from '@bigtest/agent';
+import { TestResult } from '@bigtest/suite';
+import { actions, Subscription } from './helpers';
+
+describe('running tests on an agent', () => {
+  let agent: Agent;
+
+  beforeEach(async () => {
+    await actions.startOrchestrator();
+    agent = await actions.createAgent();
+
+    agent.send({
+      type: 'connected',
+      data: {
+        platform: {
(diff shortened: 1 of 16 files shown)


packages/server/test/run-tests.test.ts
```diff
@@ -0,0 +1,148 @@
...
+            stack: ['here.js', 'there.js']
+          }
+        })
+      });
+
+      it('marks that step as failed', async () => {
+        await actions.fork(subscription.until(run => {
+          return run.tree.children
+            .find(child => child.description === "Signing In" ).steps
+            .find(child => child.description === "when I fill in the login form").status === 'failed';
```
jnicklas, line 124: Maybe this is premature, but would it make sense to add some helper function, at least in the tests, to find some test by path? So something like `testByPath(run.tree, ["Signing in", "when I fill in the login form"])`. I guess this could be problematic if there is ambiguity between steps/assertions/children. What do you think?

jnicklas:
😍 It's amazing!

PR #224, opened 2020-04-03
♻Extract Agent library into re-useable library

cowboyd:
Motivation
---------- 

We want to write big tests for the bigtest orchestrator. That means only interfacting with it in the same way as external process will be interfacing with it.

That mean, only using the graphql API to query and initiate state changes and to only use external processes communicating over wesockets to represent the agent.

Approach
----------

This refactors the agent code to extract the actual portion concerned with communicating with the orchestrator into an implementation that can be shared between the web agent, and any other javascript environment. That way, we can just import an `Agent` into our tests, and connect it to the server under test.

From the README:

``` typescript
 import { main } from 'effection';
 import { Agent, Command } from '@bigtest/agent';

 main(funtion*() {

   let socket = WebSocket('http://localhost:1234');
   let agent = yield Agent.start(socket);

   // let the orchestrator know that you're here!
   agent.send({
     type: 'connected',
     data: 'secret agent'
   });

   // await for commands from the orchestrator
   while (true) {
     let command: Command = yield agent.receive();

     // the orchestrator wants us to do something. let's do it!
     yield handleCommand(command);
   }
 });
 ```

The protocol of allowable messages is implemented using a collection of TypeScript interface.

### CODE REVIEW

File: packages/agent/shared/agent.ts [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+    let mailbox = new Mailbox();
+
+    yield suspend(subscribe(mailbox, socket, ['open', 'message', 'close', 'error']));
+    yield suspend(monitor(function*() {
+      let { args: [error] } = yield mailbox.receive({ event: 'error' });
+      throw error as Error;
+    }));
+    yield mailbox.receive({ event: 'open' });
+    yield suspend(ensure(() => socket.close()));
+
```
jnicklas, line 20: So itching to get rid of all of these suspends 😬

jnicklas:
Looks great! Eventually we might want to split all of these parts into separate packages, but let's not go crazy with that now. But it might become relevant once we start writing non-web agents.

jnicklas:
Saw something which I think is worth fixing after I'd already approved and merged 😅 It feels a bit odd to me that the agent does *not* create the websocket, but does actually add an `ensure` handler to it, that feels a bit strange. It effectively adds a handler to an object it doesn't "own".

PR #225, opened 2020-04-07
Replace suspend with effection resources

jnicklas:
Effection now supports resource objects on an experimental branch, this makes it possible to tie the lifetime of a context to an object, and return it out of its context to pass ownership of the context to the parent. This provides an alternate escape hatch to using suspend, which is more disciplined and composes better.

See: https://github.com/thefrontside/effection.js/pull/89


jnicklas:
So this PR has turned into a bit of a monster 😬 The main reason for this is that patterns that we've previously used simply don't work anymore after the resources PR to effection. `suspend` obviously doesn't work anymore, but also the `parent` and `self` patterns that we've used are broken, because they work by returning a context object, which now does some other stuff. So I had to rewrite a lot of stuff into this new style so that it could actually utilize resources instead of relying on accessing `context.parent` in various ways.

Maybe some of these refactorings could actually be moved out of this PR if it is too daunting to review.

Overall, using resources feels very nice to me, it's intuitive in a way which `suspend` is absolutely not.

jnicklas:
Also, I went with adding a `setTimeout` to force the emit on `send` on `Mailbox` to run in a different tick. This is not a great solution at all, but I felt like I've already spent way too much time on this.

jnicklas:
OMG IT'S GREEN! 🍀

PR #230, opened 2020-04-15
Inverted subscription

jnicklas:
Depends on #225 

Currently a client subscription uses a form of internal iteration like this:

``` javascript
client.subscribe("{ foo }", function*(data) {
  // do something
});
```

This changes the subscription to use external iteration instead, like this:

``` javascript
let subscription = yield client.subscribe("{ foo }");

while(true) {
  let data = yield subscription.receive();
  // do something
}
```

The major advantage to this is that we can at some point decide we are done, whereas we have no good way of doing this with internal iteration. With internal iteration we have to surreptitiously obtain a handle to the current ExecutionContext and halt it, which is pretty gross.

We will need this for the CLI, since a CLI command will fire off a subscription, but must be able to decide that it is now done at some point.


packages/server/src/client.ts
```diff
@@ -17,13 +17,12 @@ export class Client {
 
     let client = new Client(socket);
     let res = yield resource(client, function*() {
-      yield ensure(() => {
-        socket.close();
-      });
-      yield monitorErrors(socket);
+      yield ensure(() => socket.close());
```
cowboyd, line 20: In order to harmonize with, and make transition to future syntaxes easier, shouldn't we use `try/finally` wherever possible? In other words,

```ts
try {
  yield;
} finally {
  socket.close();
}
```

jnicklas: I think that `ensure` as an operation is fine from a future compatibility POV, since it's pretty trivial to implement it, even without control functions:

```
function ensure(fn) {
  return spawn(function*() {
    try {
      yield;
    } finally {
      fn();
    }
  });
}
```

From a philosophical perspective we might want to avoid it though, not sure about that.

cowboyd: It's definitely awesome that for compatibility it doesn't make any difference because it's an operation. 😄 

I'm mostly thinking from the perspective of maintaining an API that can be accomplished without any API and having to justify that in the documentation and what weight this abstraction carries.

For me, the closest thing that I come up with is that `try` blocks are cumbersome and add a lot of noise (two extra lines, some not very useful rightward drift, and often a bit of fiddling with lexical bindings to make sure things are in scope when they should be). In fact, I have often wished that JavaScript supported Ruby-style bare `finally` blocks (or implicit `try`) for all methods, so that could you say:

```javascript
function* doSomething() {
  let resource = yield allocate();
  return yield resource.operation();
  catch(error) {
    console.error(error);
  } finally {
    resource.destroy();
  }
}
```

When I want to reach for `ensure`, the urge feels like it comes from wanting to avoid that `try/catch` noise. So that we can emulate something like the above with:

```javascript
function* doSomething() {
  let resource = yield allocate();
  yield rescue(error => console.error(error));
  yield ensure(() => resource.destroy());

  return yield resource.operation();
}
```

That is more appealing to me than the previous version, but it also raises questions about semantics like what is the order of multiple ensure operations? What if I want to do asynchrony? Maybe `rescue` isn't possible... why not? On the other hand, if there are no semantics but JavaScript semantics, then we don't have to answer any of those questions.

I'm still on the fence, but these are the two major lines of thought in my mind.

I'll leave it up to you.... Somewhat relevant in the long-term, but not in the short-term: it seems there is proposal for explicit resource management [being considered for inclusion in the language][1] that would render the need for a `finally` blocks mostly moot.

jnicklas: Yeah, I agree, `try/catch` is a pretty noisy syntax, and it doesn't really add any clarity in this case, so I think that I'd probably prefer `ensure` as an operation in cases like this, but I think it's mostly a matter of personal preference.

jnicklas:
🚢

packages/server/src/client.ts
cowboyd: I just realized I left a broken link to the the [explicit resource management proposal](https://github.com/tc39/proposal-explicit-resource-management). I've updated the comment and dropped it here too. It's definitely poking at some of the same problems we are.

PR #231, opened 2020-04-15
Simplify tsconfig.json setup for the entire project

taras:
## Motivation

I'm working on refactoring cli-mob branch to use Atom. In the process, I came across the config.dist.json pattern that we've been using. I found it thoroughly confusing because we have a tsconfig.json that includes tests and a special tsconfig for building package. This is a double negative equivalent of implementation. After some experimentation, it turned out that we don't need to include information about tests in tsconfig.json because `-r ts-node/register` takes care of whats necessary for tests. 

## Approach

Removed all test specific changes in tsconfg.json and deleted all of the tsconfig.dist.json files. 

## TODO

- Move these configs to their own package and consume them from that package

### DIFF
--- a/packages/agent/package.json
+++ b/packages/agent/package.json
@@ -35,10 +35,10 @@
     "playwright": "^0.11.1",
     "regenerator-runtime": "^0.13.3",
     "ts-node": "*",
-    "typescript": "^3.7.0",
-    "@bigtest/effection": "*"
+    "typescript": "^3.7.0"
   },
   "dependencies": {
+    "@bigtest/effection": "*",
     "bowser": "^2.9.0",
     "effection": "^0.6.0",
     "express": "^4.17.1",
--- a/packages/agent/package.json
+++ b/packages/agent/package.json
@@ -17,7 +17,7 @@
     "mocha": "mocha -r ts-node/register",
     "start": "parcel serve --out-dir dist/app app/index.html app/harness.ts",
     "bundle": "parcel watch --out-dir dist/app app/index.html app/harness.ts",
-    "prepack": "tsc --outdir dist --project tsconfig.dist.json --declaration --sourcemap && tsc --project tsconfig.app.json --noEmit && parcel build --no-minify --out-dir dist/app app/index.html app/harness.ts",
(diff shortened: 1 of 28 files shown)


packages/agent/package.json
```diff
@@ -13,10 +13,10 @@
   ],
   "scripts": {
     "lint": "eslint '{app,bin,src,test}/**/*.ts'",
-    "test": "mocha -r ts-node/register test/**/*.test.ts",
+    "test": "TS_NODE_PROJECT=tsconfig.test.json mocha -r ts-node/register test/**/*.test.ts",
```
cowboyd, line 16: The `yarn test` command works just the same for me if we don't include the `TS_NODE_PROJECT` variable. Seems like this is necessary?

taras: @cowboyd it turns out that we don't need any of this. I removed TS_NODE_PROJECT and all of tsconfig.test.json files and tests are running green.

taras:
> Can we just not have tsconfig.test.json and it works?

@cowboyd yes

packages/agent/package.json
```diff
@@ -35,7 +35,8 @@
     "playwright": "^0.11.1",
     "regenerator-runtime": "^0.13.3",
     "ts-node": "*",
-    "typescript": "^3.7.0"
+    "typescript": "^3.7.0",
+    "@bigtest/effection": "*"
```
cowboyd, line 39: `@bigtest/effection` is a runtime dependency

taras:
> It had the advantage that we could treat @effection/node and @effection/events as their own packages that were psychologically cordoned off from the rest of the bigtest codebase, but in the future, I probably would not do it because of nagging paper cuts like this.

@cowboyd we have an effection multipackage repo now. Should we move everything over there and end the practice of using module aliases this way?

cowboyd:
> @cowboyd we have an effection multipackage repo now. Should we move everything over there and end the practice of using module aliases this way?

@taras We've already removed module aliases (reference was to a historical commit), so I think it's safe to proceed by just getting rid of `tsconfig.dist.json` entirely.

The only think is that this PR introduces a devDependency on `@bigtest/effection`, but I believe this is a runtime dependency.

packages/agent/package.json
taras: Done

PR #236, opened 2020-04-16
Move Atom class into it's own package

taras:
## Motivation

I'm continuing to work on the focus management in cli-mob. I need a better mechanism for tracking state that allows to update child focusable elements without capturing closures provided by hooks in focusable component. Atom should fit the bill nicely but it's in @bigtest/server which is not exported. 

## Approach

1. Created package called @bigtest/atom
2. Moved Atom & State classes into that package
3. Updated @bigtest/server to use @bigtest/atom
4. Updated @bigtest/server to export the same types as previously to reduce the amount of necessary changes. This also makes it a little more convenient to use State type in @bigtest/server


packages/server/src/orchestrator/atom.ts
```diff
@@ -1,106 +1,20 @@
...
-    let mailbox = yield Mailbox.subscribe(this.subscriptions, "state");
-
-    while (true) {
-      let { args: [state] } = yield mailbox.receive();
-      yield fn(state);
-    }
+import { Atom as _Atom, Slice as _Slice } from "@bigtest/atom";
+import { OrchestratorState } from "./state";
+
+export class Atom extends _Atom<OrchestratorState> {
```
cowboyd, line 4: Can we do this without extending atom? Maybe by having a function `OrchestratorAtom` that returns `Atom<OrchestratorState>`?

jnicklas:
Huge +1 on making Atom generic over the state it contains. ❤️

packages/server/src/orchestrator/atom.ts
taras: @cowboyd like this?

```ts
export const OrchestratorAtom = () => new Atom({
  manifest: {
    description: "None",
    fileName: "<init>",
    steps: [],
    assertions: [],
    children: [],
  },
  agents: {},
  testRuns: {},
});
```

cowboyd: Pretty much, although you might need to explicitly mark it so that TS can figure out that `S` === `OrchestratorState`

```ts
function OrchestratorAtom(): Atom<OrchestratorState> {
  return new Atom({
   manifest: {
    description: "None",
    fileName: "<init>",
    steps: [],
    assertions: [],
    children: [],
  },
  agents: {},
  testRuns: {},
  });
}
```

taras: @cowboyd done

packages/server/src/orchestrator/atom.ts
```diff
@@ -1,20 +1,18 @@
...
+    fileName: "<init>",
+    steps: [],
+    assertions: [],
+    children: [],
+  },
+  agents: {},
+  testRuns: {},
+});
+
+export type Atom = _Atom<OrchestratorState>
```
cowboyd, line 16: What are we gaining by aliasing `Atom` and `Slice` this way?

taras: There are a bunch of places where Atom is referenced. I would need to go through and change it all to `Atom<OrchestratorState>` this allows me to no do that. Do you want me to?

cowboyd: I don't know what the best practice for TS, but it just seemed like something I've never seen before. I don't know if this is a code-smell or not?

taras: Not sure... I can apply `Atom<OrchestratorState>` to everything. It's 10 places. I think

cowboyd: I have a slight preference for being explicit and not aliasing, but I don't have a strong opinion, so I'll leave  up to you.

taras: Removed the alias

PR #238, opened 2020-04-17
Remove convergence and interactor from monorepo

minkimcello:
## Motivation
Just following orders. 👮 

Interactor and convergence packages are outdated and no longer being used so we're removing them for tidying up purposes.

## Approach
Removed `interactor` and `convergence` from the packages directory and updated `test` + `release` workflows.


.github/workflows/release.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       with:
         before_all: yarn prepack
         npm_publish: yarn publish
-        ignore: packages/website packages/convergence packages/interactor packages/cli
+        ignore: packages/website packages/cli
```
cowboyd, line 19: Why are we ignoring `cli`?

minkimcello: Hmmm I'm not sure. I was wondering about that too. I checked to see who added it and turns out I added it from the get go ([PR 132](https://github.com/thefrontside/bigtest/pull/132/)). 🤷 

I'll go ahead and remove it.

taras:
@cowboyd should we add a mechanism to release action to somehow deprecate packages?

cowboyd:
> @cowboyd should we add a mechanism to release action to somehow deprecate packages?

That would actually be awesome.... how do you deprecate an NPM package?

taras:
@cowboyd https://docs.npmjs.com/cli/deprecate

cowboyd:
Right. In that case, what about adding a `deprecated` flag to the package.json and then we have our action synchronize the deprecation state with.

minkimcello:
Moved discussion to [this](https://github.com/thefrontside/actions/issues/57) issue.

PR #240, opened 2020-04-17
Unifying typescript configurations

### LINKED ISSUE #234: Unify all TypeScript config in @bigtest/tsconfig package
We want to apply the same TypeScript config to all projects in BigTest repository. There might be some exceptions but that should be the general rule. After #231 is merged, we should create @bigtest/tsconfig package, move config in there and update all packages to get tsconfig from this new package. 

## TODO

- [ ] @bigtest/tsconfig should use strict mode which enables all of these flags: `--noImplicitAny, --noImplicitThis, --alwaysStrict, --strictBindCallApply, --strictNullChecks, --strictFunctionTypes and --strictPropertyInitialization`

minkimcello:
## Motivation
Many packages = many typescript configurations. So the plan is the unify all the configurations and put it inside `@bigtest/tsconfig` package. This resolves #234.

## Approach
- Skipped `convergence` and `interactor` because they're on they way [out](https://github.com/thefrontside/bigtest/pull/238).
- Skipped `website` too.
- Put most of the overlapping `tsconfig` of packages and put them inside `packages/tsconfig/tsconfig.json`.
- Using strict mode on all of them so now we have a whole batch of errors to go over.

### DIFF
--- a/packages/agent/app/create-harness.ts
+++ b/packages/agent/app/create-harness.ts
@@ -1,5 +1,7 @@
 import { ParentFrame } from './parent-frame';
-import { TestImplementation } from '@bigtest/suite';
+import { TestImplementation, ErrorDetails } from '@bigtest/suite';
+import { Operation } from 'effection';
+import { Deferred } from '@bigtest/effection';
 
 declare const __bigtestManifest: TestImplementation;
 
@@ -26,11 +28,15 @@ export function* createHarness() {
   }
 }
 
-function serializeError({ message, fileName, lineNumber, columnNumber, stack }) {
-  return { message, fileName, lineNumber, columnNumber, stack };
-}
+const serializeError: (error: ErrorDetails) => ErrorDetails = ({ message, fileName, lineNumber, columnNumber, stack }) => ({
+  message,
+  fileName,
+  lineNumber,
+  columnNumber,
+  stack
+});
 
-function *runTest(parentFrame: ParentFrame, test: TestImplementation, path: string[], prefix: string[] = []) {
+function *runTest(parentFrame: ParentFrame, test: TestImplementation, path: string[], prefix: string[] = []): Operation<void> {
   let currentPath = prefix.concat(test.description);
 
   console.debug('[harness] running test', currentPath);
@@ -75,17 +81,19 @@ function *runTest(parentFrame: ParentFrame, test: TestImplementation, path: stri
   }
 }
 
-function loadManifest(manifestUrl: string) {
-  return ({ resume, ensure }) => {
-    let scriptElement = document.createElement('script') as HTMLScriptElement;
-    let listener = () => {
-      resume(__bigtestManifest);
-    }
+function* loadManifest(manifestUrl: string): Operation<TestImplementation> {
+  let { resolve, promise } = Deferred<TestImplementation>();
 
-    scriptElement.addEventListener('load', listener)
-    ensure(() => { scriptElement.removeEventListener('load', listener) });
+  let scriptElement = document.createElement('script') as HTMLScriptElement;
 
+  let listener = () => resolve(__bigtestManifest)
+
+  try {
+    scriptElement.addEventListener('load', listener);
     scriptElement.src = manifestUrl;
     document.head.appendChild(scriptElement);
+    return yield promise;
+  } finally {
+    scriptElement.removeEventListener('load', listener);
   }
 }
--- a/packages/agent/test/agent.test.ts
+++ b/packages/agent/test/agent.test.ts
@@ -1,16 +1,18 @@
 import { Local, WebDriver } from '@bigtest/webdriver';
 
-import { main, Operation, Context, resource } from 'effection';
+import { resource } from 'effection';
 import { describe, it } from 'mocha';
 import * as expect from 'expect';
 import * as express from 'express';
 import fetch from 'node-fetch';
 import * as fixtureManifest from './fixtures/manifest.src';
 
-import { AgentConnectionServer, AgentServer } from '../src/index';
+import { AgentConnectionServer, AgentServer, AssertionResult } from '../src/index';
 
 import { Mailbox, ensure } from '@bigtest/effection';
 import { throwOnErrorEvent, once } from '@effection/events';
+
+import { spawn } from './helpers';
 
 function* staticServer(port: number) {
   let app = express().use(express.static("./test/fixtures"));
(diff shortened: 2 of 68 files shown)


packages/effection/test/mailbox.test.ts
```diff
@@ -181,11 +181,11 @@ describe("Mailbox", () => {
       spawn(function*() {
         yield source.pipe(destination);
         yield;
-      });
+      } as OperationFn<unknown>);
```
cowboyd, line 184: Let’s use operation where we can. OperationFn is also an Operation and it will be deprecated.

Also, Return type is not unknown, it is void

packages/agent/test/agent.test.ts
```diff
@@ -31,14 +30,18 @@ describe("@bigtest/agent", function() {
...
   function spawn<T>(operation: Operation): Promise<T> {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
     return World["spawn"](operation);
   }
 
   function launch(browserType: BrowserType, options = {}): Promise<Browser> {
     return spawn(({ resume, fail, context: { parent } }) => {
       browserType.launch(options)
         .then(browser => {
```
cowboyd, line 41: This is easily rewritten as a generator and would improve typing.

packages/agent/test/agent.test.ts
```diff
@@ -119,8 +122,13 @@ describe("@bigtest/agent", function() {
       });
 
       describe('sending a run message', () => {
-        let success;
-        let failure;
+        interface Result {
+          status: string;
```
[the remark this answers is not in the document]
taras: Where should it go?

packages/effection/src/monitor-errors.ts
```diff
@@ -3,7 +3,7 @@ import { Operation, monitor } from 'effection';
 import { once } from './events';
 
 export function monitorErrors(source: EventSource): Operation {
-  return monitor(function*() {
+  return monitor<unknown>(function*() {
```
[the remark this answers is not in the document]
taras: `never` doesn't work.

image

packages/agent/test/agent.test.ts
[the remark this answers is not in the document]
cowboyd: Sorry, what I meant to say was that it already exists in `@bigtest/suite` https://github.com/thefrontside/bigtest/blob/master/packages/suite/index.ts#L80-L102

We should be able to import them from there.

packages/effection/src/monitor-errors.ts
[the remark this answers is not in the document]
cowboyd: Hmmm. I think the problem is with the effection typings, for now, can we use `monitor<void>`?

packages/agent/test/agent.test.ts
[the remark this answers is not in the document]
taras: @cowboyd none of those types have error property

packages/effection/src/monitor-errors.ts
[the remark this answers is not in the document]
taras: yeah, it works

packages/agent/test/agent.test.ts
taras: Can you help me rewrite this?

packages/agent/app/create-harness.ts
```diff
@@ -26,11 +28,15 @@ export function* createHarness() {
...
+const serializeError: (error: ErrorDetails) => ErrorDetails = ({ message, fileName, lineNumber, columnNumber, stack }) => ({ 
+  message, 
+  fileName, 
+  lineNumber, 
+  columnNumber, 
+  stack 
+});
 
-function *runTest(parentFrame: ParentFrame, test: TestImplementation, path: string[], prefix: string[] = []) {
+function *runTest(parentFrame: ParentFrame, test: TestImplementation, path: string[], prefix: string[] = []): Generator {
```
cowboyd, line 39:
```suggestion
function *runTest(parentFrame: ParentFrame, test: TestImplementation, path: string[], prefix: string[] = []): Operation<void> {
```

packages/agent/test/agent.test.ts
```diff
@@ -31,14 +31,18 @@ describe("@bigtest/agent", function() {
 
   let World: Context;
   function spawn<T>(operation: Operation): Promise<T> {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
     return World["spawn"](operation);
```
cowboyd, line 36: We figured out the way to use `World.spawn` right?

packages/agent/test/agent.test.ts
cowboyd: I honestly don't understand why I tried to be clever in the first place
```typescript
  function launch(browserType: BrowserType, options = {}): Promise<Browser> {
    return spawn(function*() {
      let browser: Browser = yield browserType.launch(options);
      try {
        yield;
      } finally {
        browser.close();
      }
    });
  }
```

packages/effection/test/helpers.ts
```diff
@@ -6,6 +6,8 @@ class World {
   context: Context = main(undefined);
 
   spawn<T>(operation: Operation): Context & PromiseLike<T>{
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
     return this.context['spawn'](operation);
```
cowboyd, line 11: We figured out a way around this?

packages/server/src/command-server.ts
```diff
@@ -48,13 +48,22 @@ let testRunIds = (function * () {
   }
 })()
 
+interface RootValue {
```
cowboyd, line 51: Where are we using this interface?

taras: Removed

packages/effection/test/helpers.ts
```diff
@@ -14,7 +16,7 @@ class World {
   }
 }
 
-export function spawn<T>(operation: Operation): Context & PromiseLike<T> {
+export function spawn<T>(operation: Operation<T>): Context & PromiseLike<T> {
```
cowboyd, line 19: `Context` should already _be_ PromiseLike, so I think it's redundant?

cowboyd:
Not super sure why the agent tests are failing, but it just so happens that we're in the process of [removing the dependency on playwright][1] which might provide a fix. Still, this has never been flaky before 🤔 

[1]: https://github.com/thefrontside/bigtest/pull/251

PR #241, opened 2020-04-20
Test results should be per-agent

jnicklas:
Currently when we generate the state for a particular test run, we assume that the test run only has one agent, we did this to simplify things initially, but eventually we want to allow a single test run to run against multiple agents. This PR allows us to track results per agent.

### DIFF
--- a/packages/agent/shared/protocol.ts
+++ b/packages/agent/shared/protocol.ts
@@ -8,6 +8,7 @@ export interface AgentProtocol {
 
 interface Connect {
   type: 'connected';
+  agentId?: string;
   data: unknown;
 }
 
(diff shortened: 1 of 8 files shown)


jnicklas:
@cowboyd @pittst3r I'm having some problems on the nexus types with this. The tests pass, but I get an error like this when running prepack:

```
yarn prepack                                                                                                                                                                                                                        ✓  10058  13:13:19
yarn run v1.21.1
$ ts-node ./src/schema.ts && tsc --outdir dist --module commonjs --sourcemap --declaration
src/schema.ts:168:55 - error TS2339: Property 'agents' does not exist on type '{ status: string; testRunId: string; }'.

168           resolve: (testRun) => Object.values(testRun.agents)
                                                          ~~~~~~

src/schema.ts:175:49 - error TS2339: Property 'agents' does not exist on type '{ status: string; testRunId: string; }'.

175           resolve: (testRun, { id }) => testRun.agents[id]
                                                    ~~~~~~


Found 2 errors.
```

It seems the generated file does not have any of the typings for `agents` and `agent`. Not sure how to wire this all up.

cowboyd:
Hmmm. It looks like it’s typing the `testRun` parameter incorrectly? It has a type which looks close in that it has the status and Id. I wonder where it is getting those fields from?

jnicklas:
@cowboyd I integrated your changes from #248 and it seems to work well. Maybe we should even add these typings for all of the schema object types? We could do that in a separate PR.

PR #245, opened 2020-04-21
Mute warning about event listener leak

jnicklas:
In the command processor, we subscribe to the same mailbox many, many times. The underlying EventEmitter complains if there are more than 10 listeners, since it suspects a memory leak. However, in our case, this is expected behaviour, so we should just increase this limit.


jnicklas:
I opted to expose the same API that EventEmitter exposes, rather than increasing this limit everywhere. `10` does seem awfully low, maybe we should at least have a somewhat higher limit on mailboxes by default?

PR #246, opened 2020-04-21
Only generate schema artifacts explicitly

### LINKED ISSUE #244: Loading of context.ts fails in compiled package
Since #243, we are pulling in the file `src/schema/context.ts` by specifically loading it here:

https://github.com/thefrontside/bigtest/blob/f653e7ac901df894f9f2cb5200225cc68e2f9299/packages/server/src/schema.ts#L15

However, this does not work when running the server as a compiled package, since the file now has the extension `.js` instead of `.ts`.

This generates a warning like this:

```
GraphQL Nexus: Unable to find file or module /Users/jonasnicklas/Projects/bigtest/packages/server/dist/src/schema/context.ts, skipping
```

cowboyd:
> closes #244 

Motivation
-----------
By default, nexus _always_ tries to generate schema artifacts like the `.graphql` schemafile and the typings. However, [this fails in production][1] because it tries to load typescript modules which have long since been compiled into javascript and are not shipped with the distribution.

Approach
----------
Nexus provides a flag to prevent schema generation presumably for cases precisely such as this one. We use this flag here to stop generation unless the `BIGTEST_GENERATE_SCHEMA` environment variable is explicitly set which is done only from the `yarn generate:schema` step.

While this feels a bit awkward and hackish, it's the only workaround that I've been able to find.

[1]: https://github.com/thefrontside/bigtest/issues/244

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -15,8 +15,8 @@
     "test": "mocha ./test/setup test/**/*.test.ts",
     "mocha": "mocha ./test/setup",
     "test:app:start": "ts-node ./bin/todomvc.ts",
-    "generate:schema": "ts-node ./src/schema.ts",
-    "prepack": "ts-node ./src/schema.ts && tsc --outdir dist --module commonjs --sourcemap --declaration"
+    "generate:schema": "BIGTEST_GENERATE_SCHEMA=true ts-node ./src/schema.ts",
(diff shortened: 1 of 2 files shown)


cowboyd:
@taras that's the flag we're using, just using an environment variable to control it

https://github.com/thefrontside/bigtest/pull/246/files#diff-ea36f5197e724d91cad6300896d08289R22

Is there a better way to use the flag?

taras:
@cowboyd oops :)

PR #248, opened 2020-04-21
Calibrate and fix typings in test run for schema

cowboyd:
It appears that nexus determines the type of the "source" object in its graphql resolvers by default as the union of all the "primitive" fields in type definition. So in the case of "TestRun", it sees that it has an `id("testRunId")` and `string("status")` fields, and expects the source value to be just an object with the `{ testRunId, status }` structure. This was causing errors, because in reality we have a`TestRunState` off of which we want to get the `agents` field.

You can set the type of the source object in a resolver using the `rootTyping` field, so this change sets the root typing for a test run to `TestRunState`. Upon setting this, it caught a type error, where we were not returning the `testRunId` for a `TestRunAgent`. This also fixes that type error.

> Note: I tried to use `{...agent, testRunId: testRun.testRunId }` to assign the test run id, but it caused a type error "Typescript: Spread types may only be created from object types" which felt wrong, but 🤷🏼

### DIFF
--- a/packages/agent/shared/protocol.ts
+++ b/packages/agent/shared/protocol.ts
@@ -8,6 +8,7 @@ export interface AgentProtocol {
 
 interface Connect {
   type: 'connected';
+  agentId?: string;
   data: unknown;
 }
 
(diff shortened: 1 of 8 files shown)


jnicklas:
I think `TestRunAgent` having a `testRunId` is actually a mistake.

cowboyd:
Cool! I mean not cool that it was a mistake, but maybe the typings are kinda working? Feel free to close this PR if you want to just integrate the typing for TestRun

PR #251, opened 2020-04-24
Replace playwright with selenium webdriver

cowboyd:
Motivation
-----------

We want the ability to launch agents automatically as part of starting a BigTest run. In other words, running:

```
$ bigtest --browser firefox
```

Will automatically start the Firefox browser for you, direct it to the agent server to load the agent application, and then have that agent connect back to the orchestrator. In order to do this, we need a quick way of launching browsers.

Approach
----------

This is the first step in accomplishing this by creating a way to automatically launch browsers that are automatically connected as agents via the agent server. To test this, we use this technique in testing the agent app itself. In order to launch an agent from a running agent server:

```typescript
let driver: WebDriver = yield agentServer.launch({
  connectBackURL: 'ws://bigtest.orchestrator.com:3000',
  headless: false,
  browser: 'chrome'
});
```

`driver` is an effection resource that governs the scope of the launched browser.

A major benefit of this is that we can expand this to support anything that implements the w3c Webdriver protocol such as remote browsers like those on BrowserStack and SauceLabs.

Finally, a hidden benefit is  that this replaces playwright which was a very unwieldy development tool whose 10 minute yarn installs were a never-ending source of annoyance.

Notes
------

* chrome driver is *much* faster to launch than gecko driver, but appears to be more flaky when tearing down and setting up repeatedly, which is why we're using it in tests.

### DIFF
--- a/packages/agent/src/server.ts
+++ b/packages/agent/src/server.ts
@@ -4,9 +4,21 @@ import * as Path from 'path';
 import { Server } from 'http';
 import { ensure } from '@bigtest/effection';
 
+import 'chromedriver';
+import 'geckodriver';
+import { WebDriver, Builder } from 'selenium-webdriver';
+import { Options as ChromeOptions } from 'selenium-webdriver/chrome'
+import { Options as FirefoxOptions } from 'selenium-webdriver/firefox'
+
 interface Options {
   port: number;
   externalURL?: string;
+}
+
+interface LaunchOptions {
+  browser: 'chrome' | 'firefox';
+  connectBackURL: string;
+  headless: boolean;
 }
 
 export class AgentServer {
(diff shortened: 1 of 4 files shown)


cowboyd:
Note, this depends on https://github.com/thefrontside/effection.js/pull/103

packages/agent/src/server.ts
```diff
@@ -35,6 +47,14 @@ export class AgentServer {
   *listen(): Operation { return; }
 
   *join(): Operation { yield; }
+
+  *launch(options: LaunchOptions): Operation<WebDriver> {
```
cowboyd, line 51: this field kinda ambiguous.... maybe it should be name `launchBrowser` or `launchConnectedBrowser`

packages/agent/src/server.ts
```diff
@@ -78,3 +98,17 @@ function listen(app: xp.Express, port?: number): Operation {
...
+  let builder = new Builder().forBrowser(browser);
+  switch (browser) {
+    case 'chrome':
+      let chromeOptions = new ChromeOptions();
+      chromeOptions = headless ? chromeOptions.headless() : chromeOptions;
+      return builder.setChromeOptions(chromeOptions).build();
+    case 'firefox':
+      let firefoxOptions = new FirefoxOptions();
+      firefoxOptions = headless ? firefoxOptions.headless() : firefoxOptions;
+      return builder.setFirefoxOptions(firefoxOptions).build();
```
jnicklas, line 112: I think it's not necessary to set these options conditionally. If a different browser is used the additional options are ignored.

cowboyd:
This won't be needed with the advent of #257

PR #254, opened 2020-04-24
Upgrade effection, improve linkability

cowboyd:
Motivation
----------

One of the challenges of working with effection inside the bigtest repo has been trying to use a linked version or a preview version and not have things go terribly awry. This is because each package has it listed as a runtime dependency. However, most packages don't actually care that there be a particular version of effection installed, merely that _somme_ version is available. [I _think_ that this is why you want to have peer dependencies.][1]

Throughout our tree, we want to make sure that everyone is handed the same copy of effection (in this case 0.6.2) and the only way to guarantee that is via a peer dependency.

Approach
----------

By using peer dependency everywhere there is a reference to effection, every library will get the same version of effection that is loaded by the root level require. 

This is a bit annoying because we have to include effection twice: once in `devDependencies` and once in `peerDependencies`, but the payoff is that if I link effection inside the `@bigtest/server` package, and import from `@bigtest/agent`, because `@bigtest/agent` is required from `@bigtest/server` it will defer the version of the library to its parent with is the version in the devDependencies of `@bigtest/server`

[1]: https://classic.yarnpkg.com/blog/2018/04/18/dependencies-done-right/

### DIFF
--- a/packages/atom/package.json
+++ b/packages/atom/package.json
@@ -19,13 +19,15 @@
   "devDependencies": {
     "@types/mocha": "^7.0.1",
     "@types/node": "^12.7.11",
+    "@types/ramda": "types/npm-ramda#dist",
     "expect": "^24.9.0",
     "mocha": "^6.2.2",
     "ts-node": "*"
   },
+  "peerDependencies": {
+    "effection": "^0.6.0"
+  },
   "dependencies": {
-    "@types/ramda": "types/npm-ramda#dist",
-    "effection": "^0.6.0",
     "ramda": "0.27.0"
   },
   "volta": {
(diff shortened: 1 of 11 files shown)


packages/atom/package.json
```diff
@@ -19,13 +19,15 @@
   "devDependencies": {
     "@types/mocha": "^7.0.1",
     "@types/node": "^12.7.11",
+    "@types/ramda": "types/npm-ramda#dist",
     "expect": "^24.9.0",
     "mocha": "^6.2.2",
     "ts-node": "*"
   },
+  "peerDependencies": {
+    "effection": "^0.6.0"
```
taras, line 28: Should this be ^0.6.2?

PR #257, opened 2020-04-28
Create @bigtest/webdriver

cowboyd:
Motivation
----------

BigTest needs to be able to start and stop browsers automatically in order to create the browser agents that will be part of a test run. So, if we want to run our test suite on Firefox _and_ Chrome within the single test run, we'll need to be able to start thosebrowsers, and then direct them at the agent server.

Approach
--------

All the major browsers support the [webdriver protocol][1], both to startup and to control a browser over a series of HTTP requests. Rather than try and implement our own control mechanisms, this uses webdriver to launch a managed browser and have it navigate to a url.

This arrangement is ideal because most of the actual direction of the test happens within the browser itself as part of the agent/harness combination found in the `@bigtest/agent` package. In fact, the only operation that we need currently beyond starting and stopping is the `WebDriver#navigateTo()` which loads a url inside of the browser under control.

In this change, we introduce the subset of webdriver functionality that we need in order to get a bigtest running. This can be found in the `WebDriver` interface. Then, we provide a `Local()` operation to create a `WebDriver` that uses a locally installed Chrome or a Firefox to actually run the session.

For example:

```typescript
import { Operation } from 'effection';
import { WebDriver, Local } from '@bigtest/webdriver';

export function* connect(url: string): Operation<WebDriver> {
  // get the instance of the driver
  let driver: WebDriver = yield Local('chromedriver', { headless: true });

  // use it to got to a web page
  yield driver.navigateTo('https://frontside.com');

  // pass the driver resource to the caller
  return driver;
}
```

Screenshots
-------------

Using the local chrome driver to navigate to the bigtest and frontside websites:

Using the local chrome driver to navigate to the bigtest and frontside websites

Questions
---------

**Q:** Does this use Selenium?

**A:** No. This implementation does not use [Selenium][5]. It uses the webdriver protocol *directly* over HTTP.  As a result, it is much faster, and more tightly integrated with the [`effection`][4] paradigm which makes it more robust to extremely rapid setup and teardown.

**Q:** What about remote browsers launched from a cloud service?

**A:** While the scope of this change only includes launching local browser processes and not connecting to remote webdriver servers such as [BrowserStack][2] or [Sauce Labs][3], the implementation should be relatively straightforward, since even communication with the local processes happens over HTTP using webdriver REST calls.

**Q:** What about browsers and devices that don't support webdriver?
**A:** Between all the various cloud services it is actually quite difficult to find a browser that cannot be launched via webdriver. For those that don't, our useage of the protocol is very slim. In fact, it is limited to three REST calls: `POST /session`, `DELETE /session/${id}`, and `POST /session/${id}/url`. In the event that a device does not support webdriver, emulating those three calls should be within the realm of possibility. Finally, remember that BigTest does not require that it be the one to launch a browser in order for that browser to be an agent, only that the browser connect to the agent server url. It will always support just opening a browser and entering in the agent connect url.

[1]: https://w3c.github.io/webdriver/
[2]: https://www.browserstack.com
[3]: https://saucelabs.com
[4]: https://github.com/thefrontside/effection
[5]: https://www.selenium.dev

### DIFF
--- a/packages/webdriver/src/fetch.ts
+++ b/packages/webdriver/src/fetch.ts
@@ -1,0 +1,31 @@
+import { Operation } from 'effection';
+import
+{ fetch as nativeFetch,
+  AbortController,
+  RequestInfo,
+  RequestInit,
+  Response
+} from './native-fetch';
+
+export { AbortController, RequestInfo, RequestInit, Response };
+
+/**
+ * Use `fetch` as an effection operation.
+ *
+ * This implements the `fetch` api exactly as it appears in the
+ * JavaScript spec. However, an abort signal is automatically connected
+ * to the underlying call so that whenever the `fetch` operation
+ * passes out of scope, it is automatically cancelled. That way, it is
+ * impossible to leave an http request dangling. E.g.
+ *
+ *   let response = yield fetch('https://bigtestjs.io');
+ */
+export function* fetch(resource: RequestInfo, init: RequestInit = {}): Operation<Response> {
+  let controller = new AbortController();
+  init.signal = controller.signal;
+  try {
+    return yield nativeFetch(resource, init);
+  } finally {
+    controller.abort();
+  }
+}
--- a/packages/webdriver/src/local.ts
+++ b/packages/webdriver/src/local.ts
@@ -1,0 +1,42 @@
+import 'chromedriver';
+import 'geckodriver';
+
+import { Operation, resource } from 'effection';
+import { ChildProcess } from '@effection/node';
+import { once } from '@effection/events';
+
+import { findAvailablePortNumber } from './find-available-port-number';
+import { untilURLAvailable } from './until-url-available';
+import { WebDriver, Options, connect } from './web-driver';
+
+type LocalDriverName = 'chromedriver' | 'geckodriver';
+
+/**
+ * Create a local `WebDriver` resource based on `driverName` (either 'geckodriver' or
+ * 'chromedriver').
+ *
+ * In order to function, the local executables for `Chrome` and
+ * `Firefox` must be separately installed onto the machine. So in CI
+ * for example, you would need to make sure that the browsers are
+ * present on your container.
+ *
+ * In order to keep things simple, and because the geckodriver only
+ * suopports a single Firefox session per driver process, we spawn a
+ * new process for each local driver.
+ */
+export function * Local(driverName: LocalDriverName, options: Options): Operation<WebDriver> {
+
+  let port: number = yield findAvailablePortNumber();
+  let driverURL = `http://localhost:${port}`;
+
+  let driver = yield resource(new WebDriver(driverURL), function*() {
+    let child = yield ChildProcess.spawn(driverName, [`--port=${port}`]);
+    yield once(child, 'exit');
+  });
+
+  yield untilURLAvailable(`${driverURL}/status`, 1000);
+
+  yield connect(driver, options);
+
+  return driver;
+}
--- a/packages/webdriver/src/native-fetch.ts
+++ b/packages/webdriver/src/native-fetch.ts
@@ -1,0 +1,6 @@
+export { RequestInfo, RequestInit, Response } from 'node-fetch'
+export { AbortController } from 'abort-controller';
+
+import fetch from 'node-fetch';
+
+export { fetch };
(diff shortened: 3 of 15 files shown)


packages/webdriver/src/fetch.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function* fetch(resource: RequestInfo, init: RequestInit = {}): Operation<Response> {
+  let controller = new AbortController();
+  init.signal = controller.signal;
+  try {
+    return yield nativeFetch(resource, init);
+  } finally {
+    controller.abort();
+  }
+}
```
jnicklas, line 31: Since we're already using this in other places (the server test suite comes to mind) would it make sense to extract this into effection as @effection/fetch or something?

packages/webdriver/src/native-fetch.ts
```diff
@@ -0,0 +1,6 @@
+export { RequestInfo, RequestInit, Response } from 'node-fetch'
+export { AbortController } from 'abort-controller';
+
+import fetch from 'node-fetch';
+
+export { fetch };
```
jnicklas, line 6: Why is this in a separate file?

packages/webdriver/src/local.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (details && details.value && details.value) {
+      throw new Error(details.value.message);
+    } else {
+      throw new Error(`RequestError: ${response.status} ${response.statusText}`)
+    }
+  }
+
+  let json = yield response.json();
+
+  session = json.value;
```
jnicklas, line 89: What's the point of assigning this variable?

packages/webdriver/src/local.ts
```diff
@@ -0,0 +1,110 @@
...
+      let response = yield fetch(`${driverURL}/session/${session.sessionId}/url`, {
+        method: 'post',
+        body: JSON.stringify({ url })
+      });
+
+      if (!response.ok) {
+        throw new Error(`GetError: GET ${url}\n${response.status} ${response.statusText}`);
+      }
+    }
+  };
```
jnicklas, line 56: How do you feel about making this a class instead?

packages/webdriver/src/until-url-available.ts
```diff
@@ -0,0 +1,23 @@
...
+    throw new Error(`timed out waiting ${maxWait}ms for ${url} to become available`)
+  });
+  while (true) {
+    try {
+      let response: Response = yield fetch(url);
+      if (response.ok) {
+        return;
+      }
+    } catch (error) {
+      /* This is ok. Error is raised whenever a network connect fails */
```
jnicklas, line 20: Can we be more specific about this? We probably don't want to fetch *any* error, just a network error. Would be better to rethrow if we get a different error.

jnicklas:
Awesome stuff! :D Left a few minor comments.

packages/webdriver/src/fetch.ts
cowboyd: 💯

packages/webdriver/src/native-fetch.ts
cowboyd: I was trying to make the `fetch` code itself be agnostic about where the native fetch function was coming from so that we could use `@effection/fetch` in both the browser and node since in the browser, `fetch` is a global variable. Not sure the best way to do this in a multi-platform way, but figured that isolating the symbol definitions was a start.

packages/webdriver/src/local.ts
cowboyd: We need the `session.sessionId` in order to make further request to the Webdriver session.

packages/webdriver/src/local.ts
cowboyd: That would probably be wise. Looking at this code, it seems like there's actually nothing "local" about it. It's independent of the fact that the Webdriver server was started by running a local executable.

packages/webdriver/src/fetch.ts
cowboyd: Let's do this in a follow-on PR

packages/webdriver/src/local.ts
cowboyd: done.

packages/webdriver/src/until-url-available.ts
cowboyd: Now we check to see if it is a `FetchError` which is raised whenever the network conditions prevent a fetch from being performed.

PR #263, opened 2020-04-30
Project config and CLI

jnicklas:
Extract project config and load from config file

This separates the configuration of the orchestrator from the configuration of the project. The definition of those options now lives in @bigtest/project, along with methods for loading the configuration files.

I also added a `createServer` method to @bigtest/server, which essentially just starts the orchestrator, but without having to create an atom manually. This can serve as the main entry-point for the server package and is used as such in the CLI.

Finally, we're actually using the retrieved configuration and wiring it up with the CLI, so we can start a server and even run our tests from the CLI.

Supersedes #228


jnicklas:
One issue I ran into when testing this:

It's once again proving very tricky to shut down processes in the correct way. When we start the app server, we start it with its own process group, so that we can kill the entire process group, because it is otherwise very difficult to ensure that all of its children are killed off when we shit it down. But this is proving problematic when we run the CLI itself as a child process, as we do in the test suite of the CLI. Killing the CLI process group then does NOT kill the app server.

This is because we are killing processes by forcibly TERM-ing them, but since the app server is in a separate process group it won't actually die.

I went with some kind of hybrid solution where we are sending en INT to the entire process group, waiting for the processes to die for a little while, and if they don't, send them all a TERM.  I think that this might be the solution we want to adopt for all subprocesses.

PR #265, opened 2020-05-04
Percolate failures upwards

jnicklas:
If a test has steps, assertions or even children that fail, the test itself should also be marked as failed. This way, we report the status up the tree and are able to report an aggregate status for the entire suite.

It is worth noting that in a user interface, we are most likely interested in mainly showing the status of the *assertions*, not the tests, so while at first glance this might generate a lot of "failure-noise", in practice it probably won't even be visible to the user in most cases, but it will make it much easier to report on the aggregate result of test runs.


jnicklas:
I really want to refactor the command processor, the code is very hard to follow at the moment, but I'm leaving this to a separate PR.

PR #267, opened 2020-05-06
introduce `parallel` operation.

cowboyd:
Instead of manually munging forks, what about using a parallel helper instead?


jnicklas:
This is neat, I like it!

PR #269, opened 2020-05-07
Automatically launch agents from the orchestrator/cli

cowboyd:
Motivation
-----------
We want to be able to weave the starting and stopping of browsers into test flow seamlessly, whether in CI, or during development.

This adds the capability to manage a set of browsers to the orchestrator, and then specify which managed browsers to launch when the orchestrator spins up.

Approach
----------
The tack we take is to create a set named of "drivers" associated with the project that specify how to launch a runtime that will contain an agent so that tests can be run. At this moment, "driver" is really a synonym for "webdriver local", but the vocabulary was left intentionally more abstract in order to accomodate other kinds of drivers such as remote webdrivers as well as iOS, Android, and Node drivers.

In the `.bigtest.json` config, this adds the "drivers" field and a "launch" field. Each entry in the "drivers" field specifies a driver configuration that can be used by the "launch" field (and also by the `--launch` command line option). The idea is that the "driver" is just an opaque blop of configuration that can be passed to the underlying driver when it is launched. For example, you might have two drivers for chrome, one that uses the default window size, and another that launches with a small window in order to test the small breakpoint in a responsive layout:

```json
"drivers": {
  "chrome": {
    "browserName": "chrome",
    "headless": true,
  },
  "chrome-small": {
    "browserName": "chrome",
    "headless": true,
    "width": 600
  },
  "launch": ["chrome"]
}
```

Now, when starting the server, the orchestrator will see the "chrome" option in the project config's "launch" field, and automatically launch that driver with the configuration in the driver field. While "chrome-small" is _available_ to be launched, it will not be launced by default in the absence of a `--launch` parameter. However, if you _do_ want to launch it, here's how you would:

```
$ bigtest server --launch chrome-small
```

All bigtest projects come pre-configured with the `chrome`, `chrome.headless`, `firefox`, `firefox.headless`, and `safari` drivers.

> Note: In order to facilitate having a single opaque configuration parameter, the `@bigtest/webdriver` package was refactored to take a single parameter.

### Screenshots

Starting the server with custom launches for chrome and Firefox. Both browsers connect to the orchestrator, and then a second command is issued to run the tests.

TODOs and Open Questions:
-------------------------

- [x] Not super happy with the additions to Project config. The problem is, that from the command line you want to be able override configuration like setting "headless" to `false`, so the format of the "launch" field has to have the same format as the "driver" field, but that's somewhat confusing since the purpose of the "launch" field isn't to provide configuration, just specify _which_ browsers to launch. Feels more like it should be an array like `"launch": ["chrome"']`. I only did it this way so that `createServer` could take a `ProjectOptions` object and not have to diverge. **simplified options make this a non-problem **
- [x] The argument parsing for the `--launch` flag feels really quite flimsy. It parses everything after the `@` sign as [`JSON5`][1] with an implicit bracket, which feels hackish, but more importantly it lacks the things you expect from a robust CLI like:
    - at a minumum validate of options, so `--launch chrome@blah:true` is an error as would be `--launch chrome@headless:5`
    - completition: `--launch X` should complete X from the list of available drivers. `--launch chrome@` should should completation  of the chrome drivers attributes, etc.. ** by making named drivers the only way to specify launch options, this is no longer a problem**
- [x] In order to validate CLI options, we need to load the config, which is an effection operation, but we don't actually do anything effection-wise until we're deep in yargs. I'm thinking eventually we'll need to find a way to interleave yargs and effectful code. ** Not addressed in this PR, but we will eventually parse the CLI arguments into a data structure and then proceed from there**
- [x] Don't want to make the config confusing so that it's constantly a source of frustration. Still, drivers will have complex configuration, especially for remote cloud-based drivers that will need to specify driver urls and auth tokens. Maybe making configuration sparse? ** default configurations have been included for popular browsers **
- [x] There is a potential race condition lurking in the browser launching resource. According to the webdriver spec, the navigation command will return once the `load` event has been dispatched on the document. This makes sense, but we really don't want to consider the launch complete until we receive the corresponding `connect` event from the agent as it calls home to activate itself. If we try to run the test before this connect event happens, or worse, if the agent code somehow fails after the `load` event happens, we'll think the agent should be connected when in fact it is not. ** CRUSHED with Atom.once **

[1]: https://json5.org


packages/cli/test/cli.test.ts
```diff
@@ -21,7 +20,7 @@ describe('@bigtest/cli', function() {
     let child: Process;
 
     beforeEach(async () => {
-      child = await World.spawn(run('server'));
+      child = await World.spawn(run('server', '--launch', 'chrome@headless:true'));
```
jnicklas, line 23: Awesome to be able to re-use this in the tests, love it!

packages/server/bin/start.ts
```diff
@@ -26,5 +26,19 @@ main(createServer({
...
+  drivers: {
+    "chrome": {
+      browserName: "chrome",
+      headless: true
+    },
+    "firefox": {
+      browserName: "firefox",
+      headless: true
+    }
+  },
```
jnicklas, line 39: What is the purpose of having this as configuration, do we actually expect people to change it?

packages/server/src/browser-manager.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    for (let launch of launches) {
+      let driver: WebDriver = yield Local(launch);
+      yield driver.navigateTo(options.connectURL);
+    }
+
+    ready.resolve();
+
+    yield;
+  })
```
jnicklas, line 35: This is a bit of a departure from the current design of the orchestrator, where it doesn't use resources, but rather the `delegate` pattern to report readiness. I think that our experience with resources shows that they are nicer to work with and easier to understand, so this is probably a good thing, still wanted to call it out so we can think about if this is the direction we want to evolve the orchestrator into as well.

packages/webdriver/src/local.ts
```diff
@@ -24,7 +22,9 @@ type LocalDriverName = 'chromedriver' | 'geckodriver';
  * suopports a single Firefox session per driver process, we spawn a
  * new process for each local driver.
  */
-export function * Local(driverName: LocalDriverName, options: Options): Operation<WebDriver> {
+export function * Local(options: Options): Operation<WebDriver> {
+
+  let driverName = options.browserName === 'chrome' ? 'chromedriver' : 'geckodriver';
```
jnicklas, line 27: This doesn't seem very future proof, what about Safari, Edge, etc?

packages/server/bin/start.ts
cowboyd: So this has given me great consternation as to the best way to represent this. The reason I put the driver config as an opaque piece of information would be so that we could extend it with other drivers in the future. The near future being cloud drivers like BrowserStack. Still, this feels gross, and it should probably be sparse.

The case where I can see people tweaking and changing is when we start adding configurations like adjusted window sizes, network latencies.... minimized vs maximized... all of which Webdriver supports.

Ugly and confusing config is something we want to avoid for sure, so I'd love to figure out ways to make this intuitive.

One thing I've been thinking is that might be easier in the long run if we make our config `.js` instead of `.json`, that way we can make this simpler, as well as making plugins able to provide their configuration options:

```js
const { Local } = require('@bigtest/project');
const { BrowserStack } = require('@bigtest/browserstack');
module.exports = {
  drivers: {
    chrome: Local.chrome(),
    safari: Local.safari(),
    cloudsafari = BrowserStack({
      "platformName": "iOS",
      "deviceName": "iPhone 7",
    })
}
```

Would love to hear your thoughts.

packages/server/src/browser-manager.ts
cowboyd: I agree 100%. I think symmetry of code is often more important than using the latest and greatest. That said, using resources and here felt _amazing_ here. I feel very strongly that this should be the new normal for our other services. As I see it, our options are:

1) go with this, and then refactor the other "raw forks" to become resources later.
2) refactor other "raw forks" into resources now, then rebase this onto that.
3) re-implement this using the "raw fork" model and then refactor all of them together in one go.

packages/webdriver/src/local.ts
cowboyd: I'm not sure, but I don't think so. For local Webdriver, there is only a finite list to be supported. 'chromedriver', 'safaridriver', 'gecko driver'. When we add support for those drivers, we can resolve those too.

For remote Webdriver configs, when we land them, the `browserName` is just passed through to the Webdriver session transparently and _it_ will decide which browser process to route to.

The main goal with this change was to compress the driver options into a single object instead of two, so that they could be specified in the project config.

packages/server/bin/start.ts
taras: Are [these](https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/webdriver.d.ts#L229) the options that you can configure with Webdriver?

cowboyd: As with everything WebDriver it's a bit more complicate :) If I understand correctly, these are the options supported on all WebDrives https://github.com/webdriverio/webdriverio/blob/master/packages/webdriver/webdriver.d.ts#L186-L227

Those others and optionally supported.

cowboyd:
@jnicklas This raises an interesting question with respect to effection resources. I tried to refactor the web drivers to start in parallel, but there wasn't an easy way that I could find to retain a list of resources that is generated by the parallel computation. It's almost like what's needed is an array resource that then retains each resource in its set.

I took the easiest path for now, making the actual resource allocation serial so that they would be retained by the main resource context, and then spawned in parallel the tasks to load the agent code.

It's something we should look at in the future.

jnicklas:
Wow, that's evil. It makes total sense that it would behave this way, but it bears thinking about how we could solve this.

PR #272, opened 2020-05-13
Add DSL for writing test suites

jnicklas:
This is a DSL which builds the datastructure needed for our test suites, but in a more convenient and less verbose way.

It is built to be typescript friendly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
       "packages/project",
       "packages/todomvc",
       "packages/suite",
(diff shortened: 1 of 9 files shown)


jnicklas:
This is only a proof-of-concept so far. It allows threading through a context type, which bigtest does not actually support yet. But it demonstrates how such a thing is possible, reasonably ergonomic and can actually be made totally typesafe in typescript.

PR #289, opened 2020-05-15
Do not let steps run indefinitely

cowboyd:
Motivation
-----------
Steps are async functions which means that the test runner hands over control to them completely and only receives control back once the async function is complete. However, this means that a step action that never returns would keep the entire test run from never returning.

Approach
----------
This constrains each step with a 2s timeout at which point, the step will be reported as a failure and a new `timeout` parameter will be set on the result message that indicates that this was not an ordinary failure, but rather a case of the step never returing.

Open Questions
--------------

- [x] The timeout is not configurable at the moment since that feature can  be easily added as a separate change since the developer ergonomics around that require some thought. #290 
- [x] Part of me is philsophically opposed to having a a nullable boolean as was added in the StepResult.boolean?. Something about it feels lazy since it relies on the default being false. I mean, there's only two values. However, I went with it anyhow since this is really a polymorphic data structure and you aren't interested in the `timeout` property unless status is `failure` (similar to `error?`)
- [x] We don't currently support running assertions continuoly and adding a timebox around those, which will make them less robust. #291
- [x] It would be nice to capture step and assertion metrics to see how long a step is taking to run even when it does succeed. #292


packages/agent/app/create-harness.ts
```diff
@@ -47,14 +49,31 @@ function *runTest(parentFrame: ParentFrame, test: TestImplementation, context: T
...
+        })
+      } else {
+        parentFrame.send({
+          type: 'step:result',
+          status: 'failed',
+          timeout: false,
+          error: serializeError(error),
+          path: stepPath
+        });
+      }
```
jnicklas, line 76: Didn't realize this was originally introduced here, I have the same comment as on #293: can we do this by racing the timeout against the success case instead? The try/catch seems a bit awkward to me.

PR #293, opened 2020-05-15
Make assertions convergent

### LINKED ISSUE #291: Make assertions convergent over time
The runner [only runs an assertion once][1], but it may be possible that effects of the actions in a step have yet to be fully realized in a form that will make the assertion pass. We've done this in the past with `@bigtest/convergence` to happy effect

picture of many assertions that eventually pass once an effect has been realized in application state

[1]: https://github.com/thefrontside/bigtest/blob/master/packages/agent/app/create-harness.ts#L87

cowboyd:
> resolves #291 
> depends on #289 

Motivation
-----------
It's possible that the effect of a step action takes some time to be realized before we can actually make an assertions that verify it.

This will waits for an assertion to become true over the course of 2s, and only then fails if it does not. Rather than throwing an error on a failure as had been happening before, this uses the `Maybe` pattern` since the end result that we're really interested in is the error itself, but it could be thrown hundreds of times over

Open Questions
--------------

- [ ] Should we convert the `timebox` function to use `Maybe` instead of throwing a TimeoutError.
- [ ] How do we configure assertion timeouts? The same principles and questions apply to this as to step timeouts... should they be the same?


packages/agent/app/timebox.ts
```diff
@@ -0,0 +1,12 @@
+import { Operation, spawn, timeout } from 'effection';
+
+export function* timebox<T>(operation: Operation<T>, timelimit: number): Operation<T> {
+  yield spawn(function*(): Operation<void> {
+    yield timeout(timelimit);
+    let error = new Error(`timelimit of ${timelimit}ms exceeded`);
+    error.name = 'TimeoutError';
+    throw error;
```
jnicklas, line 8: Not sure of the best practices here, should we subclass `Error` instead?

packages/agent/app/maybe.ts
```diff
@@ -0,0 +1,21 @@
...
+  return !isNothing(maybe);
+}
+
+export type Nothing = {};
+
+export const Nothing = {};
+
+export function isNothing<T>(maybe: Maybe<T>): maybe is Nothing {
+  return maybe === Nothing;
+}
```
jnicklas, line 21: Is there some library we can use for this, rather than rolling it on our own?

packages/agent/app/create-harness.ts
```diff
@@ -47,30 +51,48 @@ function *runTest(parentFrame: ParentFrame, test: TestImplementation, context: T
...
       console.error('[harness] step failed', step, error);
-      parentFrame.send({ type: 'step:result', status: 'failed', error: serializeError(error), path: stepPath });
+      if (error.name === 'TimeoutError') {
+        parentFrame.send({
+          type: 'step:result',
+          status: 'failed',
+          timeout: true,
+          path: stepPath
+        })
+      } else {
```
jnicklas, line 70: Can we do this without using exceptions for control flow? We could implement this as two operations racing against each other. Though we'd need an Effection equivalent to `Promise.race`.

jnicklas:
Personally, I'd kind of like to hold off on making assertions convergent until we have a better understanding of the interactor API. My gut feeling is still that assertions should be async functions. My thinking is that adding convergence to assertions will make the interactor API more inconsistent, and ultimately confusing to use. I know I haven't done a good job of articulating why I think this is the case, so I'll try to explain.

The main reason for this is the need for "anchoring" steps through assertions. Basically this is like adding debug asserts to code to ensure that you are in the right place. An easy to understand example would be a paginated list. Imagine you want to navigate to the *third* page of a paginated list. You'd have to click "next" twice:

``` javascript
// I'm on the first page
await Link("next").click()
// Am I on the second page?
await Link("next").click()
// Wait, where am I now?
```

The problem here is that you cannot know whether you have actually transitioned to the second page before you click the "Next" link the second time. So you might end up hitting "Next" twice on the first page, leaving you on the second page instead. So you'd want to anchor this test by adding an assertion in between:

``` javascript
await Link("next").click()
await Headline("Page 2").exists()
await Link("next").click()
```

There is no longer any ambiguity in this.

> **Sidenote:** A mistake which I think most frameworks in this space make is they fail to realize that the first case is inherently ambiguous and *cannot* definitively be resolved. They try to guess what the user meant and often end up getting it wrong. Writing a test like this is simply a mistake, and we should teach people to do the right thing, rather than making bad guesses.

But now we have a way of asserting on things in steps, which is asynchronous and in some way "contained" in the interactor API. It feels awkward to me to then have a *different* way of asserting in actual assertions. Additionally, making non-convergent access to properties on an interactor too easy, means that it's liable to be misunderstood or misused when it *does* need to be convergent on its own. Imagine if we have two different ways of verifying the `href` of a link:

``` javascript
// non-convergent
Link('Foobar').href === 'http://example.com';
// convergent
await Link('Foobar').hasProperty('href', 'http://example.com');
```

IMO it's harder to explain the difference between these two, rather than just always using the latter.

Also I'm a bit worried that we will need a deeper integration between interactors and the framework, but I'm not sure whether this is actually true.

jnicklas:
I commented here originally but moved the comment here instead, since it makes more sense organizationally, imo: https://github.com/thefrontside/bigtest/issues/291#issuecomment-630207212

cowboyd:
We're going to go with async assertions + safegaurds at the interactor level.

PR #297, opened 2020-05-19
Clean up proxy requests

jnicklas:
We noticed that if the Effection context shuts down very fast after loading the application, it can sometimes lead to uncaught exceptions. We tracked this down to the proxy server, and specifically requests which have started and have not yet completed by the time we shut down.

Interestingly this is probably the only place in the entire codebase where we're using regular evented code, rather than Effection based loops, and it is also the one place where it was broken. By using a proper effection based loop for the event handling the code becomes a lot wordier, but more correct, and the uncaught errors disappear.


jnicklas:
Ehh, I didn't see #296 😂

PR #299, opened 2020-05-20
Add GraphQL subscriptions for listening to test events

jnicklas:
We currently support live queries, where the result updates whenever we make a change to the state in the Atom, this is useful for building stateful UIs, but for building streaming UIs, it would be more convenient to be able to generate events during the test run and stream those, so our client doesn't have to reimplement the conversion back from stateful results to event stream.

Todos & Open Questions
-------------------------

- [ ] need tests for passing, failing, and disregarded steps/assertions
- [ ] clean up the commit history
- [ ] push up latest stashed changes which include non-local events.
- [ ] remove global event bus.
- [ ] some unit level test?
- [ ] Q: can this be broken up into smaller PRs?


jnicklas:
Did a rebase of this as a starting point for cleaning this up, but there were a lot of conflicts, so it was a bit of a mess.

jnicklas:
#341 replaces this.

PR #300, opened 2020-05-21
Turn CLI inside out

cowboyd:
Motivation
-----------

Adding commands and sharing code is difficult with the current CLI because it's driven through callbacks via yargs. That means we have to explicitly enter an effection context any time that we want to use an operation. What we'd like is to always be in effection from the get-go.

Approach
----------
This makes the yargs parsing, which is synchronous, a simple function that returns a data structure which can then be used to run the CLI. 

One thing is that there is a chicken and the egg problem with loading the config because we don't know what the config file is until we parse the options and we don't know the all the options until we parse the config. This temporarily punts on this by removing the `config-file` option. What we get in return is better validation of the options because the choices and defaults are all drawn from the config.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -8,128 +8,128 @@ import { setLogLevel, Levels } from '@bigtest/logging';
 import { loadConfig } from './config';
 import * as query from './query';
 
-export function CLI(argv: string[]): Operation {
-  return ({ fork, resume }) => {
-    yargs({})
-      .scriptName('bigtest')
-      .command('server', 'start a bigtest server', (yargs) => {
-        yargs
-          .option('launch', {
-            describe: 'launch specified driver at server startup',
-            type: 'array',
-            default: []
-          })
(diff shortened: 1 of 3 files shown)


jnicklas:
😍

packages/cli/src/cli.ts
```diff
@@ -8,128 +8,128 @@ import { setLogLevel, Levels } from '@bigtest/logging';
...
-    resume({})
-  };
+export function * CLI(argv: string[]): Operation {
+  let config: ProjectOptions = yield loadConfig();
+
+  let options = parseOptions(config, argv);
+
+  setLogLevel(options.logLevel);
+
+  if (options.command === 'server') {
```
jnicklas, line 18: This is definitely something we could do in a follow-up PR as well, but it'd be pretty nice to extract the bodies of the branches of this if-statement into functions.

cowboyd: 100% agree

PR #301, opened 2020-05-22
New Interactors

jnicklas:
This is my stab at an implementation of interactors. It is heavily based on @pittst3r's work in #109  and #220, with some notable implementations. I started off basing this off of #220, but ended up restarting from scratch, mainly because the implementation turned out to be sufficiently different, that I couldn't reuse too much, but #220 certainly served as inspiration for this.

My main goals with this implementation were:

- Excellent type safety when used with TypeScript. Interactors should be fully type-safe in both definition and usage.
- Composition of interactors
- Laziness: the element that an interactor works on should only be resolved at the very last instant. When performing an action, the element *including all of its ancestors* should be found in the same event loop tick as the action is performed. This prevents race conditions where the page structure changes between the moment of finding and performing the action.
- Synchronous actions: to make misuse of the actions API virtually impossible, we make actions synchronous so that they must be performed within one tick of the event loop.
- Convergent assertions: assertions like `exists` and `absent` (names bikesheddable) are convergent in the sense that they wait for the condition to be true.
- Introspectible definitions: when an interactor cannot be found we want to be able to provide really great errors. The best way of achieving this is by moving as much of the definition of the interactor out of executable code, and into introspecitble data instead.


packages/interactor/src/index.ts
```diff
@@ -0,0 +1,121 @@
...
+    } else if(matchingElements.length === 0) {
+      throw new Error(`${this.description} does not exist`);
+    } else {
+      throw new Error(`${this.description} is ambiguous`);
+    }
+  }
+
+  async resolve(): Promise<HTMLElement> {
+    return converge(defaultOptions.timeout, this.unsafeSyncResolve.bind(this));
+  }
```
jnicklas, line 78: This method is actually a bit dangerous as well, since a user might be tempted to do something like `(await Link("Foo").resolve()).href === '/foo' which is prone to race conditions. The correct thing to do would be something more akin to: `await Link("Foo").hasProperty('href', '/foo')` or something like that.

packages/interactor/src/index.ts
```diff
@@ -0,0 +1,121 @@
...
+export class Interactor {
+  protected parent?: Interactor;
+
+  constructor(private specification: InteractorSpecification, private locator: LocatorSpecification) {
+  }
+
+  find<T extends Interactor>(interactor: T): T {
+    let child = Object.create(interactor);
+    child.parent = this;
+    return child;
```
jnicklas, line 35: Not sure if this is acceptable or if it's too clever? Manually messing with prototypal inheritance doesn't seem to common in JS.

jnicklas:
One weakness of this as it stands is that the type of the element is always HTMLElement, but we'd actually like this type to be generic, because something like an `HTMLLinkElement` provides additional useful properties which we could use. The problem with this is the signature of `interactor`, currently it has the signature `interactor<A extends ActionSpecification>(...)`, we really, *really* want the type of `A` to be inferred, but unfortunately, we cannot then add another generic type to it which is expected to be explicitly given. It seems that the only way around this is to make `interactor` a curried function, like this:

```
// it looks a bit dumb
const Link = interactor<HTMLLinkElement>()({ name: "link", ... });
// or maybe this to make it look a bit less dumb:
const Link = interactor<HTMLLinkElement>('link')({ ... });
```

This  currying really isn't necessary for any other reason that being able to specify the type of an element while leaving the action specification up to inference. I'm not sure how I feel about twisting an API in this way just to make TypeScript happy.

Is there any other solution?

jnicklas:
One problem that I realized with this is that the implementation of composability is too simple. Imagine a case like this:

(in peudo code)

```
let one = A.find(B).find(C)
let two = X.find(Y)

let three = one.find(two)
```

We would expect `three` to now be: `A -> B -> C -> X -> Y`, but in fact, what we get is `B -> C -> Y`, which is obviously completely broken.

The problem is that we try to stitch the composed interactor by changing the `parent`, but we would actually need to walk up the ancestor chain, and stitch the top most ancestor's parent. I've thought about this back and forth a bit, but I haven't yet come up with an elegant solution.

dagda1:
> > Is there any other solution?
> 
> Not that I can see... I usually defer to @pittst3r or @dagda1 for these things :)

Really hard for me to know what is required here as I'm not up to speed with the code but I think my first step would be to remove all explicit casts and 

```typescript
        fullSpecification as unknown as InteractorSpecification<HTMLElement, LocatorSpecification<HTMLElement>>,
        locator as unknown as Locator<HTMLElement>
```

This is hammering the type on no matter what and if the compiler cannot figure it out then I would be inclined to listen to the compiler here.  

`as unknown as whatever` generally never ends well.


I'm agreeing with the compiler here also:

```javascript
return interactor as Interactor & ActionImplementation<E, A>;
```

I think there is a lot going on here and `Partial` and also this `unknown` will lead to problems

```javascript
export type ActionSpecification<E extends HTMLElement> = Record<string, (element: E) => unknown>
```

I think I would have a generic return type here:


```javascript
export type ActionSpecification<E extends HTMLElement, R> = Record<string, (element: E) => R>
```
That at least should help inference.

You guys know better than me but I'd personally try and simplify this right down.  There is too much going on for my money and the type is not getting narrowed so tsc complains

jnicklas:
@dagda1 I think I have addressed some of your concerns. All instances of `as unknown as T` have been removed. Unfortunately I wasn't able to remove all explicit casts, and this is mostly because in some places I have to downcast from e.g. a `Locator<E>` to `Locator<Element>`, since the `Interactor` class is not generic over its element.

It would actually be nice to make it generic over its element, but I was unable to get TypeScript to accept this for some reason. Both the type of `ancestors` and `find` seemed to be impossible to specify correctly when `Interactor` had generic types.

> this unknown might lead to problems
> 
> ``` typescript
> export type ActionSpecification<E extends HTMLElement> = Record<string, (element: E) => > unknown>
> ```

It's worth noting that `ActionSpecification` is only used as a *constraint* on a generic type (i.e. `<S extends ActionSpecification>`), not as an actual type, and we use this in a mapped type, so the actual type of the returned value will be known when the generic type is instantiated. The user of the library will never see `unknown` in a signature on their interactors. Using a generic type on `ActionSpecification` will not work, since I think that would constrain all of the actions to have the *same* return type, which is not what we want, they should be able to have distinct return types. Maybe there is a better type to use here than `unknown` though?

I did make a (in my opinion) [simpler implementation of the types](https://github.com/thefrontside/bigtest/blob/f4a19b0c3db6fb31a93c717e416436192074b444/packages/interactor/src/specification.ts), this is a little less confusing, I think, since there is no longer a need to have separate types for the action and locator specifications. It is a fact though that there is some seriously advanced type-hacking going on here. In my mind this is a bit unavoidable if we want to have an interface that is both succinct to define, succinct to use, and works equally well in typescript and vanilla JS.

packages/interactor/src/converge.ts
```diff
@@ -0,0 +1,23 @@
+import { main, timeout as effectionTimeout } from 'effection';
+
+// TODO: this API is available on browsers as `window.performance`, we need to figure out
+// a way to package this so it'll work on both browsers and node.
+import { performance } from 'perf_hooks';
+
+export async function converge<T>(timeout: number, fn: () => T): Promise<T> {
+  return await main(function*() {
```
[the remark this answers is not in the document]
jnicklas: True!

packages/interactor/src/converge.ts
```diff
@@ -0,0 +1,23 @@
+import { main, timeout as effectionTimeout } from 'effection';
+
+// TODO: this API is available on browsers as `window.performance`, we need to figure out
+// a way to package this so it'll work on both browsers and node.
+import { performance } from 'perf_hooks';
```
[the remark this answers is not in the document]
jnicklas: Nice!

dagda1:
> simpler implementation of the types

Had a quick look and it really is looking so much simpler.  Great job!  I was a bit lost last time but not now.  Really clean and when I. hover over the types in vscode, I don't get big unions.  The inference will work a lot better now.

I get your point about the generic constraint but I think it might help with type narrowing and inference if it can narrow on something better than unknown.  Even though it is a constraint the compiler will use that outside of the type argument in the function body.  I might need to prove that is the case so don't shoot me if I am wrong.

If I was writing this code, I would remove all the return types from the functions in `Interactor`.

i.e.

`private unsafeSyncResolve(): Element {`

becomes

`private unsafeSyncResolve() {`

You want to be specifying the bare minimum of types and let inference kick in wherever it can.  Now that the types are so much better, everything should work the way you expect.

In the same way as 

`let a = 'b'`

Is better than

`let a: string = 'b'`

On a side note, I know @cowboyd is big on `let` for stylistic reasons which I totally get but it can on occasion in typescript make types less specific so something to bear in mind.

One question about `Interactor#find`

Do you need to use `Object.create` which is not type-safe.  Could you do this instead which is type-safe?  

```javascript
let child = {...interactor};
```

Or is there more going on?

jnicklas:
@dagda1 

> Do you need to use Object.create which is not type-safe.

We could do this if we don't used a class for Interactor. I'm not entirely opposed to this, but it seems a bit wasteful to me, in terms of resource usage. Using the prototype chain seems much more efficient to me, rather than copying in a whole bunch of properties. Maybe that's old-school of me, but I don't like allocating a bunch of stuff when it can be avoided.

packages/interactor/src/index.ts
jnicklas: Good suggestion!

packages/interactor/test/interactor.test.ts
```diff
@@ -0,0 +1,149 @@
...
+  defaultLocator: (element) => element.textContent || ""
+});
+
+const Div = interactor({
+  name: 'div',
+  selector: 'div',
+  defaultLocator: (element) => element.id || ""
+});
+
+function dom(html: string) {
```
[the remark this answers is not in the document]
jnicklas: Neat, but I will leave it as is, I think.

dagda1:
@jnicklas how about something [like this](https://github.com/thefrontside/bigtest/commit/5776488ae363c0a9932c678767b75179818dd2f8) for making `Interactor` generic

dagda1:
You might want to put the explicit return type back into `Interactor#find`

`find<T extends Interactor<E>>(interactor: T): T {`

I do think `Object.create` is a premature optimization here and you are sacrificing type safety for those reasons.  Also other problems with Object.create and making the source object the prototype so I avoid this approach

packages/interactor/test/define-interactor.test.ts
```diff
@@ -0,0 +1,220 @@
+import { describe, it } from 'mocha';
+import * as expect from 'expect'
+import { JSDOM } from 'jsdom';
+
+import { defineInteractor, setDefaultOptions } from '../src/index';
+
+process.on('unhandledRejection', () => {
```
cowboyd, line 7: Is this necessary? I almost always find that when I'm getting `unhandledRejection` warnings, its because I probably should be explicitly catching something (even if I'm going to be ignoring the exception). In other words, explicit error swallowing is almost always preferable to implicit error swallowing.

jnicklas: It actually seems to work perfectly fine without it, I can't even remember why I added this 🤔

jnicklas:
> should we use createInteractor instead of defineInteractor? I know that we're declaring stuff that will be evaluated lazily, but createX seems to be the prevailing convention.

`createInteractor` sounds good to me!

> How do we compose interactors other than with find? Is there a way to "extend" an interactor, so if I wanted to interact with an element that was a div but also had a tabindex, would I say something like Focusable(Div)?

There is currently no way of doing inheritance of interactors, so you wouldn't be able to create an interactor which extends another interactor. I think this shouldn't be too difficult to add if we want to, actually, since it would just be extending and overriding the interactor specification.

packages/interactor/src/converge.ts
```diff
@@ -0,0 +1,22 @@
+const win: { performance?: unknown } = (typeof(window) === 'object') ? window : {};
+const performance = (typeof(win.performance) === 'object') ? win.performance : require('perf_hooks').performance;
```
pittst3r, line 2: Is it possible to use a dynamic `import()` for this instead? Using `require()` here isn't a huge deal because it's supported by build tools like Webpack, but it is a node thing and I like to try to avoid using it in code that could be run in the browser. I'm not sure about the performance implications.

packages/interactor/src/errors.ts
```diff
@@ -0,0 +1,23 @@
+export class NoSuchElementError extends Error {
+  get name() { return "NoSuchElementError" }
+}
+
+export class AmbigousElementError extends Error {
```
pittst3r, line 5:
```suggestion
export class AmbiguousElementError extends Error {
```

packages/interactor/src/interactor.ts
```diff
@@ -0,0 +1,86 @@
...
+        value: [...this.ancestors, this, ...interactor.ancestors]
+      }
+    });
+  }
+
+  get description(): string {
+    return this.ancestors.slice().reverse().reduce((desc, interactor) => {
+      return `${desc} within ${interactor.name} ${interactor.locator.description}`
+    }, `${this.name} ${this.locator.description}`);
+  }
```
pittst3r, line 31: Something to keep in mind for the future, but I can see a trace with more than two ancestors getting hard to read on one line. We may want to display this like how tsc shows type error traces.

packages/interactor/src/interactor.ts
```diff
@@ -0,0 +1,86 @@
+import { converge } from './converge';
+import { InteractorSpecification } from './specification';
+import { Locator } from './locator';
+import { defaultOptions } from './options';
+import { NoSuchElementError, AmbigousElementError, NotAbsentError } from './errors';
+import { interaction, Interaction } from './interaction';
+
+export class Interactor<E extends Element> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private ancestors: Interactor<any>[] = [];
```
pittst3r, line 10: Just a nitpick, but this would be a more conventional way to type this:

```suggestion
  private ancestors: Array<Interactor<any>> = [];
```

packages/interactor/src/interactor.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  private unsafeSyncResolve(): E {
+    let root = defaultOptions.document?.documentElement;
+
+    if(!root) {
+      throw new Error('must specify document');
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let ancestorChain: Interactor<any>[] = [...this.ancestors, this];
```
pittst3r, line 41: You may be able to remove the type annotation here too.

```suggestion
    let ancestorChain: Array<Interactor<any>> = [...this.ancestors, this];
```

packages/interactor/src/locator.ts
```diff
@@ -0,0 +1,21 @@
...
+export type LocatorSpecification<E extends Element> = Record<string, LocatorFn<E>>;
+
+export class Locator<E extends Element> {
+  constructor(public locatorFn: LocatorFn<E>, public value: string, public name?: string) {}
+
+  get description(): string {
+    if(this.name) {
+      let name = this.name.toString().replace(/^by/, '');
+      name = name.charAt(0).toLowerCase() + name.slice(1);
+      return `with ${name} ${JSON.stringify(this.value)}`;
```
pittst3r, line 12: Why do we need to do `JSON.stringify()` if `this.value` is of type `string`?

packages/interactor/src/interactor.ts
jnicklas: Totally agree! We should in general put more thought into the design of the errors than I've done here. If filters become a thing this will become even more unreadable.

packages/interactor/src/converge.ts
jnicklas: The idea is that we shouldn't hit the `require` in the browser at all, since we want to use `window.performance` in the browser and only use the `perf_hooks` package in a node environment. So we would actually want this to *not* be packages. But this may not at all be the right way of achieving this.

packages/interactor/src/interactor.ts
jnicklas: I'm not really sure why, but TypeScript complains when I remove this type annotation.

```
src/interactor.ts:44:75 - error TS2339: Property 'specification' does not exist on type 'this | Interactor<any>'.

44       let elements = Array.from(parentElement.querySelectorAll(interactor.specification.selector));
                                                                             ~~~~~~~~~~~~~
src/interactor.ts:45:70 - error TS2339: Property 'locator' does not exist on type 'this | Interactor<any>'.

45       let matchingElements = elements.filter((element) => interactor.locator.matches(element));
```

My guess is that this is also some obscure variance/subtyping issue.

packages/interactor/src/locator.ts
jnicklas: Mostly I prefer doing `JSON.stringify(foo)` over simply `"${foo}"` because it properly handles stuff like double quotes contained in the string. It's not a huge deal, but IMO it's just a little bit nicer.

jnicklas:
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #303, opened 2020-05-23
New interactors generic elements

jnicklas:
This is an implementation of making the element of an interactor generic, so that we can use a subclass of `HTMLElement`, if we know that we're working exclusively with a certain type of element. This means we can use all of the nice helper methods and properties provided in the DOM API, rather than being restricted to just what is available on `HTMLElement`.

Unfortunately it does make the types a lot more complicated, and it also necessitates a change of the `interactor` function so that it becomes a curried function, due to inference issues. Since we want the type of the generic type parameters `A` and `L` to be inferred, but `E` to be explicitly given, we need to do this currying trick, since TypeScript currently [does not support](https://github.com/microsoft/TypeScript/issues/26242) partial inference of types.

It's a bit unfortunate to have to do this workaround, especially since it is completely unnecessary when using vanilla JS, but if we want to specify the type of the element, I think this is what we have to do. I've been unable to find any other workaround which is reasonably ergonomic.

### DIFF
--- a/packages/interactor/src/action.ts
+++ b/packages/interactor/src/action.ts
@@ -1,5 +1,5 @@
-export type ActionSpecification = Record<string, (element: HTMLElement) => unknown>
+export type ActionSpecification<E extends HTMLElement> = Record<string, (element: E) => unknown>
 
-export type ActionImplementation<T extends ActionSpecification> = {
-  [P in keyof T]: T[P] extends ((element: HTMLElement, ...args: infer TArgs) => infer TReturn) ? ((...args: TArgs) => Promise<TReturn>) : never;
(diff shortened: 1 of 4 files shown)


jnicklas:
I made this PR into the `new-interactors` branch, since I'm not sure myself if this is actually a good idea.

jnicklas:
Worth noting that the `Interactor` class itself is *not* generic. If we did make it generic then we could have `resolve` return a promise of a properly typed element, which is nice, I guess, but it also makes it much more cumbersome to store and pass around locators, and since `resolve` is more a debugging tool, rather than part of the API, we probably shouldn't optimize for it.

PR #309, opened 2020-06-01
Make assertions asynchronous

jnicklas:
This is a less opinionated take on assertions. Instead of forcing assertions to always be convergent, we're leaving the convergence up to the interactors. This both minimizes the contraints placed on the test framework, making it more widely useful, as well as making interactors more consistent, because we don't need a separate way of specifying actions and assertions for steps vs assertions.

Todos, Open Questions
------

Based on conversations yesterday, we want to make sure that signposts clearly direct folks toward the AAA pattern since our entire testing strategy is geared around the idea that assertions are _ready only_

- [ ] When customizing interactors, clearly mark via the API that you're implementing either a step or an assertion.
- [ ] Have the Runner implement a global activity like "step" or "assertion"
- [ ] throw errors if interactors try to run actions during an assertion tick.


jnicklas:
While disallowing asynchronous assertions makes side effects somewhat less natural, it definitely doesn't make them impossible. DOM manipulations are synchronous, for example. Also calling an async function and simply not awaiting it is totally possible in a synchronous context. I don't think there is any way we can completely prevent side effects in assertions.

One thing we could do is have some hook in the interactor library which disables (or warns) calling an action from an assertion. I think this is probably reasonably easy to implement. It's obviously only going to prevent one class of side effects, but given how we want to encourage writing our tests, it could maybe be the most common type.

> What is the case where the right thing is to have an async function other than to run an assertion over time? 

That is actually a very good point, and I don't really have an answer. The only counter-argument I can come up with is based more on wider applicability. If we think that BigTest might be used for other types of tests in the future, then those might have synchronous assertions, and it would be a bit odd not to be able to delineate synchronous from asynchronous assertions from each other. It seems a bit magical that assertions are always convergent.

Convergent assertions by their very nature always have to wait for the timeout to elapse before reporting a failure, this is also something to consider.

cowboyd:
> One thing we could do is have some hook in the interactor library which disables (or warns) calling an action from an assertion. I think this is probably reasonably easy to implement. It's obviously only going to prevent one class of side effects, but given how we want to encourage writing our tests, it could maybe be the most common type.

How would you see this working? Would we insert a hook in the test runner?

PR #311, opened 2020-06-02
DSL shorthand for interactors

jnicklas:
Depends on #309 

If we use interactors with our DSL we end up repeating a lot of information:

``` typescript
.step("I click the 'next' link", async() {
  await Link('next').click();
});
```

It would be amazing if we could simply infer this information, and write this more succinctly like this:

``` typescript
.step(Link('next').click())
```

It should be pretty obvious why this is a huge win. Interactors already implement the required functionality for this to work in #301. All interactions are "lazy" promises so they will not actually do anything until they are awaited, this allows us to use the same API in this context, even though it is run in a "static" context. Additionally, interactions have a `description` property which describes the action.


cowboyd:
I agree that this is a feature we want, but the implementation is clever enough to make me wonder if it oughtn't be made more explicit, (and as a result more general) to allow for anything to present itself as a step.

What about using a symbol inter-op a-la `Symbol.iterator`:

```ts
export const ToStep: unique symbol = Symbol.for('Symbol.ToStep');
export interface StepLike {
  [ToStep](): StepImplementation;
}
```

This has the advantage not only of following JS convention, but also being more future-proof to the extension of the `Step` interface, so for example, if we wanted to add constraints as first class citizens of a step, that would be really straightforward to capture.

If we want to provide a path for other libraries than interactor to present themselves as steps, it would be much clearer what to do simply by looking at the interface declaration.

jnicklas:
Hmm, interesting. This would allow us to side step the issue of having to have some kind of special-casing for promise here, which is nice, but I personally find this kind of symbol-based solution to be *more* complex. We also end up with a bit of a chicken-and-egg situation in which library should define the `ToStep` symbol. We don't really want an inter-dependency between `@bigtest/interactor` and `@bigtest/suite` if we can avoid it.

cowboyd:
@jnicklas Is the complexity the fact that it uses a symbol, or that it uses the delegate pattern at all? (like the `to_thing()` pattern in Ruby)

As far as dependencies go, we could put it in a separate package. Also, we could use `Symbol.for()` to implement the interface. This type of one-off, super-loose coupling was (I think) designed to enable precisely this kind of interface. The trade off is that the symbol isn't statically linked so it's a little less safe if you mis-spell it, but that should be very quickly detected

jnicklas:
Funnily enough, I've actually become a bit concerned with this approach ;) The reason being that it needs to violate the promise spec in order to work properly, as pointed out by @pittst3r. The idea of a lazy promise is pretty good, but the problem is that if the promise is created in a static context, it is a violation of the spec to evaluate it multiple times. A promise value must not change once it has been assigned. In fact out current implementation doesn't do this, so this PR as implemented doesn't actually work. If a step is evaluated multiple times, only the first time will actually run.

I think your original idea of having some kind of `toStep` method is better (except it could use a symbol). An interaction is then still a lazy promise, but we don't use it as such when used in the static context of the DSL.

cowboyd:
> I think your original idea of having some kind of `toStep` method is better (except it could use a symbol). An interaction is then still a lazy promise, but we don't use it as such when used in the static context of the DSL.

Lol. Well, I can definitely see the advantages of a `Symbol.for('@bigtest.toStep')` and a `Symbol.for('@bigtest.toAssertion') approach 😄 

But I do share your original concern plus perhaps one more. The first is, do we have interactor depend on `@bigtest/suite`? This is actually not strictly necessary since you can write a class as:

```ts
class X {
  [Symbol.for('@bigtest.toAssertion')]() {}
  [Symbol.for('@bigtest.toStep')]() {}
}
```

and it will work when adding to the DSL.... kinda. The wrinkle is typescript. How do we slide the created step into the type chain of the `TestBuilder` in an abstract way?

We could constrain delegated steps to assume no context and always return void, and that would probably work in the case of interactors, but I could also see steps that inject things into the context as a single unit like our "createSimulator" step we use.

Another option would be to pass the TestBuilder object to the `Symbol.for('@bigtest.toStep')` method:

```ts
[Symbol.for('@bigtest.toStep')](builder: TestBuilder<C>): TestBuilder<R> {
  return builder.step("foo", () => doThing());  
}
```

but I don't like that because it means that we are now coupled to the specific test builder DSL, and I feel pretty strongly that you should be able to add anything that is a step to any test tree, no matter what DSL you're using. It also requires that you'll have to parameterize your class with the step inputs and outputs.

With that in mind, and given that we have a `Step` interface already, what if we just parameterized the core `Step` interface directly to include the TestBuilder API:

```ts
interface Step<C extends Context = {}, R extends Context | void = void> {
  description: string;
  action: (context: C) => Promise<R>;
}
```

The step builder class now cues its type chain from combining core steps, and all we need to do is implement an async `action` function on our interaction class and voilà, it's a `Step<{},void>` via the magic of structural typing without having to couple to anything.

Then, we can add an overload to our DSL to just accept any `Step` object directly.

We can still add a well-known symbol later, but then this also solves the problem by just making it a simple delegation, because then we can say:

```ts
import { toStep } from '@bigtest/symbols';

class X {
  [toStep] = () => ({
    description: "Do X",
    action:  () => Promise.resolve({hello: 'world'})
  })
}
```

jnicklas:
Superseded by #504 504

PR #312, opened 2020-06-03
Add hook to disable actions

jnicklas:
This adds a hook on the interactor library so we can disable actions, it remains to be solved how we hook this up into the runner.

### DIFF
--- a/packages/interactor/src/define-interactor.ts
+++ b/packages/interactor/src/define-interactor.ts
@@ -4,6 +4,7 @@ import { Interactor } from './interactor';
 import { interaction } from './interaction';
 import { converge } from './converge';
 import { defaultOptions } from './options';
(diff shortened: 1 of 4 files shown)


cowboyd:
Instead of making this a setting on interactor, what if interactors read state from the runner and adjusted behavior accordingly? That allows for more than one way to enforce which activities are acceptable during which phase.

jnicklas:
Yeah, I'm thinking that this probably isn't the ideal way of doing this. I've converted this to draft for now and will try and revise it.

jnicklas:
See also #315

jnicklas:
Somehow this got accidentally merged 🤔 I removed the commit from the `new-interactors` branch, so this is no longer a part of #301

Issue #286, opened 2020-05-14
Proposal: agent/harness responsibility and communication tweaks for wider applicability


jnicklas:
While lying in bed last night, I was thinking about interactors and how they would work, and my mind got stuck on a question: what happens if you navigate away from the page in the middle of an interaction? I realized that this question is actually broader than just interactors: what happens if you navigate away in the middle of a test?

So far, we have assumed that bigtest would be used to test SPAs, but what if that didn't have to be true? Can we actually use bigtest to test classical request/response applications? After thinking about this for a while, I think with some minor architecture tweaks we actually can! And just as an added bonus, I think this design also solves the most serious architecture problem that we've punted on so far.

### Status quo

Let's review on how the agent and harness interact at the moment and their respective responsibilities:

The agent will receive a message from the orchestrator telling it to run some tests, the orchestrator attaches a listing of all tests it wants to run to this message. The agent takes this data structure and splits it into *lanes*, since our tests are structured as trees, there is one lane for each leaf node in the tree. The agent then tells the harness that it want to run a particular lane, it does this by giving the harness the path of the leaf node that the lane ends in. The harness then loads the application, loads the test manifest, finds the lane in the manifest and runs the lane and reports the results of doing so to the agent, which in turn sends them to the orchestrator.

### Problems with the current approach

1) We have previously realized that the Harness cannot maintain state across tests, since as soon as it reloads the page, all state will be lost. We solved this by moving the state of the test run to the agent. However, there is still a huge amount of state on the Harness today: the current state of each test lane. Also, a vexing problem here is that we cannot in any way persist this state, since the state is code, not data (we ain't writing this in LISP 😉), so part of our state is where we are right now in a procedure, and this is something we cannot serialize or deserialize in any way. This works fine for SPAs, but it won't work if the application does any page reloads in between. Imagine a test like this:

``` javascript
step("go to product page", async () => {
  await Link("Products").click(); // navigates to product page
  await Link("Hoverboard").click();
});
```

If we navigate away in the middle of this step, how could we possibly "resume" by jumping back into the middle of this step?

[...]

PR #314, opened 2020-06-04
Prepare agent for merge of agent and proxy server

jnicklas:
Depends on #313 

This prepares the agent for merging with the proxy server. It removes a lot of the AgentServer machinery. One unfortunate side-effect of #286 is that we will not be able to have a reloadable development workflow while working on the agent. I don't really see a way of making this work, unfortunately. So much of the code in AgentServer, which is there to make this workflow necessary, is no longer needed.


packages/agent/src/connection-server.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Mailbox, readyResource } from '@bigtest/effection';
+import { express, Express, Socket } from '@bigtest/effection-express';
+
+interface AgentConnectionServerOptions {
+  port: number;
+  inbox: Mailbox;
+  delegate: Mailbox;
+}
+
+export class AgentConnectionServer {
```
cowboyd, line 11: Can we please call this object `AgentHandler` it's just too good a pun to pass up, and it's also very descriptive, not often that this happens!

PR #316, opened 2020-06-08
prepare the way for node drivers

cowboyd:
Motivation
----------

In order to use bigtest on a project that has both a web application _and_ a server side application, we need the ability to use bigtest to to test _both_. This means having a node agent that can run node tests. However, a pre-requisite for having a node agent, is allowing bigtest to manage node processes in the same way that it manages browser processes. Currently, the way this works is to abstract away the process management of agent into a _driver_ which is responsiblefor starting the agent process, connecting the agent process to the orchestrator, and ultimately shutting down the agent when it is no longer needed. But the only drivers that we have now are web drivers that manage browser processes.

Approach
--------

This clears the way to make drivers more of a general purpose concept in BigTest rather than being implicitly local webdriver instances. Now, the server does not actually reference `@bigtest/webdriver` at all directly, instead using only the abstract driver interfaces.

To accomplish this, we adopt a "plugin-like" architecture, where part of the configuration of the driver is the node module where it can be found. Whereas before the drivers section the project config looked like:

```json
{
  "chrome": {
    browserName: "chrome",
    headless: true
  }
}
```

We now specify the module implementing the driver, and a set of "options" that will be passed through transparently to the driver. So the previous example will let us say:

```json
{
  "chrome": {
    module: "@bigtest/webdriver",
    options: {
      browserName: "chrome",
      headless: true
    }
  }
}
```

This will allow us to create a node driver that can be configured with:

```json
{
  "node": {
    module: "@bigtest/nodedriver",
    options: {
      version: 12
      unhandledPromiseRejection: 'strict'
    }
  }
}
```

To implement the driver, the module specified should have a `create` export that implements the `DriverFactory` interface.


packages/driver/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "private": true,
```
cowboyd, line 2: We can remove this once we think that it is a good idea.

packages/driver/index.ts
```diff
@@ -0,0 +1,48 @@
...
+  connect(agentURL: string): Operation<void>;
+}
+
+export interface DriverFactory<TOptions, TData> {
+  (spec: DriverSpec<TOptions>): Operation<Driver<TData>>;
+}
+
+export function load<TOptions, TData>(spec: DriverSpec<TOptions>): Operation<Driver<TData>> {
+  try {
+    let exports = require(spec.module);
```
jnicklas, line 20: Should this use `import` rather than `require`? I'm a bit unsure of which is preferred?

jnicklas:
I like it! Very modular, and also paves a way forward for something like a native driver which for example launches an iOS Simulator or similar.

packages/driver/index.ts
cowboyd: The primary difference is that `import` is async, whereas `require` is synchronous which would mean that we'd have to make `load` itself an async operation. This would definitely be more flexible from a future-proof standpoint in the event that we wanted to do some stuff like lazily initialize plugins or something.

Issue #156, opened 2020-02-10
Add an interactive init command to bootstrap an application


jnicklas:
Similarly to `yarn init` we should have an interactive command which guides the user through the steps needed to get going with bigtest.

Issue #138, opened 2020-02-21
How to "arrange" the browser environment from the test-case


cowboyd:
In some cases, we want to begin the browser in a particular state before booting the application and navigating to a url. The most common case of this is setting up local storage with authentication information so that the app begins its run. However, some tests you'll want to actually test login/logout and you _won't_ want have local storage setup in a given way.

The problem is how to specify this on a per-test basis. The reason this requires special attention is that some actions would need to run _before_ the app boots, whereas other run _after_. 

The harness itself is loaded _before_ the app, so how do we mark certain steps as running immediately, before anything else? Also, these steps cannot be asynchronous, because the app _will_ continue loading once the `harness.js` script is finished executing.

Another option is to do some two-phased loading where we prepare the iframe first, and then proceed to load the app, but I'm not sure exactly how this will work.

jnicklas:
Uff, this is a tough issue. I have some ideas for workarounds, but this is a potential showstopper we really haven't paid enough attention to. 🤯

cowboyd:
This has been fixed by making the agent and the harness running on the same domain #377

PR #317, opened 2020-06-08
Consume fetch from effection

minkimcello:
## Motivation
To complete the refactor of [#138](https://github.com/thefrontside/effection/pull/138) and [#156](https://github.com/thefrontside/effection/pull/156) from `@thefrontside/effection`. We moved out `webdriver/fetch` to `effection/fetch` and in the process implemented `cross-fetch` and moved away from `node-fetch` to offer multi-browser compatibility.

## Approach
- Removed `fetch` from `webdriver`.
- Added `effection/fetch` as a dependency to `webdriver`.
- Updated imports to wherever we were calling `fetch`.
- Ran tests.

## TODOs
- [x] _Can I assume everything is fine if all the tests are passing?_ I can.


.eslintrc.json
```diff
@@ -1,3 +1,3 @@
 {
-  "extends": "@frontside"
+  "extends": "@frontside/eslint-config"
```
cowboyd, line 2: Is this out of band for this change?

packages/webdriver/src/until-url-available.ts
```diff
@@ -12,7 +12,7 @@ export function* untilURLAvailable(url: string, maxWait: number): Operation<void
   });
   while (true) {
     try {
-      let response: Response = yield fetch(url);
```
cowboyd, line 15: It's still a good idea not to remove this annotation

.eslintrc.json
minkimcello: Oh, it's something that I just came across. How does eslint know which config to take just by extending `@frontside`?

cowboyd:
> @cowboyd Can I assume everything is fine if all the tests are passing?

yep. fetch is a core mechanic and would cause tests to fail epically if not working properly.

PR #326, opened 2020-06-10
Make atom subcribable

cowboyd:
Motivation
-----------
It's easier to implement the `once` and `each` method using subscriptions.

Approach
----------
This just makes `Atom` subscribable, and then refactors current Atom operations to consume itself as a subscription.

Follow-on Work
--------------

- The `each` method is no longer even needed since it is really just a
  wrapper around `forEach`. It should be deprecated and removed.

### DIFF
--- a/packages/atom/src/atom.ts
+++ b/packages/atom/src/atom.ts
@@ -1,7 +1,7 @@
 import { EventEmitter } from "events";
 import { Operation } from "effection";
 import { on } from '@effection/events';
-import { Subscribable, SymbolSubscribable } from '@effection/subscription';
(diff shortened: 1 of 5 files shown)


packages/atom/src/atom.ts
```diff
@@ -1,14 +1,18 @@
...
-export class Atom<S> {
+export class Atom<S> implements Subscribable<S,void> {
   initial: S;
   state: S;
 
   subscriptions = new EventEmitter();
 
+  states = Subscribable.from(on(this.subscriptions, 'state'))
+    .map(([state]) => state as S)
+
```
jnicklas, line 15: This is like the perfect use case for channels, see https://github.com/thefrontside/effection/pull/144.

PR #329, opened 2020-06-12
Nodedriver

cowboyd:
Implement an in-process node driver. This is spike PR from which we'll be drawing more digestible chunks.

#### Bundling
 After some research, it appears that we will need to provide a separate bundling mechanism for the web and for node. The problem is that parcel bundles web stubs of node modules. In reality though, when you require `fs` we don't want a bundled `fs`, we want the *real* `fs`. We need to think about how we transmit test bundles to agents.


cowboyd:
I'm going to go ahead and close this since we can always have a look back on the branch

cowboyd:
At some point, we'll want to tackle this, but probably not in 2020

PR #332, opened 2020-06-16
Merge proxy and agent servers

jnicklas:
This completes the architectural changes needed for #286. Now we only need to actually move some of the harness code over to the agent.


jnicklas:
@pittst3r decided that instead of doing crazy type magic to wrap the inner Express object, like we were stuck on yesterday, we could just give direct access to the underlying express object as an escape hatch for adding non-effection middleware and stuff.

packages/server/src/orchestrator.ts
```diff
@@ -61,10 +54,10 @@ export function* createOrchestrator(options: OrchestratorOptions): Operation {
   })
 
   yield fork(createProxyServer({
+    agentServerConfig,
     delegate: proxyServerDelegate,
     port: options.project.proxy.port,
     targetPort: options.project.app.port,
-    inject: `<script src="${agentServerConfig.harnessUrl()}"></script>`,
```
jnicklas, line 67: I decided to remove this argument. Previously the proxy server was just concerned with injecting something into the proxied page, but now it is actually much more specific to bigtest anyway, so we might as well generate this in the proxy server itself.

cowboyd:
@jnicklas Can you add a changeset for this?

PR #336, opened 2020-06-16
upgrade effection

cowboyd:
Motivation
-----------

We recently changed effection to use a [unique symbol][1] which should make the need to have it as a peer dependency everywhere obsolete. 

Approach
----------
This upgrade effection to the "universal symbol" version and then upgrades all bigtest packages to consume that version as a normal dependency.


cowboyd:
And what a yak it was!

PR #342, opened 2020-06-18
Add unit tests for `Atom` class

pittst3r:
## Motivation

There were no tests for this class.

## Approach

- Add unit tests
- Make some properties on `Atom` private since they shouldn't be public

### DIFF
--- a/packages/atom/test/atom.test.ts
+++ b/packages/atom/test/atom.test.ts
@@ -1,0 +1,155 @@
+import { describe, it } from 'mocha';
+import * as expect from 'expect';
+import { Atom } from '../src/atom';
+import { spawn, converge, never } from './helpers';
+import { Slice } from '../src/slice';
+
+describe('@bigtest/atom', () => {
+  describe('Atom', () => {
+    let subject: Atom<string>;
+
+    beforeEach(() => {
+      subject = new Atom('foo');
+    })
+
+    describe('.get()', () => {
+      it('gets the current state', () => {
+        expect(subject.get()).toEqual('foo');
+      });
+    });
+
+    describe('.update()', () => {
+      beforeEach(() => {
+        subject.update(previous => {
+          expect(previous).toEqual('foo');
+          return 'bar';
+        });
+      });
+
+      it('updates the current state', () => {
+        expect(subject.get()).toEqual('bar');
+      });
+    });
+
+    describe('.each()', () => {
+      let result: string[] = [];
+
+      beforeEach(() => {
+        spawn(subject.each(function*(state) {
+          result.push(state);
+        }));
+
+        subject.update(() => 'bar');
+        subject.update(() => 'baz');
+      });
+
+      it('performs given operation for each state change', async () => {
+        await converge(50, () => {
+          expect(result).toEqual(['bar', 'baz']);
+        });
+      });
+    });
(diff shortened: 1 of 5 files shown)


packages/atom/test/atom.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+      beforeEach(async () => {
+        subject = new Atom({ foo: { bar: "baz" } });
+        result = subject.slice(['foo', 'bar']);
+      });
+
+      it('returns a slice of the Atom with the given path', async () => {
+        expect(result.get()).toEqual('baz');
+      });
+    });
```
jnicklas, line 153: Since the implementation of `Slice` is in a separate file, should we maybe move the tests to a separate file as well?

packages/atom/test/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export function spawn<T>(operation: Operation<T>): Promise<T> {
+  return currentWorld.spawn(operation);
+}
+
+async function wait(ms: number) {
+  await new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+export async function converge<T>(timeout: number, fn: () => T): Promise<T> {
```
cowboyd, line 24: My prefered API is to have this just default to the mocha timeout, that way, you don't have to worry about it. Also, what about calling it `when`

packages/atom/test/atom.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+          subject.reset();
+        });
+
+        it('resets to the initial value', async () => {
+          expect(subject.get()).toEqual('foo');
+        });
+
+        it('removes listeners', () => {
+          expect(subject['subscriptions'].listenerCount('state')).toEqual(0);
```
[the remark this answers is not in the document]
pittst3r: Ah yes, this test shouldn't have been in there. The `removing listeners` tests should be testing the unsubscription. What do you say to introducing this refactor in a separate PR?

packages/atom/test/helpers.ts
pittst3r: Yeah, both of those changes seem good to me. I actually just copied this from another bigtest package's test suite; should we make a new package that has this function?

packages/atom/test/atom.test.ts
pittst3r: I was trying to test that `.slice()` successfully returned a slice object with the given path beyond just using `instanceof`. In light of that does this seem okay or is there a better way to test `.slice()`?

packages/atom/test/helpers.ts
cowboyd: Yeah, or maybe just a symlink to a directory in the project root? For now, let's copy-pasta :)

pittst3r: I made these changes in place.

packages/atom/test/atom.test.ts
```diff
@@ -44,7 +44,7 @@ describe('@bigtest/atom', () => {
       });
 
       it('performs given operation for each state change', async () => {
-        await converge(50, () => {
+        await when(50, () => {
```
cowboyd: Can we remove the timeout?

PR #345, opened 2020-06-20
Add streaming output of test results

jnicklas:
Depends on #344 

This allows us to finally actually output some decent test results! 🥳

Screenshot 2020-06-20 at 14 16 03


jnicklas:
Merged this! There is more work to do in this area, but it's such a huge improvement that merging this now is very valuable.

PR #350, opened 2020-06-23
Move to `v*` branches instead of `master`

### LINKED ISSUE #349: Move to`v*` branches and away from `master` branch verbiage
Frontside has decided to move away from `master` branch verbiage in order to be more welcoming to all employees and contributors.

pittst3r:
Closes #349 

## Motivation

`master` can be a violent word and make people not feel welcome as an employee or contributor.

## Approach

Rather than simply changing the name of the default branch, we took a step back and looked at what could be a better branching strategy in general. We've landed on having major release branches (such as `v0`, `v1`, etc.) and setting the default to the latest major release branch. When we begin working on a new major release we:

1. Make a new branch for that version
1. Set the default branch to that version

Note: we will need to change the changeset `config.json` `baseBranch` option manually for each new major release branch.

### To-do

- [ ] Make a new branch off of `master` called `v0` and set it as the default
- [ ] Point existing PRs to new `v0` branch

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -4,6 +4,6 @@
   "commit": false,
   "linked": [],
   "access": "restricted",
-  "baseBranch": "master",
+  "baseBranch": "v0",
   "updateInternalDependencies": "patch"
 }
--- a/.github/workflows/changeset.yml
+++ b/.github/workflows/changeset.yml
@@ -3,7 +3,7 @@ name: Changeset
 on:
   push:
     branches:
-      - master
+      - 'v[0-9]+'
 
 jobs:
   create-release-pr:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,7 +3,7 @@ name: Lint
 on:
   pull_request:
     branches:
-      - master
+      - 'v[0-9]+'
 
 jobs:
   build:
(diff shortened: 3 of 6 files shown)


cowboyd:
👍  how should we coordinate the merging? I'm thinking we should call out this change in discord to give everyone a heads up and then merge it on Thursday or Friday?

pittst3r:
> how should we coordinate the merging?

I think we can follow these steps in quick succession, perhaps tomorrow morning while @jnicklas is around:

1. Stop merging PRs
1. Make new `v0` branch
1. Switch default to `v0`
1. Update this PR to be against `v0`
1. Merge this PR

> I'm thinking we should call out this change in discord to give everyone a heads up and then merge it on Thursday or Friday?

Should definitely call it out on discord, not sure we need to wait that long though? I feel like we can just do it tomorrow.

.changeset/config.json
```diff
@@ -4,6 +4,6 @@
   "commit": false,
   "linked": [],
   "access": "restricted",
-  "baseBranch": "master",
+  "baseBranch": "v0",
```
cowboyd, line 7: What is baseBranch used for I wonder? I hope we can still release to multiple branches.

pittst3r: This file will need to be different in each version branch, but I don't think that's a problem?

pittst3r: Here is what that option is for though: https://github.com/atlassian/changesets/blob/0f0b6ad10744de9a8a32305796bd78d33bb1a426/packages/cli/CHANGELOG.md#minor-changes-5

PR #355, opened 2020-06-24
use @frontsidejack to create changset release pull requests

cowboyd:
Pull requests created with the default github actions bot do not trigger checks and workflows (in addition to being not branded).

This uses an access token of @frontsidejack to allow running workflows on the PRS that are created as part of the changeset release


minkimcello:
There's something funky going on when changeset is run in the same workflow as synchronize-with-npm. I think it's because the changeset action checks out to a different branch and that's persisting. Having a look into it now.

Issue #315, opened 2020-06-05
Prevent interactor actions when running an assertion


jnicklas:
The test runner can be either running steps or assertions. Assertions are meant to be side-effect-free and run in parallel. That means we should take measures to prevent accidentally running side-effectful actions during assertions.

PR #357, opened 2020-06-24
Add globals package for shared global variables

jnicklas:
We want bigtest components to be able to share configuration with each other, without having to explicitly call into each other, and with zero configuration. The only way to enable this in today's JavaScript is through shared global variables. To make this a bit nicer and safer, we add a package which can be used to access these variables in a typesafe and convenient way.

This paves the way for #315 and also for completing #286.


jnicklas:
Two things:

1) the changeset only affects `@bigtest/interactor`, while there are changes to other packages, they are purely internal, it seems to me that internal changes do not require a changeset? `@bigtest/globals` is a new package, so does not need a changeset, right?

2) this uses [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) which is not available on IE, but is polyfillable should we need to.

jnicklas:
The package name is bikesheddable, of course.

Also, this could be part of `@bigtest/project`, but I felt that it would be better encapsulated for this to be its own package. `@bigtest/interactor` for example has no need to depend on any of the other stuff in `@bigtest/project`. Also it draws a clearer line between the two, since their purpose is quite different.

.changeset/globals.md
```diff
@@ -0,0 +1,4 @@
+---
+"@bigtest/interactor": minor
```
cowboyd, line 2:
```suggestion
"@bigtest/interactor": minor
"@bigtest/server": patch
```

Issue #310, opened 2020-06-02
How do we express filter conditions on interactors?


jnicklas:
Some interactors will probably want to have pre-conditions or filters, which we want to use to narrow the scope of what to find. For example, an input field might be disabled. Interacting with a disabled field is not what we want, probably. So we'll want to exclude disabled fields from our search. But do we always want to exclude these fields? What if we DO want to use a disabled field, can we work with both from within the same interactor? Also, it would be pretty nice if we could get error messages like "tried to fill in field 'Email' but it was disabled" or something like this.

Capybara has [such a feature](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selector/definition/button.rb#L35-L36), but it is a bit cumbersome, and the error messages generated aren't great. If we want to have this type of feature, we should take care to design it in such a way that it is intuitive and also generates great errors.

cowboyd:
Maybe some sort of "constraint" API?

```ts
Button("Submit")
  .validate(assert, element => {
    assert(element.enabled, "must be enabled in order to click, but it was disabled"{
  })
  .click();
```

In practice though, it seems like most of the time you would want to add these constraints to the action itself, so that it would be defined within the `click()` action.

Also, it would be great to somehow have access to these contraints before test-run time. I'm not sure how that would work.

jnicklas:
Hmm, interesting to add these constraints on the action, are there any cases where a filter should be allowed on one action, but not on another? I can't really think of any of the top of my head, but there might be cases?

I also, just started heading down a rabbit hole with regards to visibility. Capybara by default only matches "visible" elements. This is actually hugely helpful, since JS-heavy apps often do stuff like hiding content via CSS and this might prevent an element from being shown on the page, which means that a user wouldn't be able to interact with it, so we don't allow it either, though it is overrideable in the rare case where you DO want to interact with a hidden element.

The question then is what does "visible" even mean? Selenium has a `isDisplayed` method which tries to determine visibility of the element, and this is probably a good starting point, since it's been proven to be reasonably robust. With some digging I found their [implementation here](https://github.com/SeleniumHQ/selenium/blob/e09e28f016c9f53196cf68d6f71991c5af4a35d4/javascript/atoms/dom.js#L437-L518).  Selenium is Apache2 licensed, so I think with proper attribution we would be allowed to re-use this. This method is obviously pretty complicated, but I think that given that it's just a bunch of if-statements, it should be pretty fast, so I'm not too worried about it.

[...]

PR #359, opened 2020-06-24
Add filters to interactors

jnicklas:
Filters allow the selection of elements to be narrowed. A default value can also be provided. See #310 

I intentionally left this functionality rather minimal.

This also adds the ability to check for filters via an assertion on an interactor like this:

``` typescript
await TextField("Email").has({ value: "test@example.com" });
```

IMO this reads a little better than:

``` typescript
await TextField("Email", { value: "test@example.com" }).exists();
```

### DIFF
--- a/.changeset/interactor-filters.md
+++ b/.changeset/interactor-filters.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/interactor": minor
+---
+
+Can filter interactors to narrow selection without having to define a separate interactor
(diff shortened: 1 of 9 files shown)


cowboyd:
Do we retain the set of elements that would have been matched _without_ the filters in place? I'm thinking of the error messaging: "Unable to find TextField with 'text@examplec.om'. Did you mean `text@example.com?"

jnicklas:
We do have a reference to these elements when we generate the error message:

https://github.com/thefrontside/bigtest/blob/1e10456f8e8645e9c32a5775992b9e0520f0fc72/packages/interactor/src/interactor.ts#L44-L54

However, we're not actually using this information yet. Actually generating good error messages for this is pretty challenging, so I've sort of put that aside for now. The problem is that we have a *lot* of information, and displaying this information in a way which makes sense to the user requires some design work.

PR #366, opened 2020-06-25
Interactor is generic over its specification

jnicklas:
Currently an interactor is typed `Interactor<E extends Element>`, this changes the type to `Interactor<E extends Element, S extends InteractorSpecification<E>>` that is the type of an interactor class and its instance includes a reference to the type of the specification that generated it.

The reason that this is useful is because it allows methods on the interactor to refer to the specification. As part of #359, I would like to add the ability to check for a given filter via `has({ filter: whatever })`, but this requires knowledge of the type of the filter specification to make it type-safe.

Why didn't we do this in the first place then? It does have some downsides. The type of the specification is not easily namable, so when an interactor needs to be referenced somewhere where its type must be explicitly given, then that is not easily done. Also, previously we had not yet worked out how to work around the variance issues that crop up around `find`, so we were unable to make this work.

### DIFF
--- a/.changeset/interactor-specification-type.md
+++ b/.changeset/interactor-specification-type.md
@@ -1,0 +1,4 @@
+---
+"@bigtest/interactor": minor
+---
+Interactors are generically typed over the specification that they use.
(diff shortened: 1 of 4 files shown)


cowboyd:
Does this add any cognitive overhead on how you define interactors?

jnicklas:
@cowboyd no, the definition of the interactor is not affected at all. Where it could add some overhead is in storing references to interactors, for example if you'd wanted to have an array of interactors for some silly reason. The type erasure we had makes the type of an interactor easier to name.

PR #368, opened 2020-06-25
WIP: Parcel upgrade

pittst3r:


packages/parcel/src/parcel-run.ts
```diff
@@ -1,13 +1,20 @@
...
+};
+
 yargs
   .command('$0 [files..]', 'run the parcel server', x => x, (argv) => {
-    let options: ParcelOptions & { port?: number } = JSON.parse(argv.options as string);
+    let options: ParcelOptions = JSON.parse(argv.options as string);
+    let ctx = main(createParcelServer(argv.files as string[], options));
 
-    main(createParcelServer(argv.files as string[], options));
+    process.on('SIGTERM', () => {
```
cowboyd: This should all be handled by the [`main()` method from `@effection/node`](https://github.com/thefrontside/effection/blob/master/packages/node/src/main.ts) which will halt the `createParcelServer` operation on both `SIGINT` _and_ `SIGTERM` If this isn't happening, then it's definitely a problem.

packages/parcel/src/parcel-server.ts
```diff
@@ -1,53 +1,44 @@
...
+  let parcel = new Parcel({ entries: entryPoints, ...options });
 
-  let middleware = bundler.middleware();
-  let server = createServer(middleware)
+  let events = new Mailbox();
 
-  try {
-    yield throwOnErrorEvent(server);
+  yield spawn(function*() {
+    let subscription: ParcelAsyncSubscription = parcel.watch(() => events.send({ type: 'update' }));
```
cowboyd:
```ts
let subscription: ParcelAsyncSubscription = yield parcel.watch(() => events.send({ type: 'update' }));
```

This will let you actually get the subscription, and then call it directly in your finally block. Also, if the promise returned by `watch` fails at this point, an exception will be thrown into your operation, which is what you want.

Later, you won't need to unbox the promise of the subscription:

```ts
try {
  yield;
} finally {
  subscription.unsubscribe();
}
```

PR #371, opened 2020-06-26
Eliminate `createInteractor` partial application API

pittst3r:
## Motivation

The `createInteractor()()` partial application API does not provide any benefit, and may be confusing for a new user since it serves no purpose.

## Approach

Squash the two functions into one, resulting in an API like:

```ts
createInteractor('paragraph', { selector: 'p' });
```

### DIFF
--- a/.changeset/elim-interactor-partial-application.md
+++ b/.changeset/elim-interactor-partial-application.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/interactor": minor
+---
+
+Eliminate `createInteractor()()` partial application API so we can provide all
+arguments to a single function.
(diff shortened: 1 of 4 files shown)


jnicklas:
@pittst3r It seems that the reason this didn't fail was that the inferred types didn't work as intended in the first place, and were overly permissive. I've addressed this in #374 and also added some type tests using dtslint to verify that the types behave as intended. Rebasing this on #374 there are a bunch of test failures which should show why the currying was needed in the first place. Unfurtunately, unless there are language changes in TypeScript, the currying is the most convenient way of solving this issue for the time being.

PR #373, opened 2020-06-26
Getting manifest-builder to copy over the map file and updating the sourceMappingURL

### LINKED ISSUE #176: fingerprinted manifest source map url is incorrect.
We're server each manifest at it's own fingerprinted location like `http://localhost:3000/manifest-xxyyzz.js` but the sourceMap url at the bottom is like:

```
//sourceMapURL=manifest.js
```

It should have the same name as the fingerprinted file. (and also we need to make sure and copy the source map into place

minkimcello:
## Motivation
To resolve #176. The orchestrator wasn't copying over the source map file to its dist directory, and [debugging](https://developer.mozilla.org/en-US/docs/Tools/Debugger/How_to/Use_a_source_map) isn't very descriptive without the `*.js.map` file.

## Approach
- Once parcel generates the map file into the `build\`, we copy it over to `dist\` with its fingerprint included in the filename.
- Afterwards we update the manifest's `sourceMappingURL` so that it looks like this:
Screen Shot 2020-06-26 at 5 06 46 PM

## TODOs
- [x]  Write tests.
- [x] Add changesets.


packages/server/src/manifest-builder.ts
```diff
@@ -21,14 +21,31 @@ interface ManifestBuilderOptions {
   distDir: string;
 };
 
+function* updateSourceMapURL(filePath: string, sourcemapName: string){
+  let { size } = fs.statSync(filePath);
+  let readStream = fs.createReadStream(filePath, {start: size - 16});
+  readStream.on('data', currentURL => {
```
cowboyd: When working with effection, anything asynchronous needs to be an operation. The reasons are many, but the most important ones are:

1. If an error occurs while doing the url substitution, then we need to raise an error in the `updateSourceMapURL` operation.
2. The `updateSourceMapURL` operation, like _any_ effection operation, can be halted at any time. As we have it here, the `readStream.on` callback will continue to fire, even if the caller has gone out of scope for whatever reason.
3. If either 1 or 2 happens, we need to unregister the `on('data')` callback, otherwise it will be firing into dead code.

The same applies to the call to `truncate()`

You can use the `on` operation from `@effection/events`, and the `fs.promises` api to use operations everywhere:

```ts
import { on } from '@effection/events';

function* updateSourceMapURL(filePath: string, sourcemapName: string){
   let { size } = fs.statSync(filePath);
   let readStream = fs.createReadStream(filePath, {start: size - 16});
   let currentURL = yield on(readStream, 'data');
   if (currentURL === '/manifest.js.map') {
     yield fs.promises.truncate(filePath, size - 16);
     fs.appendFileSync(filePath, sourcemapName);
   }
}

File: packages/server/src/manifest-builder.ts [hunk 2]
```diff
@@ -21,14 +21,31 @@ interface ManifestBuilderOptions {
...
+    if(currentURL == '/manifest.js.map'){
+      fs.truncate(filePath, size - 16, () => {
+        fs.appendFileSync(filePath, sourcemapName);
+      })
+    }
```
cowboyd: We've handled the happy path of finding what we're looking for, what's to be done if we _don't_ find the sourcemapping URL?

cowboyd (changes requested):
Nice to see it coming together. 

I see that you've marked a todo for writing test cases. As you write them, shoot for answering the question of "if the bundler output changes it subtle ways, which tests will fail immediately to let me know exactly what is going on?"

packages/server/src/manifest-builder.ts
minkimcello: Oh I see. That makes sense. 👍

packages/server/src/manifest-builder.ts
minkimcello: Since `soucemapName` is `${fileName}.map`, would it be appropriate to do
```
if(currentURL == '/manifest.js.map' && sourcemapName != 'undefined.map' ){
  yield truncate(filePath, size - 16);
  fs.appendFileSync(filePath, sourcemapName);
};
```

cowboyd: I don't think so. I'm wondering what happens in an `else` branch of this `if` statement.

minkimcello: Once the tests are in place, do we need to do anything in the `else`?

cowboyd: If we don't find the sourcemapping url, then what does that mean?

minkimcello: Oh you mean the URL from within `manifest.js`? The two cases I can think of would be if the `manifest.js` is missing from `dist/` or the manifest is built differently than what we're expecting.

packages/server/src/manifest-builder.ts
minkimcello: When I do `let currentURL = yield on(readStream, 'data')`, I get back `ChainableSubscription`. How do I retrieve the value?

minkimcello: Does this look okay?

```ts
function* updateSourceMapURL(filePath: string, sourcemapName: string){
  let { size } = fs.statSync(filePath);
  let readStream = fs.createReadStream(filePath, {start: size - 16});
  let onRead = yield on(readStream, 'data');
  let currentURL = yield onRead.next();
  
  if( currentURL.value.toString('utf-8') == '/manifest.js.map'){
    yield truncate(filePath, size - 16);
    fs.appendFileSync(filePath, sourcemapName);
  };
}
```

packages/server/src/manifest-builder.ts
cowboyd: Yeah, those are the two cases I can think of too. If the `manifest.js` is missing, then the copy from the `build` -> `dist` will fail earlier on, so we can be relatively safe in the assumption that it is there. My question is what should we do when the manifest is built differently than what we're expecting?

minkimcello: Can we rely on the tests to catch that or do we need something in between for cases when manifests gets built correctly in tests but not in production?

cowboyd: We have to handle this case in production, since it indicates something is very wrong.

packages/server/src/manifest-builder.ts
```diff
@@ -28,10 +28,16 @@ function* updateSourceMapURL(filePath: string, sourcemapName: string){
   let onRead = yield on(readStream, 'data');
   let currentURL = yield onRead.next();
```
cowboyd: You should use `once` here, which accomplishes the same thing without needing to go through an event stream. Instead, it just gets the aguments of the `data` event:

```ts
let [currentURL] = yield once(readStream, 'data');
```

packages/server/src/manifest-builder.ts
```diff
@@ -28,10 +28,16 @@ function* updateSourceMapURL(filePath: string, sourcemapName: string){
...
-  if( currentURL.value.toString('utf-8') == '/manifest.js.map'){
-    yield truncate(filePath, size - 16);
-    fs.appendFileSync(filePath, sourcemapName);
-  };
+  try {
+    if( currentURL.value.toString('utf-8') == '/manifest.js.map' ){
+      yield truncate(filePath, size - 16);
+      fs.appendFileSync(filePath, sourcemapName);
+    } else {
+      throw "Unexpected placement and or formatting of sourceMappingURL in the manifest"
```
cowboyd: While JavaScript lets you throw _anything_ It's always a good idea to use an `Error` or subclass of `Error` for the main reason that you get a uniform API for handling errors at any point in the code, including a helpful message and a stack trace.

```suggestion
      throw new Error(`Expected a sourcemapping near the end of the generated test bundle, but found "${currentURL}" instead`);
```

Also, I'm not sure if we should be throwing a hard error, or just a warning. what do you think?

packages/server/src/manifest-builder.ts
```diff
@@ -28,10 +28,16 @@ function* updateSourceMapURL(filePath: string, sourcemapName: string){
...
+    };
+  } catch (error) {
+    throw error;
```
cowboyd: You don't need a try catch here since all it does is propagate any errors.

cowboyd (changes requested):

packages/server/src/manifest-builder.ts
minkimcello: I would think a warning would be more appropriate than a hard error because a mismatch in the sourcemapURL would just mess up the reporting aspect of BigTest and not the tests itself, right?

cowboyd (commented):
One last question: is it possible to test the failure case?

minkimcello:
@cowboyd I don't know if this is the right approach but we could export `updateSourceMapURL()` and then pass in an empty file to force the `else` statement.
```
// updateSourceMapURL()

if( currentURL == '/manifest.js.map' ){
  yield truncate(filePath, size - 16);
  fs.appendFileSync(filePath, sourcemapName);
} else {
  console.warn(`Expected a sourcemapping near the ...`);
  return 1;
};
```
```
// manifest-builder.test.ts

await writeFile('tmp/manifest-builder/fakeManifest.js', '');
expect(await updateSourceMapURL('tmp/manifest-builder/fakeManifest.js', '')).toBe(1);
```

I'm also not entirely sure if we can unit test Operations like that.

packages/server/src/manifest-builder.ts
cowboyd: Hmmm.... I see that, but on the other hand, it means that something is wrong with the "system" and maybe we should be doing a hard fail.

we can always decide to catch the hard fail higher up and log a warning there. After seeing this in action, let's go with an error. A nice knock-on side-effect of this is that it will make testing the failure condition much eaiser.

packages/server/test/manifest-builder.test.ts
```diff
@@ -85,8 +87,14 @@ describe('manifest builder', () => {
     });
     it('updates the sourcemapURL of dist manifest with fingerprinted file', () => {
       expect(distMapURL).toMatch(/manifest-[0-9a-f]+\.js.map/);
-    })
-  })
+    });
+    it('warning when sourcemapURL is not generated at the bottom', async () => {
+      // wip
+      await writeFile('tmp/manifest-builder/fakeManifest.js', 'hello');
+      let error = await updateSourceMapURL('tmp/manifest-builder/fakeManifest.js', '');
```
cowboyd: I think it's fine to have a fixture for this that is always there, lke `test/fixtures/manifest-without-source-mapping.js`

Also, in JavaScript we should use normal error handling rather than use error codes. In other words `updateSoureMapURL` should raise an exception.

minkimcello: I'll just use the current `test/fixtures/empty.js`.

How can I test for an `Error`? I'm having difficulty getting an Operation to return anything but an empty object.

cowboyd: You have to run the operation. This is what the `actions.fork` helper is for. It will convert the operation into a promise that resolves to the result of running the operation, or rejects if an operation fails.

```ts
expect(actions.fork(updateSourceMapURL('test/manifest/empty.js').rejects.toMatch({message: "blah blah"});
```

cowboyd (approved):
Looks great!

PR #374, opened 2020-06-28
Add tests for interactor types

jnicklas:
Some of the properties of the interactor types are not verifiable through the test suite, this adds some type tests using dtslint to verify that the types behave as we intend.

It turns out that we did not actually maintain the properties we wanted, due to using `Partial<InteractorSpecification>` which messed up the inferred types somehow.

### DIFF
--- a/packages/interactor/src/create-interactor.ts
+++ b/packages/interactor/src/create-interactor.ts
@@ -1,13 +1,13 @@
-import { defaultSpecification, InteractorSpecification, InteractorInstance, InteractorType } from './specification';
+import { InteractorSpecification, InteractorInstance, InteractorType, LocatorFn } from './specification';
 import { Locator } from './locator';
 import { Interactor } from './interactor';
 import { interaction } from './interaction';
 import { converge } from './converge';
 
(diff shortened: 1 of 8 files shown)


packages/interactor/src/create-interactor.ts
```diff
@@ -1,14 +1,14 @@
...
+import { InteractorSpecification, InteractorInstance, InteractorType, LocatorFn } from './specification';
 import { Locator } from './locator';
 import { Interactor } from './interactor';
 import { interaction } from './interaction';
 import { converge } from './converge';
 import { defaultOptions } from './options';
 
-export function createInteractor<E extends Element>(interactorName: string) {
-  return function<S extends InteractorSpecification<E>>(specification: Partial<S>): InteractorType<E, S> {
-    let fullSpecification: InteractorSpecification<E> = Object.assign({ selector: interactorName }, defaultSpecification, specification);
```
jnicklas: Also nice to be able to get rid of this ugly line :)

pittst3r (approved):

PR #375, opened 2020-06-28
Add initial draft of flaky test blogpost

jnicklas:
This is an initial draft of my blog post on test flakiness. I felt some inspiration today and had to jot this down. Any feedback on this is greatly appreciated!

[Rendered](https://github.com/thefrontside/bigtest/blob/flakiness-blog-post/docs/flakty-tests.md)

### DIFF
--- a/docs/flakty-tests.md
+++ b/docs/flakty-tests.md
@@ -1,0 +1,260 @@
+# Never write a flaky test again!
+
+For the past few months, Frontside has been working on BigTest, a new and
+revolutionary testing framework. Testing the frontend of complex applications
+is often associated with nightmarishly inconsistent test suites, where
+seemingly randomly occurring failures cast doubt on the value of the tests
+themselves and frustrate developers to no end.
+
+Our mission with BigTest has been to eliminate flakiness in tests altogether
+and provide the most stable, most easy to use, and best performing test
+framework out there. We'll be ready to share more about BigTest soon, but for
+the moment I would like to talk about the approach we've taken to eliminate
+flakiness, and also what you need to know to write tests which are solid as a
+rock and never fail randomly.
+
+The approach we have taken is not novel. I am the original author of
+Capybara, an acceptance test framework written in Ruby, which has a
+reputation of being incredibly solid and resistant to flakiness. We have
+taken all of the lessons learned from tweaking Capybara over years and
+applied them to build the Interactor API, a powerful way of defining
+interactions with your application which allows you to write tests which are
+even more solid than what Capybara can provide.
+
+However, eliminating flakiness in the framework is only part of the solution,
+sometimes the issue is not with the framework, but with how we write our
+tests, and inherent ambiguities in the way we write them. Eliminating these
+ambiguities is what allows you to write very stable tests.
+
+Let's take a deep dive into our strategy for avoiding flakiness.
+
+## Let's wait!
+
+The approach we've taken is simple: for any interaction we want to perform,
+or any assertion we want to check, we will try it and see if it succeeds, if
+it does not, we will wait a little while and try again. We will keep doing
+this until we either succeed, or some set amount of time has expired and we
+give up and fail.
+
+This surprisingly simple strategy is enough to eliminate flakiness, provided
+that the timeout is sufficienlty long, but there are of course a whole number
+of details that need to be done right for this to work as intended.
+
+We are calling this strategy "convergence", and a function which exhibts this
+waiting behaviour is called "convergent".
+
+### Retaining references
+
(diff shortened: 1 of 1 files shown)


docs/flakty-tests.md
```diff
@@ -0,0 +1,260 @@
+# Never write a flaky test again!
```
jorgelainfiesta: How about:

_BigTest Interactors: helping you never write a flaky test again!_ 

Just to give some context right from the beginning

docs/flakty-tests.md
```diff
@@ -0,0 +1,260 @@
...
+even more solid than what Capybara can provide.
+
+However, eliminating flakiness in the framework is only part of the solution,
+sometimes the issue is not with the framework, but with how we write our
+tests, and inherent ambiguities in the way we write them. Eliminating these
+ambiguities is what allows you to write very stable tests.
+
+Let's take a deep dive into our strategy for avoiding flakiness.
+
+## Let's wait!
```
jorgelainfiesta: Maybe we can consider something like: 

_"Convergences": the waiting strategy to avoid flakiness_

docs/flakty-tests.md
```diff
@@ -0,0 +1,260 @@
...
+
+Both the locator and action functions are synchronous functions, which means
+that they do not release the event loop. This allows us to transform them
+into convergent functions on the interactor which perform the entire
+interaction of finding the element and runnign the functions against them.
+
+BigTest is written in TypeScript and our interactors are fully typesafe when
+used from TypeScript both in their definition and when using them.
+
+## Inherent ambiguities
```
jorgelainfiesta: Perhaps we can consider this?

_Further challenges: Inherent ambiguities_

docs/flakty-tests.md
```diff
@@ -0,0 +1,260 @@
...
+Which page will we end up on? Intuitively we would say that we should end up
+on the *third* page, but unfortunately the actual answer is "it depends". The
+first click on `Next` will start a request to the server which will complete
+after some time, this will change the `href` of the link. But if our second
+click happens *before* the request completes, we will navigate to the second
+page *twice*, and we actually end up on the *second* page.
+
+[GRAPHIC]
+
+### Guessing
```
jorgelainfiesta: To provide like a thread to the three subsections I suggest we append some descriptor to each approach. Preferably we'd use descriptors that go from worse to best. 

Something like:
- A naïve approach: Guessing
- A better approach: Anchoring
- The best approach: Fixing the UI

jorgelainfiesta (commented):
Hi Jonas!

I **loved** the article because it explains in simple terms a complex problem, and it goes up and down the technical detail ladder without getting lost. In my opinion, the article leaves the reader interested in knowing more about BigTest and trying it out.

I've only left some comments about the titles. In my experience, most people will just skim the article before realizing it's interesting or just scroll through before committing to reading it. Thus, it's best to leave big breadcrumbs for readers so they can decide to continue reading.

Would you mind if we published the article in the Frontside's blog? We don't have yet a channel set up in BigTest for this kind of content. But this article is pretty neat and I think we'll probably extract the upcoming BigTest guides from it soon.

Thanks!

docs/flakty-tests.md
jnicklas: Much better!

docs/flakty-tests.md
jnicklas: I love it!

docs/flakty-tests.md
jnicklas: I have changed all three of the big headlines into the format `Challenge: xyz`, I think this gives it some more easily digestible structure, let me know what you think!

docs/flakty-tests.md
jnicklas: Sounds great! I did change it to `Convergence` in the singular.

jnicklas:
Great suggestions @jorgelainfiesta, thanks!

jnicklas:
Also @jorgelainfiesta feel free to make any changes you like!

jorgelainfiesta:
Great Niklas, I just have a question about the structure.

Before the titles change, I had the impression that there were two topics in the article. First, convergence (under which you explained the challenge of building such approach: retaining references and taming the event loop). Second, inherent ambiguities (under which you listed the possible solutions). 

Now, it seems that the convergence it's a topic, and the other three are challenges. But I'm having a difficult time to understand the relationship that connects them under that structure.

jorgelainfiesta:
The edited article is going to be merged into Frontside's blog: https://github.com/thefrontside/frontside.com/pull/91

Thanks Jonas!

PR #377, opened 2020-06-30
Agent harness unification

jnicklas:
Previously the responsibility for running lanes of a test was delegated to the harness, this moves this responsibility to the agent. See #286. By moving this responsibility to the agent, we avoid losing state when we move to another page.

Essentially this makes the harness pretty useless at the moment. But I've neverless left all of the machinery for the harness in place, because I expect that we will need it once we want to do some more advanced setup where the agent shares its state with the harness.

### DIFF
--- a/.changeset/agent-harness-merge.md
+++ b/.changeset/agent-harness-merge.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/agent": minor
+---
+
+The running of the lanes has been moved from the harness to the agent.
--- a/.changeset/globals-test-frame.md
+++ b/.changeset/globals-test-frame.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/globals": minor
+---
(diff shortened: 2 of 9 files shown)


cowboyd (commented):
Looks good. Is the endgame here that our app interactor can use the test frame in order to explicitly load urls with a `navigateTo()` action?

This will definitely lock us into future parallelization at the agent level, in other words if we want to run 3 different lanes at a time, we'll need three agents..... but I think that's ok.

jnicklas:
@cowboyd yes, that is the idea. I'll follow up on this with an app interactor, but left it out of this PR to keep it smaller. It does indeed lock us into parallelization at the agent level, at least for the time being. This API is completely internal, so we should be able to change this in the future should we decide to do so.

cowboyd (approved):

PR #378, opened 2020-07-01
Add App Interactor, stop loading app automatically

jnicklas:
The agent no longer loads the app automatically for each test lane, rather the user will need to load the app manually via `App.load()` or `App.visit(path)` somewhere. These can be required from the interactor package like this:

``` typescript
import { App, Link } from '@bigtest/interactor';

export default test('some test')
  .step(App.load())
  .step(Link('whatever').click())
```

Depends on #377

### DIFF
--- a/packages/agent/app/test-frame.ts
+++ b/packages/agent/app/test-frame.ts
@@ -33,4 +33,8 @@ export class TestFrame {
     let { args: [message] } = yield this.mailbox.receive();
     return JSON.parse(message.data);
   }
+
+  clear() {
+    this.element.src = 'about:blank';
+  }
 }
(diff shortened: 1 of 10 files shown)


cowboyd (approved):
:shipit:

packages/agent/app/test-frame.ts
```diff
@@ -33,4 +33,8 @@ export class TestFrame {
     let { args: [message] } = yield this.mailbox.receive();
     return JSON.parse(message.data);
   }
+
+  clear() {
+    this.element.src = 'about:blank';
```
cowboyd, line 38: Is this syrchronous, is there any event we need to wait on to make sure this teardown is complete?

jnicklas: I've made it synchronous for now. We could wait for the element to load the `about:blank` page, but I'm not sure how useful that would be, it should be pretty much instant to load.

cowboyd (approved):

PR #384, opened 2020-07-02
Proxy fetch and XMLHttpRequest through agent frame

jnicklas:
This makes it possible to use mocking libraries like mirage from the test suite. By mocking XMLHttpRequest in the agent, we're also automatically using those mocks in the app/harness.


jnicklas:
I'd ideally like to add some tests for this, but I'm not really sure how and where they'd go?

cowboyd:
I don't imagine that we'd want to load the harness in anything but an iframe, but even uttering those words makes me feel like it's an inevitability at some point. Since for top-level frames, `window.parent` is actually a reference to itself, this indirection would result in an infinite loop. What if we wrapped these in a guard?

```js
if (window.parent !== window) {
  // implement redirect
}
```

jnicklas:
Yeah, good catch. Good to be defensive, and "surely it'll never..." seems to always come back to bite you in the end.

cowboyd (approved):

PR #389, opened 2020-07-03
Replace Parcel with Rollup for manifest builder

pittst3r:
## Motivation

I was initially attempting to upgrade to parcel@2 because:

- Typechecking
- Parcel errors if your app has parcel@2 installed while Bigtest is using parcel@1

After struggling with this I noticed:

- We are having to do gymnastics in order to run Parcel (v1 and v2) in a safe way; specifically we are needing to run it in its own process so we can kill the whole group as it creates a bunch of child processes that don't shut down on their own
- Neither the parcel@1 nor parcel@2 JS API are nice to use
- We don't need the Parcel server feature, just a watched build

## Approach

Rollup is a better fit for the problems at hand:

- Has a first-class JS API
- Does not require its own sandboxed process group

This PR switches to using Rollup for bundling the manifest.

## To-do

Typechecking and linting of test code is still a problem. I was unable to get `@rollup/plugin-typescript` to behave well, so I fell back to using Babel to compile TypeScript, which does not do any typechecking. I think it may make sense to typecheck and lint as a separate step anyway if a `tsconfig.json` or `.eslintrc.json` are found.


packages/bundler/src/bundler.ts
```diff
@@ -0,0 +1,93 @@
...
+  return bundles.map(bundle => {
+    return {
+      input: bundle.entry,
+      output: {
+        file: bundle.outFile,
+        name: bundle.globalName || undefined,
+        sourcemap: true,
+        format: 'umd',
+      },
+      external: ['perf_hooks', '@babel/runtime'],
```
cowboyd: I'm not familiar with rollup externals. Why is this necessary?

jnicklas (approved):

jnicklas:
@pittst3r :shipit: 🥳 🎈

packages/bundler/src/bundler.ts
```diff
@@ -0,0 +1,93 @@
...
+          bundler.channel.send(message as BundlerMessage);
+        });
+      } finally {
+        rollup.close();
+      }
+    });
+  }
+
+  [SymbolSubscribable]() {
+    return this.messages[SymbolSubscribable]();
```
cowboyd:
```suggestion
    return this.channel[SymbolSubscribable]();
```

`Channel` is itself `Subscribable` so you can delegate to it directly. Does this mean you don't need `messages` anymore?

packages/server/src/manifest-builder.ts
```diff
@@ -63,26 +64,47 @@ function* processManifest(options: ManifestBuilderOptions): Operation {
   return distPath;
 }
 
+function logBuildError(error: BundlerError) {
+  console.error("[manifest builder] build error:", error.message);
+  console.error("[manifest builder] build error frame:\n", error.frame);
+}
+
+function* waitForSuccessfulBuild(bundlerEvents: ChainableSubscription<BundlerMessage, undefined>, delegate: Mailbox): Operation {
+  let message = yield bundlerEvents.first();
```
cowboyd: Since the type of `first()` is `BundlerMessage | undefined`, we probably want to use `bundlerEvents.expect()` here (even though the [not-so-great support for generators in TypeScript][1] means that we have to do this type-checking in our head).

That way, if we did want to annotate explicitly to `BundleMessage`, we'd be sure to be correct.

[1]: https://github.com/microsoft/TypeScript/issues/32523

packages/server/src/manifest-builder.ts
```diff
@@ -63,26 +64,47 @@ function* processManifest(options: ManifestBuilderOptions): Operation {
...
-      outFile: "manifest.js"
-    }
+  let bundler: Bundler = yield Bundler.create(
+    [{
+      entry: options.srcPath,
+      globalName: bigtestGlobals.manifestProperty,
+      outFile: path.join(options.buildDir, "manifest.js")
+    }]
   );
+  let bundlerEvents: ChainableSubscription<BundlerMessage, undefined> = yield subscribe(bundler);
```
cowboyd, line 90: This is an interesting case of whether to use `subscribe` vs `Subscribable.from`. The first subtle difference is that the first maps an actual live subscription, whereas the latter maps an operation that produces a subscription.

From a runtime perspective, `subscribe()` produces a single subscription (tantamount to a single event listener) shared by both `waitForSuccessfulBuild` and `createManifestBuilder` vs two subscriptions, one at each site of iteration you would get from `Subscribable.from`. From a code perspective, `subscribe()` requires an extra `yield` statement and type annotation.

Otherwise, they're identical. I'm not sure I understand the tradeoffs fully, but wanted to at least highlight the differences for this use-case so that we can collect more datapoints. /cc @jnicklas

cowboyd (commented):
This ought to simplify things considerably, way to turn a chore into an architectural upgrade!

packages/bundler/src/bundler.ts
pittst3r: It's not necessary, this is left over from me trying to get everything configured properly. Will remove.

packages/server/src/manifest-builder.ts
pittst3r: One (fixable) problem with using `Subscribable.from()` is the `Chain` interface is not public, making type annotations with it impossible. This is not a great reason, but it was my deciding factor in this case. I do feel that `subscribe()` may be a better fit though as we would otherwise be recursively creating new subscriptions in `waitForSuccessfulBuild()` if I understand correctly.

As an aside, the overlap between `Subscribable` and `Chain` and `ChainableSubscription` is confusing to me. Not sure if there's room to make the differences more clear and understandable or otherwise consolidate into fewer interfaces.

jnicklas: I initially wanted to remove `Subscribable` entirely, but we left it in there, because we didn't know which type of chaining would turn out to be more useful, so we decided on keeping both. That being said, we could certainly improve on the naming of things to make it easier to understand.

cowboyd: @pittst3r seems like the right choice for a recursive method, although this is a very subtle distinction to ask users to 

I wonder if we can make some shared interface that has all the chaining, and then have the low-level thing that just has `next()` on it a `SubscriptionIterator`. We might be able to use that chaining interface regardless of whether we're chaining the subscription or the operation.

cowboyd (approved):

PR #392, opened 2020-07-07
CLI suite improvements

jnicklas:
Improves the structure of the CLI suite and also fixes a race condition.

This once again hammers home that using `once` in a loop is dangerous even when you're pretty sure it's not.

Depends on #393

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -23,7 +23,9 @@
   },
   "devDependencies": {
     "@bigtest/todomvc": "^0.5.2",
+    "@effection/channel": "^0.6.4",
     "@effection/events": "^0.7.4",
+    "@effection/subscription": "^0.9.0",
     "@frontside/tsconfig": "*",
     "@types/capture-console": "1.0.0",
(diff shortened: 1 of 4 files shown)


jnicklas:
No changeset since this only affects the test suite.

cowboyd (commented):
What is the race condition?

jnicklas:
@cowboyd it's a bit unclear, but sometimes the test suite got stuck, leading to timeouts. I've observed this on CI but thought it was just the CI server taking a really long time, but it happens locally as well. I tracked the events and saw that the event was sometimes missed, I think this is probably due to multiple writes to the stream happening in the same tick or something similar to this, we know that `once` doesn't handle this well. I didn't spend any more time tracking it down exactly though.

cowboyd (approved):
`once()` considered harmful :)

jnicklas:
> `once()` considered harmful :)

Well, it's perfectly fine on its own, just using it in a loop is a bit dangerous.

PR #393, opened 2020-07-08
Improve atom

jnicklas:
Makes some breaking changes to atom, to make it more usable.

- Switch to using `Channel` internally
- Return type of subscription is `undefined` instead of `void`
- Add `once` method to `Slice`
- Add `set` to `Atom`
- Remove `Atom#each`, use `forEach` instead.

Depends on https://github.com/thefrontside/effection/pull/173 https://github.com/thefrontside/effection/pull/175

### DIFF
--- a/.changeset/atom-channel.md
+++ b/.changeset/atom-channel.md
@@ -1,0 +1,4 @@
+---
+"@bigtest/atom": "minor"
+---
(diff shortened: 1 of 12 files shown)


cowboyd (approved):
looking good. It does make me wonder if we can share more implementation between atom and slice at some point.

PR #397, opened 2020-07-09
Replace eslint and typescript with frontside/* and remove from root

minkimcello:
## Motivation
By updating BigTest's dependency of `@frontisde/eslint-config` to [`v1.1.0`](https://github.com/thefrontside/javascript/blob/%40frontside/eslint-config-v1.1.0/packages/eslint-config/package.json), I [removed](https://github.com/thefrontside/bigtest/pull/391/files) `eslint` and `typescript` from devDependencies but in hindsight I should not have. 

With the updates of `@frontside/eslint-config@1.1.1` and `@frontside/typescript@1.0.1`, `eslint` is no longer a peer dependency on `eslint-config` and I added `eslint-config` as a peer dependency to `typescript` because `eslint` will be getting installed per package and not at the root and eslint-config can't be nested inside dependencies (issue [here](https://github.com/thefrontside/javascript/issues/10)) respectively.

## Approach
- Bump `frontside/eslint-config` to `v1.1.1`.

## TODOs
- [x] ~Should `frontside/typescript` be installed at root and in each package individually? Or just at root?~ Installed in each individual package.
  - [x] ~`frontside/typescript` comes with `typescript` and `eslint` so can I remove those two dependencies from each package and replace it with the `frontside/typescript` package?~ Did.


minkimcello:
@cowboyd `eslint` and `typescript` were both on the root of the monorepo before it was removed. Should I add it to each individual package instead?

cowboyd:
@minkimcello Yes, this was one of the things that the author of rushjs was talking about on the podcast. Don't entwine your apps and packages with an ambient environment. If you need to depend on eslint and typescript everywhere, then put them in a `@frontside/typescript` package that bundles those two packages.

cowboyd:
Since `@frontside/typescript` is coming, let's maybe wait for that to arrive and then introduce it?

minkimcello:
@cowboyd `typescript` and `eslint` come packaged together inside [`@frontside/typescript`](https://github.com/thefrontside/javascript/blob/v1/packages/typescript/package.json) so should I add dependencies like:
```
/package.json
  - `frontside/typescript`
  - `frontside/eslint-config`
  /packages
    /*
      - `frontside/typescript`
      - `frontside/eslint-config`
```
or just
```
/package.json
  - `frontside/typescript`
  - `frontside/eslint-config`
```
but I think you mentioned that these dependencies should be on each individual package so the former?

cowboyd:
@minkimcello yes, the former, but not at the top level?
```
/package.json
  /packages
    /*
      - `frontside/typescript`
      - `frontside/eslint-config`
```

And if we can figure out how to load eslint-config from `frontside/typescript`, that would be amazing.

minkimcello:
> @minkimcello yes, the former, but not at the top level?

Done. Replaced `typescript` and `eslint` with `frontside/typescript` and added `frontside/eslint-config` to every applicable package. And removed `frontside/typescript` and `frontside/eslint-config` at root.

And I kept the `.eslintrc.json` at root.

> And if we can figure out how to load eslint-config from frontside/typescript, that would be amazing.

I'll occasionally check eslint's repository to see if they ever implement that feature.

cowboyd (approved):
I think we're good here, can we rebase this puppy so that we can move it through?

PR #398, opened 2020-07-10
Pipe app log to file

jnicklas:
This is a pretty low-tech solution which gives us at least *some* ability to debug failures in the app server by looking at the log file. I won't have time to build a more comprehensive solution before my vacation, so I will leave this here and focus on building out capturing console logs, which I think is probably more valuable.

### DIFF
--- a/packages/server/src/app-server.ts
+++ b/packages/server/src/app-server.ts
@@ -3,6 +3,7 @@ import { Mailbox } from '@bigtest/effection';
 import { ChildProcess } from '@effection/node';
 import { Socket } from 'net';
 import * as process from 'process';
+import { createWriteStream } from 'fs';
 
 interface AppServerOptions {
   delegate: Mailbox;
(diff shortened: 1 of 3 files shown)


packages/server/src/app-server.ts
```diff
@@ -36,12 +38,16 @@ function isReachable(port: number, options: { timeout: number } = { timeout: 100
...
-  yield ChildProcess.spawn(options.command, options.args || [], {
+  let child = yield ChildProcess.spawn(options.command, options.args || [], {
     cwd: options.dir,
     detached: true,
     env: Object.assign({}, process.env, options.env),
   });
 
+  let logStream = createWriteStream(options.logPath);
+  child.stdout.pipe(logStream);
+  child.stderr.pipe(logStream);
```
jnicklas, line 49: These streams should terminate automatically when the process exits, which should close the pipes.

cowboyd (approved):

PR #400, opened 2020-07-10
Add bigtest package

minkimcello:
## Motivation
It would be real cool if users can install bigtest by running `npm install bigtest` as opposed to `npm install @bigtest/cli` so we're going to rename `@bigtest/cli` to just `bigtest`.

## Approach
- Created `bigtest` package with the `cli` package as dependency.


minkimcello:
500 error on the preview. My hands are clean.

cowboyd (changes requested):
Our strategy should be to use the `bigtest` package as a shim that either uses the _local_ copy of `@bigtest/cli` That way, you can use bigtest as a global executable, but what happens under the covers is a razor-thin shim.

So, for example, if I have projects one and two of which depends on `@bigtest/cli: 1.3.0` another `@bigest/cli: 2.0.0`

```
~/one$ bigtest --version
bigtest 1.3.0
~/one$ cd ../two
~/two$ bigtest --version
bigtest 2.0.0
```

In summary `@bigtest/cli` stays as-is. `bigtest` is a shim that loads `@bigtest/cli`.

cowboyd:
> * Should we continue the package version from where we left off on `cli` or reset it?

I think we need to make sure it's a major bump over the currently published version.... If the public version (before it was associated with this project) was just a placeholder, then we can make it `0.1.0`

minkimcello:
@cowboyd Oh oops that was a question before I knew we were going to keep `packages/cli`. Now the question is how should I set up `packages/bigtest` to extend `packages/cli`? Is it just a matter of specifying the input and ouput of `yarn prepack` for `packages/bigtest`?

And do we need to find an automated workflow solution for bumping `packages/bigtest` when `packages/cli` gets bumped?

pittst3r:
@minkimcello We should just make `packages/cli` a `dependency` of `packages/bigtest`.

pittst3r:
I feel like we should start the package at `0.0.1` (the last version on npm) then use changesets to bump it to `0.1.0`. What do y'all think about that strategy?

minkimcello:
@pittst3r Good idea. Forced me to not forget to add the package to the list of packages on the monorepo config. Because `packages/bigtest` has `packages/cli` as a dependency, that means when `cli` gets bumped changesets will bump `bigtest` as well... in theory. Added `echo no-op` for all three scripts for the `bigtest` package.

Should we have the preview workflow ignore bigtest package?

pittst3r:
> Should we have the preview workflow ignore bigtest package?

Will the preview workflow release a preview of `bigtest` when any of the dependencies get bumped? If so, I would say do not ignore.

minkimcello:
@pittst3r Mmm well I was originally thinking that `cli` and `bigtest` would be identical but now that I think about it, the preview workflow wouldn't pick up on `bigtest` anyway because we wouldn't be doing any direct work on it as it just mirrors `cli` (at the moment).

And if dependencies do get bumped, it would only reflect it on the changeset's generated PR which the preview workflow skips.

So in conclusion I don't think we need to worry about it. 👌

cowboyd (approved):

cowboyd:
@minkimcello Do we need to set it to publish? https://www.npmjs.com/package/bigtest

<blockquote><div> npm</div><div><strong><a href="https://www.npmjs.com/package/bigtest">bigtest</a></strong></div><div>Create unit tests from run time profiling and data flow analysis</div></blockquote>

minkimcello:
> @minkimcello Do we need to set it to publish? https://www.npmjs.com/package/bigtest

I believe because bigtest has a dependency of cli, changesets will bump bigtest if cli is bumped.

<blockquote><div> npm</div><div><strong><a href="https://www.npmjs.com/package/bigtest">bigtest</a></strong></div><div>Create unit tests from run time profiling and data flow analysis</div></blockquote>

PR #401, opened 2020-07-12
Manifest builder improvements

pittst3r:
## Motivation

I noticed a few things to clean up:

- Rollup wasn't properly excluding node_modules from watch
- The `Bundler` `try/finally` was not wrapping the code that needed it
- At the default log level, users see manifest build errors, but nothing when the build becomes successful
- There is a bug in `fs.promises.truncate()` that causes warnings https://github.com/nodejs/node/issues/34189

## Approach

- Use a regex to exclude node_modules from rollup watch
- Use `ensure()` API instead of `try/finally` in `Bundler`
- Upgrade the log level of the successful manifest build notification
- Use `ftruncate()` instead of `truncate()` until the [fix](https://github.com/nodejs/node/pull/34239#issuecomment-656168437) is released

### DIFF
--- a/packages/bundler/src/bundler.ts
+++ b/packages/bundler/src/bundler.ts
@@ -70,24 +70,21 @@ export class Bundler implements Subscribable<BundlerMessage, undefined> {
     let bundler = new Bundler();
 
     return yield resource(bundler, function*() {
-      let rollup: RollupWatcher | null = null;
+      let rollup: RollupWatcher = watch(prepareRollupOptions(bundles));;
 
-      yield ensure(() => {
-        if (rollup) {
-          rollup.close();
-        }
-      });
+      try {
+        let events: ChainableSubscription<Array<RollupWatcherEvent>, undefined> = yield subscribe(on<Array<RollupWatcherEvent>>(rollup, 'event'));
+        let messages = events
+          .map(([event]) => event)
+          .filter(event => event.code === 'END' || event.code === 'ERROR')
+          .map(event => event.code === 'ERROR' ? { type: 'error', error: event.error } : { type: 'update' });
 
-      rollup = watch(prepareRollupOptions(bundles));
-      let events: ChainableSubscription<Array<RollupWatcherEvent>, undefined> = yield subscribe(on<Array<RollupWatcherEvent>>(rollup, 'event'));
-      let messages = events
-        .map(([event]) => event)
-        .filter(event => event.code === 'END' || event.code === 'ERROR')
-        .map(event => event.code === 'ERROR' ? { type: 'error', error: event.error } : { type: 'update' });
-
-      yield messages.forEach(function*(message) {
-        bundler.channel.send(message as BundlerMessage);
-      });
+        yield messages.forEach(function*(message) {
+          bundler.channel.send(message as BundlerMessage);
+        });
+      } finally {
+        rollup.close();
+      }
     });
   }
 
--- a/packages/bundler/src/bundler.ts
+++ b/packages/bundler/src/bundler.ts
@@ -2,7 +2,6 @@ import { Operation, resource } from 'effection';
 import { on } from '@effection/events';
 import { subscribe, Subscribable, SymbolSubscribable, ChainableSubscription } from '@effection/subscription';
 import { Channel } from '@effection/channel';
-import { ensure } from '@bigtest/effection';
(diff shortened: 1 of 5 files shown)


packages/server/src/manifest-builder.ts
```diff
@@ -24,14 +24,33 @@ interface ManifestBuilderOptions {
...
+  return ({ fail, resume }) => {
+    fs.ftruncate(fd, len, err => {
+      if (err) {
+        fail(err);
+        return;
+      }
+      resume();
+    })
+  };
+}
```
cowboyd: We want to avoid using the "control function" version of operations since they will almost certainly be deprecated in the next version of effection. Instead, we can use a `Deferred` object to convert any callback-based api into a promise which can then be used as a normal operation.
```ts
import { Deferred } from '@bigtest/effection';

function*  ftruncate(fd: number, len: number): Operation<void> {
  let { resolve, reject, promise } = Deferred<void>();
  fs.ftruncate(fd, len, err => {
    if (err) {
      reject(err);
    } else {
      resolve();
    }
  });
  yield promise;
}
```

packages/server/src/manifest-builder.ts
```diff
@@ -24,14 +24,33 @@ interface ManifestBuilderOptions {
...
+
+// https://github.com/nodejs/node/issues/34189#issuecomment-654878715
+function* truncate(path: string, len: number): Operation {
+  let file: fs.promises.FileHandle = yield open(path, 'r+');
+  yield ensure(() => file.close());
```
cowboyd: We're still on the fence about when the right time is to use `ensure()`, since in the cases where it's necessary, it's almost always modeled better using a resource. In this case, a simple try/finally is "just javascript" and communicates the intent clearly to all javascript programmers.

```ts
let file: fs.promises.FileHandle = yield open(path, 'r+');
try {
  yield ftruncate(file.fd, len);
} finally {
  file.close();
}

File: packages/bundler/src/bundler.ts [hunk 3]
```diff
@@ -64,21 +68,26 @@ export class Bundler implements Subscribable<BundlerMessage, undefined> {
...
+      let rollup: RollupWatcher | null = null;
+
+      yield ensure(() => {
+        if (rollup) {
+          rollup.close();
+        }
+      });
+
+      rollup = watch(prepareRollupOptions(bundles));
+      let events: ChainableSubscription<Array<RollupWatcherEvent>, undefined> = yield subscribe(on<Array<RollupWatcherEvent>>(rollup, 'event'));
```
cowboyd: I'm still not quite sure why a try/catch won't catch the right thing. If you put it around 82-90, it will catch any exception raised therein.

cowboyd (commented):
Nice set of stability fixes! 🍷

packages/bundler/src/bundler.ts
pittst3r: This is what it was originally:

```ts
let rollup: RollupWatcher = watch(prepareRollupOptions(bundles));;
let events: ChainableSubscription<Array<RollupWatcherEvent>, undefined> = yield subscribe(on<Array<RollupWatcherEvent>>(rollup, 'event'));
let messages = events
  .map(([event]) => event)
  .filter(event => event.code === 'END' || event.code === 'ERROR')
  .map(event => event.code === 'ERROR' ? { type: 'error', error: event.error } : { type: 'update' });

try {
  yield messages.forEach(function*(message) {
    bundler.channel.send(message as BundlerMessage);
  });
} finally {
  rollup.close();
}
```

Which is wrong I think, prompting me to make a change. However, rather than just wrap more in the `try` block I thought `ensure()` was the way to go, but now that you've explained that that's not the way to go I'll update to this:

```ts
let rollup: RollupWatcher = watch(prepareRollupOptions(bundles));;

try {
  let events: ChainableSubscription<Array<RollupWatcherEvent>, undefined> = yield subscribe(on<Array<RollupWatcherEvent>>(rollup, 'event'));
  let messages = events
    .map(([event]) => event)
    .filter(event => event.code === 'END' || event.code === 'ERROR')
    .map(event => event.code === 'ERROR' ? { type: 'error', error: event.error } : { type: 'update' });

  yield messages.forEach(function*(message) {
    bundler.channel.send(message as BundlerMessage);
  });
} finally {
  rollup.close();
}
```

cowboyd (approved):

PR #403, opened 2020-07-13
Track console messages and uncaught errors

jnicklas:
When an error occurs in the agent, track the stack of console logs that have happened in the harness until that time and send them along with the error.

In the future we should also forward all logs to the orchestrator immediately.

Depends on #497

### DIFF
--- a/packages/agent/app/harness.ts
+++ b/packages/agent/app/harness.ts
@@ -1,13 +1,10 @@
 import 'regenerator-runtime/runtime';
 import { wrapConsole } from './wrap-console';
-import { HarnessMessage } from './harness-protocol';
 import { main, spawn } from 'effection';
 import { on } from '@effection/events';
 import { serializeError } from './serialize-error';
+import { getLogConfigFromAppFrame } from './log-config';
 
-function postToParent(message: HarnessMessage) {
-  window.parent.postMessage(JSON.stringify(message), "*");
-}
 
 // proxy fetch and XMLHttpRequest requests through the parent frame
 if(window.parent !== window) {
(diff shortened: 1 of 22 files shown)


jnicklas:
@pittst3r @cowboyd this is my work in progress on tracking and forwarding the console from the harness.

jnicklas:
@cowboyd this is now good to go. Since it's quite a substantial change, I didn't want to leave this unfinished. I did not actually add any changes which would show this information in the formatter so as not to conflict with #462, however, I've verified that the information does indeed percolate all the way to the CLI.

I'm pretty excited about this honestly. The ability to see exceptions from the app under test is pretty huge. Getting an exception from the failed test is interesting, but it won't tell us anything about *why* it failed. This will make debugging failures *so* much easier.

packages/agent/app/harness.ts
```diff
@@ -1,3 +1,14 @@
+import 'regenerator-runtime/runtime';
+import { wrapConsole } from './wrap-console';
+import { HarnessMessage } from './harness-protocol';
+import { main, spawn } from 'effection';
+import { on } from '@effection/events';
+import { serializeError } from './serialize-error';
```
jnicklas: One downside of this PR is that it adds a *ton* of stuff to the harness, making it *far* heavier than before. Maybe this is unavoidable in the long run, but worth considering.

packages/agent/app/harness.ts
```diff
@@ -1,3 +1,14 @@
+import 'regenerator-runtime/runtime';
```
jnicklas, line 1: Man, having to add this thing here is a real bummer.

cowboyd: Yeah, this really sucks. I guess the upshot is that we'll have an expanded base of browser support 🤷

cowboyd (commented):
This will indeed make debugging hugely helpful. In fact, I think it would be nice eventually to be able to subscribe to the agent logs as well. My only real concern here is using the `window.parent.postMessage` to communicate between the harness and test runner and what kind of guarantees we can rely on for the timing of message delivery.

How can we know that all of the console messages have been processed and appended to the console array by the time that the lane has finished running? Based on my recent experience of trying to use `postMessage` to communicate all of the test events, it can be highly latent if the browser decide that one of the frames is not a high priority.

jnicklas:
@cowboyd I share your concern. I was a bit torn on whether to use `postMessage` or some direct communication method. My thinking was that `postMessage` might be less of an impediment to performance, and that if log messages are lost, it doesn't matter too much. But I'm not sure if that's the right choice.

cowboyd (changes requested):
Let's explore merging the uncaught errors and consoleMessages into a single buffer. Also, consider using a synchronous mechanism for capturing output.

cowboyd (approved):

PR #418, opened 2020-07-24
increase time-to-connect for webdriver

cowboyd:
Motivation
-----------

Prior to the release of Chrome 84, one second was plently of time to connect to the webdriver server. However, with that release, it started timing out regularly.

Approach
----------

The long term solution is to not have timeout at all inside invididual components, but to set a deadline for the entire startup process. So that the entire orchestrator will have a certain amount of time to startup and services can take as long as they want within that timeframe. However, that requires a lot of changes in the orchestrator state and startup process, so this just hops it up to five seconds.

### DIFF
--- a/.changeset/bump-webdriver-timeout.md
+++ b/.changeset/bump-webdriver-timeout.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/webdriver": patch
+---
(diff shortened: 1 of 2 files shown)


taras (approved):
👍

PR #424, opened 2020-07-29
inject `process.env.NODE_ENV` into environment

cowboyd:
Motivation
-----------

Many third party libraries use `process.env.NODE_ENV` even when running on the browser. While this isn't strictly our responsibility to account for this, there's little we can do about it. In the case that fomented this change, we had a package (@mirage/graphql) that depended on a package (graphql) that referenced `process.env.NODE_ENV`. Parcel and webpack (through a plugin) both make this variable available, and so we should too.

Approach
----------
In order to define the global `process.env` variable, we're using what appears to be the most well-worn rollup plugin to do so. However, because of the way its exports are constructed, it did require us to tweak our `tsconfig` to use `esModuleInterop`.

I hardcoded the `NODE_ENV` to 'production' since there's no reason to slow down elements of the test bundle itself. We may want to consider allowing to set this to some other value at some point, but not too sure what form that would take.

Finally, I added a preference for ES modules as an entry point since they are usually better behaved with TypeScript.

### DIFF
--- a/packages/bundler/tsconfig.json
+++ b/packages/bundler/tsconfig.json
@@ -1,4 +1,7 @@
 {
   "extends": "@frontside/tsconfig",
-  "files": ["src/index.ts"]
+  "files": ["src/index.ts"],
+  "compilerOptions": {
+    "esModuleInterop": true
+  }
 }
--- a/.changeset/bundle-with-process-env.md
+++ b/.changeset/bundle-with-process-env.md
@@ -1,0 +1,7 @@
+---
+"@bigtest/bundler": minor
+"@bigtest/cli": patch
+---
+set process.env.NODE_ENV='production' inside the test bundle. This
+will ensure that 3rd party packages that depend on this will continue
+to function
--- a/packages/bundler/package.json
+++ b/packages/bundler/package.json
@@ -31,7 +31,8 @@
     "@rollup/plugin-node-resolve": "^8.1.0",
     "effection": "^0.7.0",
     "express": "^4.17.1",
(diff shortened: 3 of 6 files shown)


packages/bundler/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "@frontside/tsconfig",
-  "files": ["src/index.ts"]
+  "files": ["src/index.ts"],
+  "compilerOptions": {
+    "esModuleInterop": true
+  }
```
pittst3r, line 6: I'm don't think this tsconfig.json will get picked up when running tests in a downstream project using BigTest. This likely works because this is the tsconfig.json that gets used when running the Bundler tests. We may need to specify this option directly in the `@babel/plugin-transform-typescript` config.

cowboyd: This is actually so that the import of `injectProcessEnv` works from the process.env rollup plugin and can be used in our pipeline, not for the typescript in the pipeline itself if that's what you mean. basically, I couldn't get the `bundler.ts` to compile without this since `import * as plugin from 'rollup-plugin-inject-process-env'`  wouldn't work.

cowboyd: In summary, I need to change the tsconfig.json so that `bundler.ts` compiles, not so that the test bundle compiles.

pittst3r: Ah, that's right. Okay, this should be fine then.

pittst3r (approved):

PR #425, opened 2020-07-29
Start trying to switch agentId to agent

RobbieTheWagner:


packages/server/src/schema.ts
```diff
@@ -88,7 +88,18 @@ export const schema = makeSchema({
         t.string('type');
         t.string('status', { nullable: true });
         t.id('testRunId');
-        t.id('agentId', { nullable: true });
+        t.list.field("agents", {
+          type: "Agent",
+          resolve: (state) => Object.values(state.agents)
+        });
```
cowboyd, line 94: Bravo! Rather than have a list of agents _and_ a single agent on the same structure, one of which will be defined, It's likely to me that we'll have to splint out the `TestEvent` type into a [union type](https://graphql.org/learn/schema/#union-types) of `TestRunResult | TestResult | TestEvent` similar to [the way it is in TypeScript](https://github.com/thefrontside/bigtest/blob/v0/packages/server/src/schema/test-event.ts#L19-L22). It requires a [little bit of extra code](https://nexus.js.org/docs/api-uniontype) to specify it using Nexus, but I think it'll make things more coherent.

In fact, maybe we should have the test event just be a union of existing types `TestRun | TestRunAgent | TestResult | StepResult | AssertionResult`

That way, all of the information is already there and depending on which object is sent, we can say:

* `TestRun` - sent when the test run status changes (this should always be the last event sent and the status should be either `ok` or `failed`)
* `TestRunAgent` - status of a test run on a single agent has changed
* `TestResult` - one test on one agent is changing status
* `StepResult` - one step on one agent is changing status
* `AssertionResult` - one assertion on one agent is changing status

We didn't do it this way to begin with because it was all our first time with GraphQL either 😆

cowboyd:
@rwwagner90 Can we close this since I think it's superceded by #462 ?

PR #441, opened 2020-08-05
Run tests on Windows

taras:


cowboyd:
@taras any chance we can rebase with some of @jbolda's changes? Also, we should turn `fail-fast` off, so that we get the complete test output even where we have failures.

taras:
@cowboyd I think I should just close this PR and let @jbolda pick it up from scratch.

jbolda:
That was my expectation since we were thinking to split up the OSes into separate files.

PR #442, opened 2020-08-05
No longer using xvfb

taras:
## Motivation

I'm attempting to get tests running in GitHub Actions on Windows but it's failing while trying to use `apt` on windows to install xvfb. xvfb was used when interactors were running tests Chrome/Karma. I'm not sure if this is still necessary.

## Approach

Try removing xvfb


cowboyd (approved):
Makes sense. All our actions that actually run bigtest tests do not need it. Nice one !

PR #449, opened 2020-08-06
Load app from an arbitrary url

### LINKED ISSUE #416: Load app from any url, not just a locally managed dev server to enable e2e testing.
One of the most exciting things about the new bigtest architecture is that the app is loaded as a complete black-box. BigTest doesn't assume anything about the build system, instead just loading it from a url. This means that we don't have to worry about any setup or teardown since it happens at the application level, but it also has the really awesome property of not requiring that the application being loaded even exist on the local machine at all.

However, as-is, the orchestrator assumes that _it_ will be managing the application server process by starting it up before the tests, and tearing it down again afterwards. It also assumes that the application source will be served from the local development environment by only requiring port number. Thus, the app configuration section of our file looks something like:

```json
{
  "app": {
    "command": "BROWSER=none PORT=36000 yarn start",
    "port": 36000
  }
}
```

However, we'd like to be able to optionally specify that the app is running at a completely different place.

What if, instead of having a `port` and `command` options, we required the full url every time and _optionally_ accepted a `command` option to start the server:

```json
{
  "app": {
    "command": "BROWSER=none PORT=36000 yarn start",
    "url": "http://localhost:36000"
  }
}
```

If we don't receive a command option, then we assume that the app is external and we don't manage it. So

```json
{
  "app": {
    "url": "http://myapp.mydomain.com"
  }
}
```

Will just probe that URL to make sure it's available, but then happily just load all content from there for the duration of the test runs.

It should also be available as a command line option and not just a config file option:

```
$ bigtest server --app-url http://myapp.mydomain.com
```

which will run the test suite against that app instead of whatever is in the local config file.

jgnieuwhof:
Implements #416.

One of the coolest things about bigtest's architecture is its ability to test applications regardless of where they are run, or how they were started. BigTest currently undercuts its own ability to do so by making an assumption that it is both starting the application and testing it on a local port.

We can remove this assumption by making the bigtest `"command"` optional, and by specifying a full target `"url"`, rather than just a `"port"`.


packages/cli/bigtest.json
```diff
@@ -1,7 +1,7 @@
 {
   "app": {
-    "command": "bigtest-todomvc 36000",
-    "port": 36000
+    "command": "node ../todomvc/dist/start.js 36000",
```
jgnieuwhof: `bigtest-todomvc` was not available on my path, causing the cli tests to fail.

Is there some additional environment set-up I'd need to do to allow this command to work?

packages/server/package.json
```diff
@@ -74,5 +74,9 @@
     "trumpet": "^1.7.2",
     "websocket": "^1.0.30",
     "yargs": "^15.0.2"
+  },
+  "volta": {
+    "node": "12.16.0",
+    "yarn": "1.19.1"
```
jgnieuwhof, line 82: The other packages have `volta` configurations, allowing you to work with the individual package as if it weren't in a monorepo. I've added this here to follow suit.

Related volta ticket: https://github.com/volta-cli/volta/issues/378

packages/server/src/app-server.ts
```diff
@@ -1,39 +1,83 @@
...
-    detached: true,
-    env: Object.assign({}, process.env, options.env),
-  });
+export function* createAppServer({ slice, ...options }: AppServerOptions): Operation {
+  let appOptions = slice.slice<AppOptions>(['appOptions']);
+  let appStatus = slice.slice<AppStatus>(['appStatus']);
+  let current: Context;
+
+  function* startApp(appOptions: AppOptions): Operation<void> {
+    if (!appOptions) {
```
cowboyd: How is possible?

packages/server/src/app-server.ts
```diff
@@ -1,39 +1,83 @@
...
+    current = yield fork(function* () {
+      if (appOptions.command) {
+        let child = yield ChildProcess.spawn(appOptions.command as string, [], {
+          cwd: appOptions.dir,
+          detached: true,
+          env: Object.assign({}, process.env, appOptions.env),
+          stdio: 'inherit'
+        });
+
+        child.on('exit', (code: number, signal: string) => {
```
cowboyd: We don't want to use a direct event listener here, because it will never get uninstalled. Probably not much of a difference in this case since the listere and the child process have the same scope of life. In general though, for all `EventEmitter` and `EventTarget` objects, you can use the `on` method from `@effection/events`. e.g.

```ts
let [code, signal]: [number, string] = yield on(child, 'exit');
console.log("🧨", code, signal);
appStatus.set('crashed');
```

Not only do you have your lexical context freed from a closure, but you never have to worry about having to call `off()` because effection will do it for you.

packages/server/src/orchestrator.ts
```diff
@@ -119,7 +118,9 @@ export function* createOrchestrator(options: OrchestratorOptions): Operation {
       console.debug('[orchestrator] connection server ready');
     });
     yield fork(function*() {
-      yield appServerDelegate.receive({ status: 'ready' });
+      yield options.atom.once((state) => {
+        return state.appService.appStatus === 'started'
```
cowboyd: We don't want to continue here until the app is reachable. I think that we probably ought to have a separate status for the app and the app service no? The app service is pretty much running the whole time, even through multiple app restarts

jgnieuwhof: To confirm, we don't want the orchestrator to be ready until the app is reachable? This introduces different behaviour on start than once the orchestrator's already running:
- on start an unreachable URL will cause the orchestrator to never reach a ready state
- once the orchestrator is already running / ready it's possible to update the app url to an unreachable one, while the orchestrator continues be in a running / ready state

On separating state - who would be consuming the app service state? I guess I'm having a hard time formulating the app service states without knowing its purpose.

packages/server/test/orchestrator.test.ts
```diff
@@ -75,46 +43,165 @@ describe('orchestrator', () => {
...
-    it('responds successfully', () => {
-      expect(response.ok).toEqual(true);
+  describe('an externally managed application', () => {
+    beforeEach(async function() {
+      await actions.fork(function * () {
+        let child = yield ChildProcess.spawn("yarn test:app:start 24100");
+
+        actions.registerAfterDestroyPromise(
+          new Promise(resolve => {
+            child.on('exit', () => setTimeout(resolve, 100));
```
jgnieuwhof: @cowboyd - waiting for the child to exit wasn't enough to solve the port collision issue, based on experimentation I had to wait longer for the port to be available.

packages/server/test/helpers.ts
```diff
@@ -91,18 +96,25 @@ after(async function() {
...
+afterEach(async () => {
   if(globalWorld.execution.state === 'errored') {
     throw globalWorld.execution.result;
   }
   if(currentWorld.execution.state === 'errored') {
     throw currentWorld.execution.result;
   }
   currentWorld.destroy();
+
+  await Promise.all(afterDestroyPromises);
```
jgnieuwhof: FYI - `afterEach` hooks are run inside-out; we can't wait for the child process to exit in the test's `afterEach` hook because we'd be waiting _before_ the current world is destroyed.

https://github.com/thefrontside/bigtest/pull/449/files#diff-b16dc92ebd232f17c34dba7e2a2f4f97R79

packages/server/src/orchestrator.ts
cowboyd: Hmm..... these are really interesting questions. In my mind, the idea of a service being ready is simply that it is available to accept requests. In the case of the app service, it is available to service requests to set the application options, so this actually seems right the way you have it. The problem is that the `bigtest ci` command, and the test cases assume that when the orchestrator sends the ready event, that it is ready to run tests.

Of course, because we're using stateful mailboxes, the request to run will wait until the orchestrator does become ready so this should not technically be a problem, but does pose a conundrum around what exactly is "ready"

On the subject of services themselves, I'm looking ahead towards them to have a uniform shape across all of the services (app server, browser manager, manifest-generator, etc...). That way, we can build UI to inspect services in a generic way, so that if some service has an error it can be surfaced without anything specific to the service itself. Basically, I'm imagining what it would look like to build an inspector for the entire orchestrator state.

packages/server/test/helpers.ts
cowboyd: In this case, maybe it would be best to just get a fresh port for each test so that we don't have to worry about collisions using something like [get-port](https://github.com/sindresorhus/get-port)?

Side-steps the issue without any special teardown since we know the port _is_ being released eventually.

packages/server/test/orchestrator.test.ts
cowboyd: as mentioned elsewhere, what about just starting on a fresh port each time?

packages/cli/bigtest.json
jnicklas: Did you run `yarn prepack` in the root of the repository?

taras:
@cowboyd this PR is very close to being finished. There is something funky with server tests. Can we look at this together tomorrow?

cowboyd:
> @cowboyd this PR is very close to being finished. There is something funky with server tests. Can we look at this together tomorrow?

@taras let's go for it 🤜 🤛

jgnieuwhof:
@cowboyd - this may actually be good to go now. I consolidated the app options to have one source of truth (the atom), and cleaned a few things up.

jnicklas (approved):
Looks great to me!

jnicklas:
This seems to be good to go! Any objections to merging this? cc @cowboyd

cowboyd:
> This seems to be good to go! Any objections to merging this? cc @cowboyd

@jnicklas nope! no qualms here 👍

cowboyd (approved):

taras (changes requested):
Please add a changeset. Run `yarn changeset` in the root and select level of impact for each module you changed.

cowboyd:
@jgnieuwhof heads up: In order to get past the new MacOS test run introduced with #486, you'll have to rebase onto origin/v0 and then close and re-open this PR to pull in the new checks.

jgnieuwhof:
Closing and re-opening to run the new checks.

taras (approved):

Issue #279, opened 2020-05-13
Better error messaging if manifest fails to load.


cowboyd:
If the manifest bundle containing all of the test definitions is bad in any way, this crashes the server when it tries to load the manifest.

1. How do we report the fact that the manifest is bad to the server dashboard?
2. How do we report the fact that the manifest was bad when running a test?
3. How do we report it in our CI runner?

cowboyd:
I think we want to maintain the bundle in the `OrchestratorState`, this will let us make better UI on top of what the bundle is currently doing, e.g. we could have the bundle be something like "unknown" | "building" | "errored" | "green". That way, if you request a test run and it is not "green" we can issue a warning: "Tests are currently not building... running tests against last built version"

cowboyd:
The following test suite can reproduce this error:

```ts
export default test('Basic Test').child()
```

Note that this is invalid typescript, so it actually should never even try to load it. See #280 #443

dagda1:
a `RollupError` looks like this

```javascript
{
  pos: 32,
  loc: {
    file: '/Users/developer15/projects/bigtest/packages/server/test/fixtures/error.t.js',
    line: 2,
    column: 15
  },
  raisedAt: 43,
  frame: '1: export default {\n2:   description: "Signing In\n                  ^\n3: }',
  id: '/Users/developer15/projects/bigtest/packages/server/test/fixtures/error.t.js',
  hook: 'transform',
  code: 'PLUGIN_ERROR',
  plugin: 'commonjs',
  watchFiles: [
    '/Users/developer15/projects/bigtest/packages/server/tmp/test/orchestrator/manifest/src/manifest.js',
    '\u0000rollup-plugin-inject-process-env',
    '/Users/developer15/projects/bigtest/packages/server/test/fixtures/empty.t.js',
    '/Users/developer15/projects/bigtest/packages/server/test/fixtures/raw-tree-format.t.js',
    '\u0000/Users/developer15/projects/bigtest/packages/server/test/fixtures/raw-tree-format.t.js?commonjs-proxy',
    '\u0000/Users/developer15/projects/bigtest/packages/server/test/fixtures/empty.t.js?commonjs-proxy',
    '\u0000/Users/developer15/projects/bigtest/packages/server/test/fixtures/error.t.js?commonjs-proxy',
    '/Users/developer15/projects/bigtest/packages/server/test/fixtures/error.t.js'
  ]
}
```
And of course, you have a stack trace.

[...]

PR #450, opened 2020-08-07
Manifest build failure reporting

dagda1:
This is a first rough pass at #279 which creates a discriminated union to map the state and various transitions of the bundler

```typescript
import { RollupWarning, RollupError } from 'rollup';

export type BundlerError = RollupError;
export type BundlerWarning = RollupWarning;

export type BundlerState = 
  | { status: 'unbundled' } 
  | { status: 'building'; warnings: BundlerWarning[] } 
  | { status: 'green'; path: string;  warnings: BundlerWarning[] }
  | { status: 'end'; warnings: BundlerWarning[] }
  | { status: 'errored'; error: BundlerError }
  | { status: 'warn'; warnings: BundlerWarning[] }
  | { status: 'updated'; path: string };
```

The global `atom` state now has a bundler `slice`.

```typescript
export type OrchestratorState = {
  agents: Record<string, AgentState>;
  manifest: Manifest;
  bundler: BundlerState;
  testRuns: Record<string, TestRunState>;
}
```

The `bundler` starts life in an `unbundled` state before transitioning to `building` in the manifest-builder and hopefully further transitions to the happy path `green`.

When the `bundler` has transitioned from `unbundled`, the bundler can at any time transition to `errored` and an error object gets pushed onto the slice.  

I've used the `atom` to broadcast state transitions from the bundler which can be consumed anywhere that has a handle of the `atom`.

e.g. in the orchestrator:

```typescript
    yield fork(function* () {
      yield options.atom.slice<BundlerState>(['bundler']).once(({ status }) => status === 'green');
      console.debug('[orchestrator] manifest builder ready');
    });
```


packages/server/src/manifest-generator.ts
```diff
@@ -16,7 +15,7 @@ interface ManifestGeneratorOptions {
 };
 
 function* writeManifest(options: ManifestGeneratorOptions) {
-  let files = yield Promise.all(options.files.map((pattern) => promisify(glob)(pattern))).then((l) => l.flat());
+  let files = yield globby(options.files);
```
cowboyd: This seems like an awesome improvment. Can we maybe add `globby` as a separate PR?

packages/server/src/orchestrator.ts
```diff
@@ -46,6 +46,10 @@ export function* createOrchestrator(options: OrchestratorOptions): Operation {
 
   let connectTo = `ws://localhost:${options.project.connection.port}`;
 
+  yield spawn(function* () {
+    yield createLogger({ atom: options.atom,  out: console.error });
+  });
+
```
cowboyd: Since `createLogger({ atom: options.atom,  out: console.error })` is an operation, you can spawn it directly. No need to wrap it in an operation that just yields to it:

```suggestion
  yield spawn(createLogger({ atom: options.atom,  out: console.error }));
```

packages/server/test/logger.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    let atom: Atom<OrchestratorState>;
+    let logger:  (<A extends unknown[]>(...a: A) => void) = (...args) => {
+      expect(args).toEqual(["[manifest builder] build error:", "blah"]);
+    };
+
+    atom = createOrchestratorAtom();
+    
+    actions.fork(function* () {
+      yield createLogger({ atom, out: logger });
+    })
```
cowboyd: again here you can fork the logger directly, no need to wrap it:

```suggestion
    actions.fork(createLogger({ atom, out: logger }));
```

cowboyd (commented):
It's really exciting to see the bundling operations starting to get reflected in the atom! It means that the kind of UI that we're going to be able to build on top of this is going to be serious next-level.

There are two further things to consider:

1. It makes me wonder though if we should not have the coupling to the state penerate all the way into the bundling code and instead let that fall under the purview of the manifest builder. In other words, the bundler is a subscribable resource that accepts commands and streams messages, while the manifest builder is responsible for applying that stream to the atom.

2. The second is that I think that the bundler state should be part of the Manifest state so that it really is a subsystem contained within the manifest builder. That way, there is one place you can look, the manifest state, to see everything you need to know about the manifest.... what it looks like currently, if it's bundling, if it's errored, what errors there are.

dagda1:
@cowboyd Very much agree with points 1 and 2.  I actually had the bundler as part of the manifest and then had a moment of doubt :). 

If the bundler goes under the manifest, would the graphql schema need updated?

Are these actionable items that I should do for this PR or just things to consider?

cowboyd:
> Are these actionable items that I should do for this PR or just things to consider?

I think we should the low hanging fruit as well as (1) in this PR. (2) will be necessary, but would it make sense to do it as a follow-on since it would require breaking changes to the GraphQL schema?

dagda1:
@cowboyd I'll do 1 here and 2 on a separate PR.

dagda1:
@cowboyd I've added 1 in this PR

.changeset/manifest-error-reporting.md
```diff
@@ -0,0 +1,7 @@
+---
+"@bigtest/bundler": minor
+"@bigtest/project": mnor
```
pittst3r:
```suggestion
"@bigtest/project": minor
```

package.json
```diff
@@ -46,6 +46,7 @@
     "@frontside/eslint-config": "^1.1.0"
   },
   "resolutions": {
+    "globby": "11.0.1",
```
pittst3r: Just want to verify this is necessary.

pittst3r (commented):
Looks great! Just had a mnor edit and question.

package.json
dagda1: @pittst3r you are very right to ask.  I added it because it is getting clobbered.  Here is what `yarn list globby` returns without the resolution

```
yarn list v1.22.4
warning Filtering by arguments is deprecated. Please use the pattern option instead.
├─ del@5.1.0
│  └─ globby@10.0.2
├─ globby@11.0.1
├─ react-dev-utils@10.2.1
│  └─ globby@8.0.2
└─ webpack-dev-server@3.10.3
   └─ globby@6.1.0
```

dagda1:
> Looks great! Just had a mnor edit and question.

Dear oh dear, i  mistyped `minor`.  Good spot!

package.json
pittst3r: Hmm, I just tried removing this and everything looks good according to the test suite. This is what I see when I run `yarn why globby`. What do you see?

```
$ yarn why globby
yarn why v1.22.4
[1/4] 🤔  Why do we have the module "globby"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "globby@11.0.1"
info Has been hoisted to "globby"
info Reasons this module exists
   - "workspace-aggregator-ca912f35-7901-4f63-a828-3e39a7375d87" depends on it
   - Hoisted from "_project_#@bigtest#server#globby"
   - Hoisted from "_project_#@changesets#cli#@manypkg#get-packages#globby"
info Disk size without dependencies: "688KB"
info Disk size with unique dependencies: "1.3MB"
info Disk size with transitive dependencies: "6.32MB"
info Number of shared dependencies: 36
=> Found "react-dev-utils#globby@8.0.2"
info This module exists because "_project_#@bigtest#todomvc#react-scripts#react-dev-utils" depends on it.
info Disk size without dependencies: "24KB"
info Disk size with unique dependencies: "740KB"
info Disk size with transitive dependencies: "5.96MB"
info Number of shared dependencies: 44
=> Found "del#globby@10.0.2"
info This module exists because "_project_#@bigtest#webdriver#chromedriver#del" depends on it.
info Disk size without dependencies: "40KB"
info Disk size with unique dependencies: "1.59MB"
info Disk size with transitive dependencies: "7.66MB"
info Number of shared dependencies: 46
=> Found "webpack-dev-server#globby@6.1.0"
info Reasons this module exists
   - "_project_#@bigtest#todomvc#react-scripts#webpack-dev-server#del" depends on it
   - Hoisted from "_project_#@bigtest#todomvc#react-scripts#webpack-dev-server#del#globby"
info Disk size without dependencies: "40KB"
info Disk size with unique dependencies: "172KB"
info Disk size with transitive dependencies: "440KB"
info Number of shared dependencies: 13
✨  Done in 0.98s.
```

According to:

```
=> Found "globby@11.0.1"
info Has been hoisted to "globby"
```

`globby@11.0.1` is getting hoisted to the root of the project while the other versions are getting installed within the `node_modules` of their respective packages, so it should all be fine. Is there another way I can reproduce the conflict you were seeing?

packages/bundler/src/types.ts
```diff
@@ -0,0 +1,10 @@
+import { RollupWarning, RollupError } from 'rollup';
+
+export type BundlerError = RollupError;
+export type BundlerWarning = RollupWarning;
+
+export type BundlerMessage =
+  | { kind: 'START' }
+  | { kind: 'UPDATE' }
+  | { kind: 'WARN'; warning: BundlerWarning }
+  | { kind: 'ERROR'; error: BundlerError }
```
cowboyd: Our convention elsewhere in the codebase is to use `type` and lowercase for message discriminators. e.g. `{type: 'start' } 

Maybe not the best choice, but better to me consistent and then change the convention everywhere in one go.

package.json
dagda1: you are right, totally not needed.

No idea why I added now.

Removed

packages/project/src/index.ts
```diff
@@ -108,3 +108,18 @@ export function defaultConfig(configFilePath: string): ProjectOptions {
...
+
+    Object.setPrototypeOf(this, new.target.prototype);
+  }
+}
+
+export function assert(condition: unknown, msg?: string): asserts condition {
+  if (!condition) {
+    throw new AssertionError(msg || 'assertion error');
+  }
+}
```
cowboyd: I'm not sure why `assert` helper goes here? `@bigtest/project` is for working doing things like finding the bigtest config, etc... It looks like this also isn't being used anywhere? We may need a `@bigtest/assert` package, or just use [assert](https://www.npmjs.com/package/assert) from npm?

packages/server/src/manifest-builder.ts
```diff
@@ -77,59 +76,66 @@ function* processManifest(options: ManifestBuilderOptions): Operation {
...
+        break;
+      case 'UPDATE':
+        console.debug("[manifest builder] received bundle update");
+        
+        let path: string = yield processManifest(options);
+        
+        bundlerSlice.update((previous) => {
+          // as this is a typescript assertion it does more than just check the condition
+          // it will type narrow the discriminated union based on the kind discriminator
+          assert(previous.kind === 'BUILDING', `invalid transition from ${previous.kind} to 'GREEN'`);
```
cowboyd: oh I see, here we use it. Should this be a warning?

dagda1: So this does more than just throw if the condition is not met.  If the `previous.kind === 'BUILDING;`

Then I can do this

`return { ...previous, kind: 'GREEN', path };`

the `path property` only exists when the  `kind` or `type` is green.  the `assert` here narrows the type. 

If `previous.kind !== 'BUILDING;'` then it is cast to `never`.  It can't be a warning.  It needs to stop the control flow in the same way as `throw` does

dagda1: I put it in `@bigtest/project` because I thought that might be where code that is used by all project goes.

packages/bundler/src/types.ts
dagda1: I have changed it to `type`, I do believe `kind` is the convention and `type` is a redux thing.  

I could definitely be wrong and it does not matter too much and as you say, it should be consistent.

packages/project/src/index.ts
dagda1: this does more than assert, check out the `: asserts condition {` on line 120

[This explains it](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions)

pittst3r (approved):

packages/project/src/index.ts
cowboyd: Oh, ok. Do we need it in a shared location? `@bigtest/project` isn't really the place for that. It's more for project configuration (directory, testfiles, etc...)

cowboyd: To be honest, if it is shared, I don't know the right place for it. Maybe we create a `@bigtest/lib` package?

packages/server/src/manifest-builder.ts
```diff
@@ -77,59 +76,66 @@ function* processManifest(options: ManifestBuilderOptions): Operation {
...
+        break;
+      case 'UPDATE':
+        console.debug("[manifest builder] received bundle update");
+        
+        let path: string = yield processManifest(options);
+        
+        bundlerSlice.update((previous) => {
+          // as this is a typescript assertion it does more than just check the condition
+          // it will type narrow the discriminated union based on the type discriminator
+          assert(previous.type === 'BUILDING', `invalid transition from ${previous.type} to 'GREEN'`);
```
cowboyd: These type-narrowing assertions are really cool! However, they will cause the bundler, and the entire orchestrator to crash, we should indicate clearly that this is a bug in BigTest and should be reported... e.g.

```ts
assert(previous.type === 'BUILDING', `
INTERNAL ASSERTION FAILURE:
While bundling test files, an invalid state transition from ${previous.type} to 'GREEN' was received.
this is almost certainly a bug in BigTest, and we would be very grateful if you would report
it at https://github.com/thefrontside/bigtest/issues/new`
`);

**Review by cowboyd — COMMENTED** [review]
We're getting really close. I'm in a bit of a head-scratch as to where to put the assert code.

File: packages/server/src/manifest-builder.ts (hunk 9 above)
Comment (dagda1) [reply packages/server/src/manifest-builder.ts]: @cowboyd we could have a template for this type of thing or do we have one?  These errors are definitely for us and of course, should never happen but .....things do happen

Comment (dagda1) [reply packages/server/src/manifest-builder.ts]: @cowboyd I might have missed the point of the assertions, I think having a generic one might actually be defeating the purpose.  I think they should actually be for specific purposes.

Let me try another commit.

Comment (dagda1) [reply packages/server/src/manifest-builder.ts]: @cowboyd I love the type narrowing and assertions that actually live in the code.  This is where we get into the type system enforcing correctness and not just for plain old checking you can't pass a string to a function that expects a number

Comment (dagda1) [reply packages/server/src/manifest-builder.ts]: @cowboyd we could also use rollup to strip the assertions for the prod build

Comment (dagda1) [reply packages/server/src/manifest-builder.ts]: @cowboyd I have added more specialised assertions in `@bigtest/server`.  

If this is ok then I will remove the generic one and add a couple of tests.

Comment (cowboyd) [reply packages/server/src/manifest-builder.ts]: `asserts current is R`

🤯 🤯 🤯 

@dagda1 Sounds like a plan!

File: packages/server/src/assertions/bundler-assertions.ts [hunk 10]
```diff
@@ -0,0 +1,30 @@
...
+
+export class BundlerAssertionError extends Error {
+  constructor(message: string) {
+    super(`INTERNAL ASSERTION FAILURE:
+    While bundling test files, ${message} was received.
+    this is almost certainly a bug in BigTest, and we would be very grateful if you would report
+    it at https://github.com/thefrontside/bigtest/issues/new
+    `)
+
+    Object.setPrototypeOf(this, new.target.prototype);
```
cowboyd, line 11: 🤔 not grokking this 🤔

cowboyd (approved):
🕺 🕺 🕺

packages/server/src/assertions/bundler-assertions.ts
dagda1: The problem is that Javascript's built-in class Error breaks the prototype chain by switching the object to be constructed (i.e. this) to a new, different object, when you call super and that new object doesn't have the expected prototype chain, i.e. it's an instance of Error not of CustomError.

```typescript
class CustomError extends Error {
  constructor(message?: string) {
    super(message); // 'Error' breaks prototype chain here
    Object.setPrototypeOf(this, new.target.prototype); // restore prototype chain
  }
}
```
[docs]( https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html)

Using new.target has the advantage that you don't have to hardcode the prototype

PR #451, opened 2020-08-07
Interactor composition

pittst3r:
## Motivation

Currently within an interactor action you do not have access to the current interactor, you only have access to the underlying element. This precludes the ability to use other interactors scoped to the current interactor within the action. We would like to be able to use interactors within actions as this supports the composition of interactors.

## Approach

1. Supply the interactor instance to actions to support composition using `.find()`:

```ts
const Datepicker = createInteractor<HTMLDivElement>("datepicker")({
  // ...
  actions: {
    toggle: async interactor => {
      await interactor.find(TextField.byPlaceholder("YYYY-MM-DD")).click();
    }
  }
});
```

2. Replace the `Interactor#resolve()` method with `Interactor#perform()`:

```ts
const Datepicker = createInteractor<HTMLDivElement>("datepicker")({
  // ...
  actions: {
    toggle: async interactor => {
      await interactor.perform(element => { element.querySelector('input')?.click() });
    }
  }
});
```

3. Supply a helper function called `perform()` that provides sugar for no. 2 above:

```ts
const Datepicker = createInteractor<HTMLDivElement>("datepicker")({
  // ...
  actions: {
    toggle: perform(element => { element.querySelector('input')?.click() })
  }
});
```


jgnieuwhof (approved):
Cool! I'm not familiar enough with the space to catch subtleties, but the concept and implementation make sense to me.

I assume this feature is here for element scoping? Wondering if it'd be helpful to have a test that tries to locate an element outside of the passed interactor's sub-tree. Just a thought 🤷‍♂️.

pittst3r:
@jgnieuwhof Good suggestion. Just added a test demonstrating this.

packages/interactor/test/create-interactor.test.ts
```diff
@@ -42,7 +45,22 @@ const TextField = createInteractor<HTMLInputElement>('text field')({
...
+const Datepicker = createInteractor<HTMLDivElement>("datepicker")({
+  selector: "div.datepicker",
+  defaultLocator: element => element.querySelector("label")?.textContent || "",
+  filters: {
+    open: element => !!element.querySelector("div.calendar"),
+    month: element => element.querySelector("div.calendar h4")?.textContent
+  },
+  actions: {
+    toggle: async ({ interactor }) => {
+      await interactor.find(TextField.byPlaceholder("YYYY-MM-DD")).click();
```
cowboyd, line 62: Are actions allowed to be async?

pittst3r: Yes, we would need to intentionally swallow the promise to not support it.

cowboyd: Yeah, it's just a little weird because the idea originally that all element actions need to be synchronous so that the element reference does not get stale.

It feels like we should be using _either_ element _or_ interactor, but never both. Imagine you had something like:

```ts
actions: {
  toggle: async ({ interactor, element }) => {
    await interactor.find(TextField.byPlaceholder("YYYY-MM-DD")).click();
    element.data.clicked = true;
  }
}
```

`element` would be stale, and this should be an error, and in the case where you _do_ use element, a promise return value is unacceptable. What do you think?

pittst3r: > element would be stale, and this should be an error, and in the case where you do use element, a promise return value is unacceptable.

This is starting to get a little complex for my liking. I don't think we should treat the usage of arguments differently based on what is being returned. That does not seem like it would make for a simple API.

Need to think about this more. Maybe we can talk tomorrow about it.

pittst3r: Here's an idea:

1. Add an async method to the interactor interface that fetches the element; basically `unsafeSyncResolve()` wrapped in `converge()`
1. Provide only the interactor interface to actions
1. Don't run actions within a `converge()` (otherwise we would be doubling up)

A nice side-effect of this IMO is that actions would not be run more than once, which in my experience people have a hard time understanding/retaining.

cowboyd: What would an example of this look like?

pittst3r: Using your example problematic code from above:

```ts
actions: {
  toggle: async interactor => {
    await interactor.find(TextField.byPlaceholder("YYYY-MM-DD")).click();
    let element = await interactor.element();
    element.data.clicked = true;
  }
}
```

However, one could still introduce a bug like this:

```ts
actions: {
  toggle: async interactor => {
    let element = await interactor.element();
    await interactor.find(TextField.byPlaceholder("YYYY-MM-DD")).click();
    element.data.clicked = true;
  }
}
```

We could avoid this footgun with a callback API, but not sure it is desirable to mix callbacks and promises:

```ts
actions: {
  toggle: async interactor => {
    await interactor.find(TextField.byPlaceholder("YYYY-MM-DD")).click();
    await interactor.withElement(element => {
      // We would enforce this as synchronous
      element.data.clicked = true;
    });
  }
}
```

cowboyd: of those options, I think that number 3 is the safest. where you have an interactor.perform() or interactor.withElement() method that lets you access the element in the context of full validation. 

One possible syntax is we treat actions as first class objects, and then if we see a function, we treat it as the base case of lifting a simple synchronous function into an action context:

```ts
import { action } from '@bigtest/interactor';

actions: {
  toggle: action(async (interactor) => {
    await interactor.find(TextField.byPlaceholder("YYYY-MM-DD")).click();
    await interactor.perform(element => element.data.clicked = true);
  })
}
```

but still allow for simple actions:

```ts
actions: {
  click: element => element.click()
}
```

which is just sugar for:

```ts
actions: {
  click: action(async (interactor) => await interactor.perform(element => element.click()))
}
```

This is easy to type at both runtime and compile time because the action would be either a `Function`, or an instance of `Action`:

```ts
type ActionSpec = Action | (element: Element) => void;
```

pittst3r: I like this, but it feels like the inverse of what I would expect, and I think we can simplify it. What do you think of this?

```ts
actions: {
  toggle: async interactor => {
    await interactor.find(TextField.byPlaceholder("YYYY-MM-DD")).click();
    await interactor.perform(element => element.data.clicked = true);
  }
}
```

and

```ts
actions: {
  toggle: perform(element => element.click())
}
```

where the sugar happens here:

```ts
function perform(fn) {
  return interactor => interactor.perform(fn);
}
```

pittst3r: I ran with this to see what it would look like, no obligations. Typing was surprisingly easy.

cowboyd: The default should align with the 80/20 or 90/10 case. The only problem is that we don't have any data. What do we think will be the most common usage?

pittst3r: In this scenario I see `perform()` as a power user shortcut for low-level operations where we don't have a base interactor defined already. I'm hoping eventually we'll have base interactors and you won't need to `perform` very often. All that to say, I think `async interactor => interactor.find(...).doSomething()` would be the 80% use case.

cowboyd (approved):

PR #460, opened 2020-08-14
Locator stacking

pittst3r:
## Approach

Given:

```ts
let Div = createInteractor<HTMLDivElement>('div')({
  // These locators are tried in order
  defaultLocator: ['byId', 'byText'],
  locators: {
    byId: (element) => element.id,
    byText: (element) => element.textContent || ""
  }
});
```

and:

```html
<div id="some-div-id">Hello, world.</div>
```

Both of these would match the element:

```ts
Div('some-div-id');
// or
Div('Hello, world.');
```

### DIFF
--- a/packages/interactor/types/locators.ts
+++ b/packages/interactor/types/locators.ts
@@ -1,0 +1,39 @@
+import { createInteractor } from '../src/index';
+
+const Link = createInteractor<HTMLLinkElement>('link')({
+  selector: 'a',
+  defaultLocator: ['byHref', 'byTitle'],
+  locators: {
+    byHref: (element) => element.href,
+    byTitle: (element) => element.title
+  }
+});
+
+Link('foobar');
+
+Link.byHref('foobar');
+
+// cannot use wrong type argument on locator
+// $ExpectError
+Link.byHref(123);
+
+// cannot use locator which is not defined
+// $ExpectError
+Link.byValue('foobar');
+
+const Button = createInteractor<HTMLButtonElement>('button')({
+  selector: 'button',
+  defaultLocator: 'byText',
+  locators: {
+    byText: (element) => element.textContent || ""
+  }
+});
+
+Button('foobar');
+
+const TextField = createInteractor<HTMLInputElement>('text field')({
+  selector: 'input[type="text"]',
+  defaultLocator: (element) => element.placeholder,
+});
+
+TextField('foobar');
--- a/.changeset/default-locator-ref.md
+++ b/.changeset/default-locator-ref.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/interactor": minor
+---
+
+Allow an `InteractorSpecification`'s `defaultLocator` to reference by name a locator in the `LocatorSpecification`.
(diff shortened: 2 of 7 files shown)


packages/interactor/types/locators.ts
```diff
@@ -0,0 +1,43 @@
...
+    byText: (element) => element.textContent || ""
+  }
+});
+
+Button('foobar');
+
+const TextField = createInteractor<HTMLInputElement>('text field')({
+  selector: 'input[type="text"]',
+  // $ExpectError
+  defaultLocator: 'byLabel',
```
pittst3r: @dagda1 This is not erroring as I'd like it to. `byLabel` is not in the map of `locators` below, so I want that to be a type error.

packages/interactor/types/locators.ts
```diff
@@ -0,0 +1,43 @@
...
+
+Link.byHref('foobar');
+
+// cannot use wrong type argument on locator
+// $ExpectError
+Link.byHref(123);
+
+// cannot use locator which is not defined
+// $ExpectError
+Link.byValue('foobar');
```
pittst3r, line 22: @dagda1 This is also not erroring, but I feel like once one of these problems is resolved the other will be resolved too.

dagda1: This should fix it

```typescript
export type LocatorImplementation<E extends Element, S extends InteractorSpecification<E, L>, L extends LocatorSpecification<E>> = {
  [P in keyof S['locators']]: S['locators'][P] extends ((element: E, ...args: unknown[]) => unknown) ? (value: string, filters?: FilterImplementation<E, S>) => InteractorInstance<E, S, L> : never;
}
```

pittst3r: Of note is that this is not erroring as expected only because of my modifications in attempt to get the other expected error working.

cowboyd:
😎 very cool! Is this in addition to the way the default locator already works? In other words, if it's a function, then it works as before?

pittst3r:
@cowboyd Yes, this is additive and backwards-compatible.

jnicklas:
This is great and there are some use-cases where I think this is very much needed, for example one headscratcher is buttons which can be either `<button>some text here</button>` or `<input type="button" value="some text here"/>`. The user will want to do `Button("some text here")` and have that match both since a user will perceive both of these the same even though they use different HTML tags.

Since the method of matching is different (`textContent` vs `value` attribute), we will need different functions to match either. So in this case locator "stacking" is very much needed.

However I want to add a bit of a caveat to this: Capybara uses this type of "stacking" very heavily. If you do `fill_in("Foo", with: "Bar")` in Capybara, it will match all kinds of things, the `label`, `id`, `placeholder` and probably other things too. In my experience this is actually both pretty confusing, as well as a pretty big performance hit, and I would not encourage using locator stacking for this kind of purpose.

tl;dr: I like it and think it's necessary, but am cautious about this being overused.

cowboyd:
@pittst3r Have you been missing this feature as you implement ui-users? If not, then maybe we go with a "less is more" approach and just punt on this, or close for now?

pittst3r:
@cowboyd Yeah, that's why I've been sitting on this. Wanted to see just how much we actually want it. I will close as I have not been missing this.

PR #461, opened 2020-08-14
Run Interactor tests in CI

pittst3r:
## Motivation

Interactor tests were not running in CI.

## Approach

Run the Interactor tests in CI.

## To-do

Can we make it so that the list of projects to test is not manually managed?


cowboyd (approved):
whoops!

Issue #430, opened 2020-07-31
Rethink CI command with an empahsis on very helpful diagnostics


cowboyd:
Right now, our error handling is competent, but not great. For example, if we have a failed assertion in this test suite:

```ts
export default test("Simulated GraphQL API")
  .step("start GraphQL simulator", async () => ({
    simulator: await startBrowserSimulation()
  }))
  .step("create two companies", async ({simulator}) => {
    let [first, second] = simulator.createList('company', 2);

    return { first, second };
  })
  .step("fetch companies over graphql", async () => {
    let query = `{ companies { id name } }`;
    return {
      data: await request('/graphql', query)
    };
  })
  .assertion("it returns two companies", async ({ data } ) => {
    await assertEqual(data.companies.length, 3);
  })
  .assertion("the returned companies are serialized properly", async ({ data, first, second }) => {
    await assertEqual(data.companies[0].name, first.name);
    await assertEqual(data.companies[1].name, second.name);
  })

```

The output upon an assertion failure looks like this:

```
$ yarn bigtest test
yarn run v1.22.4
✓ [step]       Simulated GraphQL API -> start GraphQL simulator
✓ [step]       Simulated GraphQL API -> create two companies
✓ [step]       Simulated GraphQL API -> fetch companies over graphql
⨯ [assertion]  Simulated GraphQL API -> it returns two companies
|   AssertionError: expected 3 == 2
|       at _callee$ (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:66693:23)
|       at tryCatch (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:196:23)
|       at Generator.invoke [as _invoke] (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:413:28)
|       at Generator.next (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:249:27)
|       at asyncGeneratorStep (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:861:28)
|       at _next (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:883:13)
[...]

PR #462, opened 2020-08-15
Add new tree output

RobbieTheWagner:
Screen Shot 2020-09-02 at 3 46 46 PM

### DIFF
--- a/.changeset/empty-ducks-juggle.md
+++ b/.changeset/empty-ducks-juggle.md
@@ -1,7 +1,0 @@
----
-"@bigtest/cli": minor
-"@bigtest/project": minor
-"@bigtest/server": minor
----
-
-Add `--show-tree` option, improve output tree
--- a/.changeset/empty-ducks-juggle.md
+++ b/.changeset/empty-ducks-juggle.md
@@ -1,0 +1,7 @@
+---
+"@bigtest/cli": minor
+"@bigtest/project": minor
+"@bigtest/server": minor
+---
+
+Add `--show-tree` option, improve output tree
--- a/.changeset/tree-output-for-default-cli-output.md
+++ b/.changeset/tree-output-for-default-cli-output.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/cli": patch
+---
+default test output for CLI now shows dots for progress, and then the
+path to the failure (if any) as a tree.
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -48,9 +48,11 @@
     "@bigtest/server": "^0.11.0",
     "@effection/node": "^0.8.0",
     "capture-console": "^1.0.1",
+    "chalk": "^4.1.0",
     "deepmerge": "^4.2.2",
     "effection": "^0.7.0",
-    "json5": "^2.1.3"
+    "json5": "^2.1.3",
+    "ololog": "^1.1.161"
   },
   "volta": {
     "node": "12.16.0",
--- a/packages/cli/src/format-helpers.ts
+++ b/packages/cli/src/format-helpers.ts
@@ -1,3 +1,4 @@
+import * as chalk from 'chalk';
 import { ProjectOptions } from '@bigtest/project';
 import { ResultStatus } from '@bigtest/suite'
 import { RunResultEvent } from './query'
@@ -23,9 +24,9 @@ export function icon(event: RunResultEvent) {
 
 export function statusIcon(status: string) {
   if(status === 'ok') {
-    return "✓";
+    return chalk.green("✓");
   } else if(status === 'failed') {
-    return "⨯";
+    return chalk.red("⨯");
   } else if(status === 'disregarded') {
     return "⋯";
   }
--- a/packages/cli/src/format-helpers.ts
+++ b/packages/cli/src/format-helpers.ts
@@ -16,11 +16,17 @@ export type Summary = {
 export function icon(event: RunResultEvent) {
   if(event.type.match(/^:running$/)) {
     return "↻";
-  } else if(event.status === 'ok') {
+  } 
+
+  return statusIcon(event.status || '');
+}
+
+export function statusIcon(status: string) {
+  if(status === 'ok') {
     return "✓";
-  } else if(event.status === 'failed') {
+  } else if(status === 'failed') {
     return "⨯";
-  } else if(event.status === 'disregarded') {
+  } else if(status === 'disregarded') {
     return "⋯";
   }
 }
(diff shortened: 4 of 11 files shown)


packages/project/src/index.ts
```diff
@@ -26,6 +26,7 @@ export function *loadConfigFile(configFilePath: string): Operation<ProjectOption
 
 export type ProjectOptions = {
   port: number;
+  showTree: boolean;
```
cowboyd: I wonder if `showTree` should be a project option. On the one hand, I like the idea that there is a 1:1 correspondence between CLI options and project config options, but it does seem odd that something that is only used by the CLI is being passed by the orchestrator. On the other hand, it also allows making the option persistent.

RobbieTheWagner: I didn't know how else to get the option passed down to where we output to the console. Open to suggestions for sure!

cowboyd: Yeah, I'm thinking purely from a philosophical sense. Like, it would be awesome if every single config option could be overridden from the command line, so this probably is the best approach? Let's roll with this.

RobbieTheWagner:
@cowboyd any idea how we can make TS happy here? I think the types are correct, but we're getting failures.

cowboyd:
@rwwagner90 It looks like you need to include make the showTree option into the project options inside server. Since the interface changed, in `ProjectConfig`, we need to pass a object that implements it into `createOrchestrator`.

The reason it wasn't caught by the build step was that our tests are only compiled at runtime by `ts-node` as they are required, so we don't actually catch type errors in the tests until we actually run them.

RobbieTheWagner:
@cowboyd I am not following. I think `ProjectOptions` is always imported like so `import { ProjectOptions } from '@bigtest/project';` and it has been added to the project package here https://github.com/thefrontside/bigtest/pull/462/files#diff-c06412528738d2294a5adac6d4dc6847R29

Thoughts on how to make it work?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/bigtest/pull/462">Add `--show-tree` option by rwwagner90 · Pull Request #462 · thefrontside/bigtest</a></strong></div></blockquote>

cowboyd:
The problem is that this object does not conform to the new `ProjectOptions` interface which now requires the `showTree` field https://github.com/thefrontside/bigtest/blob/v0/packages/server/test/helpers.ts#L54

The reason that it only shows up when running the tests is that test code is only compiled "on-the-fly" when we run it with `ts-node`

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/bigtest">thefrontside/bigtest</a></strong></div><div>Tests that speed up development, not slow it down. Contribute to thefrontside/bigtest development by creating an account on GitHub.</div></blockquote>

taras (approved):
👍

jnicklas:
Sorry to throw a bit of a curveball here, I've been away on vacation for a while and still catching up on things.

How do you feel about making this a separate formatter, rather than adding it as an option to `lines`. So you would do something like:

```sh
bigtest test --formatter tree
```

This is sort of what I had in mind with formatters, and I started adding some infrastructure towards this end. If you look at the structure exported from `./src/formatters/lines.ts`, it has a `type: "streaming"` property. My thinking was that a formatter could basically pick between being a streaming formatter, or a stateful formatter.

So we could have various streaming formatters (e.g dots, XUnit), and various stateful formatters (e.g. tree view, or just a list of dots which fill in, etc...). Additionally, with some tweaks we could support third-party formatters pretty easily.

cowboyd:
This will definitely be a good thing because with our platform, because the creativity that people will be able to have with the graphql API should make for some pretty interesting UI. 

We can also leverage the formatting work that @rwwagner90 has been doing to implement this API on the streaming lines formatter to use icons and indents rather than absolute paths of each step. Question is, how best to proceed with the different areas of progress.

cowboyd:
@jnicklas after looking at this, what if we lean on our ever-improving subscription API as the basis for our reporting API?

```ts
type Reporter = (testRun: TestRun, subscription: Subscription<TestEvent, TestSummary>): Operation<void>;
```

Your basic reporter would be something like:

```ts
export function* report(testRun, subscription) {
  printPreamble(testRun);
  let summary = yield subscription.forEach(function*(event)) {
    printEvent(event);  
  });
  printSummary(summary);
}
```

The advantage of doing it this way is that representation and output is completely separated from computation of the metrics of the test run. In other words, the `TestSummary` can all be computed by core CLI. The other nice thing is that because it's just a function, we can decorate reporters using nothing more than function composition. 

If we had that, then adding something that streamed test run statistics to a cloud service could be easily added. Also, we could support multiple reporters so you could, for example, render to stdout and also render xml to a `junit.xml` file at the same time.

cowboyd (changes requested):
After pairing with @rwwagner90 yesterday, and seeing the comment from @jnicklas about the proto-formatting API, I think I have a plan for getting this over the finish line

Our goal is to have [an awesome default formatter](https://github.com/thefrontside/bigtest/issues/430) and now that we've made a couple of attempts, and think I can see how we can accomplish this in phases. In order to realize that full vision though, I think that we will have to [make some changes](https://github.com/thefrontside/bigtest/issues/487) to the underlying reporting infrastructure. However, I think that we can utilize all of the work here to proceed forward in the following steps.

1. Limit the scope of the current PR to be just an enhancement of the streaming output to show only dots and the sparse failure tree. (no --show-tree) option.
2. Introduce reporter API, and change the reporter enhanced streaming formatter to use that new API
3. Introduce a more formal plugin system for swapping out reporters and also passing options to those reporters so they can be configured.
4. Use the plugin system to enhance  the default reporter to have an optional --show-tree to show the full tree)
5. enhance default reporter to render per-agent progress bars per-agent (but dots on CI)
6. Profit 💸 

 We can hash out the formal plan on Monday, and once we agree on the work I can flesh out tickets, but fore the scope of this PR here, let's reduce its scope to (1).

RobbieTheWagner:
@cowboyd why not keep this PR as is, since we'll want something like `show-tree` eventually? The current setup here allows for hiding the whole tree unless the option is passed, which could be nice for interim CI output.

cowboyd:
@rwwagner90 heads up: In order to get past the new MacOS test run introduced with #486, you'll have to rebase onto origin/v0 and then close and re-open this PR to pull in the new checks.

RobbieTheWagner:
@cowboyd @jnicklas looks like we have a lot of conflicts here, since the old test formatters were updated. Not sure how to proceed. Should we incorporate the stack trace changes into the changes I made to show the tree?

RobbieTheWagner:
@cowboyd I am unsure why the error stack is not shown while running `yarn test` in the cli package here. It is shown when I locally run `yarn bigtest test`. Any ideas?

cowboyd (approved):
Looks good. Way to go @rwwagner90 !

PR #464, opened 2020-08-16
type in lens in Slice to ManualLens instead of any

dagda1:
Currently in Slice, lens is typed to any

```typescript

export class Slice<S, A> implements Subscribable<S, undefined> {
  //eslint-disable-next-line @typescript-eslint/no-explicit-any
  lens: any;
```

which has the knock-on effect of having to do a double cast in `get` and `set`

```typescript
  get(): S {
    return (view(this.lens)(this.state) as unknown) as S;
  }
```

This PR types `lens` to the provided ramda type of `ManualLens` and removes the double casts.

### DIFF
--- a/.changeset/type-lens-in-slice.md
+++ b/.changeset/type-lens-in-slice.md
@@ -2,4 +2,4 @@
 "@bigtest/atom": patch
 ---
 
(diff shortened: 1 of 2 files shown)


taras (approved):
Very nice

PR #467, opened 2020-08-18
Nest bundle inside manifest

dagda1:
This is a follow up to #450  makes bundler a top-level entity in the graphql schema:

```javascript
t.field("bundler", {
  type: "Bundler"
});
.....
enumType({
  name: "BundlerType",
  members: [
    'UNBUNDLED',
    'BUILDING',
    'GREEN',
    'ERRORED'
  ]
}),
objectType({
  name: "Bundler",
  definition(t) {
    t.field("type", {
      type: 'BundlerType'
    });
    t.string("path", {
      nullable: true
    }),
    t.list.field("errors", {
      type: "Error",
      nullable: true
    }),
    t.list.field("warnings", {
      type: "Error",
      nullable: true
    })
  }
}),
```

### DIFF
--- a/packages/server/src/schema.ts
+++ b/packages/server/src/schema.ts
@@ -6,6 +6,7 @@ import {
   subscriptionField,
   stringArg,
   makeSchema,
+  enumType,
 } from "@nexus/schema";
 
 export const schema = makeSchema({
@@ -52,6 +53,10 @@ export const schema = makeSchema({
 
         t.field("manifest", {
           type: "Test",
+        });
+
+        t.field("bundler", {
+          type: "Bundler"
         });
 
         t.list.field("testRuns", {
@@ -165,6 +170,34 @@ export const schema = makeSchema({
         });
         t.list.field("children", {
           type: "Test"
+        })
+      }
+    }),
+    enumType({
+      name: "BundlerType",
+      members: [
+        'UNBUNDLED',
+        'BUILDING',
+        'GREEN',
+        'ERRORED'
+      ]
+    }),
+    objectType({
+      name: "Bundler",
+      definition(t) {
+        t.field("type", {
+          type: 'BundlerType'
+        });
+        t.string("path", {
+          nullable: true
+        }),
+        t.list.field("errors", {
+          type: "Error",
+          nullable: true
+        }),
+        t.list.field("warnings", {
+          type: "Error",
+          nullable: true
         })
       }
     }),
--- a/.changeset/bundler-entity
+++ b/.changeset/bundler-entity
@@ -1,0 +1,5 @@
+---
+"@bigtest/server": patch
+---
+
+Make bundler a top-level entity


packages/server/src/schema.ts
```diff
@@ -151,7 +151,24 @@ export const schema = makeSchema({
       }
     }),
     objectType({
-      name: "Test",
+      name: "bundler",
```
cowboyd: This should be Pascal cased since it is effectively a type name and will correspond to a type declaration in our graphql. E.g.

```graphql
type bundler {
}
```

cowboyd (commented):
Now that I see this, I'm scratching my head and thinking that perhaps it should be the other way around, the manifest doesn't have a bundler... rather, the bundler has a manifest.

Currently, we had this weird chicken and the egg problem where we fill in the manifest with an "empty" manifest to start out with, but that's really not what's going on. I think maybe we should move to be more like the `Maybe` where you either have a manifest our you don't depending on the state of the bundler. If everybody who wants to access the manifest state has to unpack it from the bundler, then, like `Maybe`, it will force them to contemplate the idea that the manifest might be errored, or not yet present.

What do you think?

dagda1:
> Now that I see this, I'm scratching my head and thinking that perhaps it should be the other way around, the manifest doesn't have a bundler... rather, the bundler has a manifest.
> 
> Currently, we had this weird chicken and the egg problem where we fill in the manifest with an "empty" manifest to start out with, but that's really not what's going on. I think maybe we should move to be more like the `Maybe` where you either have a manifest our you don't depending on the state of the bundler. If everybody who wants to access the manifest state has to unpack it from the bundler, then, like `Maybe`, it will force them to contemplate the idea that the manifest might be errored, or not yet present.
> 
> What do you think?

It does make sense and a manifest having a bundler is not quite right.

So in that case, the following is true

- The bundler has the atom
- The manifest generation and building are subscribable resources that the bundler subscribes to.

I'd like to double check I have the above right before implementing it.

packages/server/src/schema.ts
dagda1: Oh yeah, typo from me.

cowboyd:
>The bundler has the atom
>The manifest generation and building are subscribable resources that the bundler subscribes to.

We have a naming problem :) 

My instinct is telling me that we should move all bundling repsonsibilities outside of the orchestrator, because eventually test bundling is going to have to be the responsibilty of the driver, and not of the orchestrator. This was exactly the problem that I ran into with the node driver was that the bundler made too many assumptions about the fact that it would be running on the web, and so when I tried to use the same bundler for running tests in node, it completely flubbed.

My vote is to kick this can down the road, since we can achieve our original objective of reporting TS errors via the atom, and we can do that now. What if we  make the `Bundler` a top level entity in the `Query` schema, and then refactor when it comes to multi-driver support?

dagda1:
> What if we make the `Bundler` a top level entity in the `Query` schema, and then refactor when it comes to multi-driver support?

Sounds like a plan.  I'll do that

dagda1:
closing in favour of #475

PR #468, opened 2020-08-19
♻ Extract @bigtest/client from @bigtest/server

cowboyd:
Motivation
-----------

Right now the `Client` class is packaged as part of the `@bigtest/server` since the only other package that consumes it is CLI which also depends on server. However, this is bad for two reasons: First, it is very difficult to evolve the client protocol independently as our needs change, but most important, we want to be able to connect to a bigtest orchestrator from within environments that do _not_ require it. Specifically, if we want to make browser UI for bigtest, we'll need a lightweight library that runs independently and only depends on websockets. Also, if we want to communicate with orchestrators from within places like bigtest suites, we'll need a client there too.

Approach
----------

There is a lot of legacy pattern and code in the client API, including dependencies on `Mailbox`, but the idea was to extract the client without _any_ change whatsoever to its API. That way, the the internal function remains exactly the same and so the extraction goes as smoothly as possible. Once we have the boundary there, we can refactor the client internals and expand the protocol iteratively.


taras (approved):
👍

packages/client/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "license": "MIT",
+  "files": [
+    "dist/*",
+    "README.md"
+  ],
+  "scripts": {
+    "lint": "eslint '{src,test}/**/*.ts'",
+    "mocha": "mocha -r ts-node/register",
+    "test": "mocha -r ts-node/register test/**/*.test.ts",
+    "prepack": "tsc --outdir dist/esm --sourcemap --module es2015 && tsc --outdir dist/cjs --sourcemap --module commonjs && tsc --outdir dist/types --emitDeclarationOnly"
```
dagda1, line 19: should we be using rollup rather than `tsc` direct?

I don't mean for this PR but more of a general question.

packages/client/test/helpers.ts
```diff
@@ -0,0 +1,89 @@
+import { Operation, resource, main } from 'effection';
+import { Channel } from '@effection/channel';
+import { Subscribable } from '@effection/subscription';
+import { express, Socket } from '@bigtest/effection-express';
+import { Mailbox } from '@bigtest/effection';
```
dagda1, line 5: if `Mailbox` is a pattern we don't want to use anymore should we use something else?

Unless I have got this wrong about `Mailbox` being deceased.

packages/client/package.json
taras: Yes, we're using [`tsdx` in `effection`](https://github.com/thefrontside/effection/blob/v0/packages/channel/package.json#L17). I thought we're using it everywhere but it doesn't look like it. I'll make an issue to switch it.

cowboyd: Thanks! I just copied this package from `@bigtest/globals` @taras when you create that ticket can you create one to do globals as well?

packages/client/test/helpers.ts
cowboyd: @dagda1 Absolutely agree. My goal with this PR is to extract the client as seamlessly as possible, wrap a test suite around it (which it didn't have before). 

Now that those are in place, we can refactor it internally in order to remove Mailbox and go with a `Subscription` based approach in follow-on PR's

PR #470, opened 2020-08-19
Show a nice error when running test command without server running

### LINKED ISSUE #427: Better error if `test` command cannot connect to an orchestrator
If you try to run the `test` command from the CLI, but no orchestrator is started, you get a websocket error:

```
$ /Users/cowboyd/Code/CVEX/trade/simulation/node_modules/.bin/bigtest test
Error: websocket server closed connection unexpectedly: [1006] connection failed
    at /Users/cowboyd/Code/CVEX/trade/simulation/node_modules/@bigtest/server/dist/src/client.js:21:23
    at Generator.next (<anonymous>)
    at /Users/cowboyd/Code/CVEX/trade/simulation/node_modules/effection/dist/effection.js:1:8094
    at e.r.advance (/Users/cowboyd/Code/CVEX/trade/simulation/node_modules/effection/dist/effection.js:1:7730)
    at e.r.trapExit (/Users/cowboyd/Code/CVEX/trade/simulation/node_modules/effection/dist/effection.js:1:8056)
    at t.n.finalize (/Users/cowboyd/Code/CVEX/trade/simulation/node_modules/effection/dist/effection.js:1:5567)
    at t.n.resume (/Users/cowboyd/Code/CVEX/trade/simulation/node_modules/effection/dist/effection.js:1:4921)
    at e.r.advance (/Users/cowboyd/Code/CVEX/trade/simulation/node_modules/effection/dist/effection.js:1:7776)
    at e.r.trapExit (/Users/cowboyd/Code/CVEX/trade/simulation/node_modules/effection/dist/effection.js:1:8056)
    at t.n.finalize (/Users/cowboyd/Code/CVEX/trade/simulation/node_modules/effection/dist/effection.js:1:5567)
```

This is true, but not helpful. Instead, we should indicate that the bigtest server is not currently running, and that you can start it with an invocation to `bigtest server`

taras:
Closes #427 

## Motivation

People should know what is happening when an error occurred. They should not have to look at stack traces to figure out what it means. When running tests without bigtest server present, we would get an ugly stack trace that said: `websocket server closed connection unexpectedly`. Instead, we want to show an error that suggests to run `bigtest server` before running tests.

## Approach

Detect when we get `websocket server closed connection unexpectedly` and show error: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.` instead.

### DIFF
--- a/packages/cli/src/run-test.ts
+++ b/packages/cli/src/run-test.ts
@@ -15,7 +15,7 @@ export function* runTest(config: ProjectOptions, formatter: StreamingFormatter):
     try {
       return yield Client.create(uri);
     } catch (e) {
-      if (e.message.includes('websocket server closed connection unexpectedly')) {
+      if (e.name === 'ConnectionAttemptFailed') {
         throw new MainError({ 
           exitCode: 1,
           message: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.`
--- a/packages/cli/src/run-test.ts
+++ b/packages/cli/src/run-test.ts
@@ -41,7 +41,6 @@ export function* runTest(config: ProjectOptions, formatter: StreamingFormatter):
       break;
     } else if(result.event) {
       let status = result.event.status;
-      console.log(result.event);
       if(result.event.type === 'testRun:result') {
         testRunStatus = result.event.status;
       }
--- a/packages/cli/src/run-test.ts
+++ b/packages/cli/src/run-test.ts
@@ -15,7 +15,7 @@ export function* runTest(config: ProjectOptions, formatter: StreamingFormatter):
     try {
       return yield Client.create(uri);
     } catch (e) {
-      if (e.name === 'ConnectionAttemptFailed') {
+      if (e.name === 'NoServerError') {
         throw new MainError({ 
           exitCode: 1,
           message: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.`
--- a/packages/cli/src/run-test.ts
+++ b/packages/cli/src/run-test.ts
@@ -9,20 +9,21 @@ import { StreamingFormatter } from './format-helpers';
 
 export function* runTest(config: ProjectOptions, formatter: StreamingFormatter): Operation<void> {
 
-  let client: Client;
   let uri = `ws://localhost:${config.port}`;
-
-  try {
-    client = yield Client.create(uri);
-  } catch (e) {
-    if (e.message.includes('websocket server closed connection unexpectedly')) {
-      throw new MainError({ 
-        exitCode: 1,
-        message: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.`
-      });
-    }
-    throw e;
-  }
+  
+  let client: Client = yield function*() {
+    try {
+      return yield Client.create(uri);
+    } catch (e) {
+      if (e.message.includes('websocket server closed connection unexpectedly')) {
+        throw new MainError({ 
+          exitCode: 1,
+          message: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.`
+        });
+      }
+      throw e;
+    }  
+  };
 
   let subscription = yield client.subscription(query.run());
 
--- a/packages/cli/src/run-test.ts
+++ b/packages/cli/src/run-test.ts
@@ -8,9 +8,24 @@ import * as query from './query';
 import { StreamingFormatter } from './format-helpers';
 
 export function* runTest(config: ProjectOptions, formatter: StreamingFormatter): Operation<void> {
-  let client: Client = yield Client.create(`ws://localhost:${config.port}`);
+
+  let client: Client;
+  let uri = `ws://localhost:${config.port}`;
+
+  try {
+    client = yield Client.create(uri);
+  } catch (e) {
+    if (e.message.includes('websocket server closed connection unexpectedly')) {
+      throw new MainError({ 
+        exitCode: 1,
+        message: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.`
+      });
+    }
+    throw e;
+  }
 
   let subscription = yield client.subscription(query.run());
+
   let stepCounts = { ok: 0, failed: 0, disregarded: 0 };
   let assertionCounts = { ok: 0, failed: 0, disregarded: 0 };
   let testRunStatus: ResultStatus | undefined;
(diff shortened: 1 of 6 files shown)


packages/cli/src/run-test.ts
```diff
@@ -8,9 +8,24 @@ import * as query from './query';
...
+    client = yield Client.create(uri);
+  } catch (e) {
+    if (e.message.includes('websocket server closed connection unexpectedly')) {
+      throw new MainError({ 
+        exitCode: 1,
+        message: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.`
+      });
+    }
+    throw e;
+  }
```
cowboyd: We can wrap the client creation into a single operation, that way no need to re-assign the `client` reference.

```suggestion
  let uri = `ws://localhost:${config.port}`;
  
  let client: Client = yield function*() {
    try {
      return yield Client.create(uri);
    } catch (e) {
      if (e.message.includes('websocket server closed connection unexpectedly')) {
        throw new MainError({ 
          exitCode: 1,
          message: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.`
        });
      }
      throw e;
    }  
  };
```

packages/cli/src/run-test.ts
```diff
@@ -8,9 +8,24 @@ import * as query from './query';
...
+  let client: Client;
+  let uri = `ws://localhost:${config.port}`;
+
+  try {
+    client = yield Client.create(uri);
+  } catch (e) {
+    if (e.message.includes('websocket server closed connection unexpectedly')) {
+      throw new MainError({ 
+        exitCode: 1,
+        message: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.`
```
cowboyd: Since we use the `runTest` command in both the `ci` and `test` commands, we have a potential weird case where the server stops responding to commands in between when it is started, and when we actually try to connect the client. This should be exceedingly rare, but I figured I'd point it out nonetheless.

packages/cli/src/run-test.ts
```diff
@@ -25,6 +40,7 @@ export function* runTest(config: ProjectOptions, formatter: StreamingFormatter):
       break;
     } else if(result.event) {
       let status = result.event.status;
+      console.log(result.event);
```
cowboyd: We should probably enable a linting rule for console.log statements.

packages/cli/src/run-test.ts
```diff
@@ -8,9 +8,24 @@ import * as query from './query';
...
 export function* runTest(config: ProjectOptions, formatter: StreamingFormatter): Operation<void> {
-  let client: Client = yield Client.create(`ws://localhost:${config.port}`);
+
+  let client: Client;
+  let uri = `ws://localhost:${config.port}`;
+
+  try {
+    client = yield Client.create(uri);
+  } catch (e) {
+    if (e.message.includes('websocket server closed connection unexpectedly')) {
```
cowboyd: We could also modify the client to throw a more informative error, so that we could ask `e.name === 'NoServerError` or something. That way it would be more robust for future users of `Client`. Not sure how that would work with TypeScript. Thoughts?

cowboyd (dismissed):
I'm wondering if we should have a timeout

packages/cli/src/run-test.ts
```diff
@@ -8,9 +8,25 @@ import * as query from './query';
...
+  let client: Client = yield function*() {
+    try {
+      return yield Client.create(uri);
+    } catch (e) {
+      if (e.message.includes('websocket server closed connection unexpectedly')) {
+        throw new MainError({ 
+          exitCode: 1,
+          message: `Could not connect to BigTest server on ${uri}. Run "bigtest server" to start the server.`
+        });
+      }
```
jnicklas: Can we raise a more specific error in the client in case this happens, so we don't have to rely on matching on the error message?

jnicklas (dismissed):

packages/cli/src/run-test.ts
taras: Done

packages/cli/src/run-test.ts
taras: Done

taras: Oops... nevermind, not done yet.

taras: Ok, done now

packages/client/src/client.ts
```diff
@@ -16,13 +17,25 @@ export class Client {
   static *create(url: string): Operation<Client> {
     let socket = new w3cwebsocket(url) as WebSocket;
 
+    yield spawn(function* detectStartupError(): Operation<void> {
+      let [error] = yield once(socket, 'error');
+      
+      if (isYaetiError(error)) {
```
cowboyd: I'd call this `isErrorEvent` since Yaeti is just the `nodejs` error event emulator for the [`ErrorEvent`](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent). When the client is running on the browser, it will actually be an `ErrorEvent`

cowboyd: Worth noting that on the web, this will _always_ be an error event.

packages/client/src/errors.ts
```diff
@@ -0,0 +1,3 @@
+export class ConnectionAttemptFailed extends Error {
```
pittst3r: Would it be more conventional to name this `ConnectionAttemptFailedError`?

packages/client/src/client.ts
```diff
@@ -16,13 +17,25 @@ export class Client {
   static *create(url: string): Operation<Client> {
     let socket = new w3cwebsocket(url) as WebSocket;
 
+    yield spawn(function* detectStartupError(): Operation<void> {
+      let [error] = yield once(socket, 'error');
+      
+      if (isErrorEvent(error)) {
+        throw new NoServerError(`Could not connect to server at ${url}`);
+      } else {
+        throw error;
```
cowboyd, line 26: I'm trying to imagine a scenario where we execute this code branch, and can't think of one other than a bug in the `websocket` library. Still, it's good to have in case there is and it will throw a big nasty stack trace.

packages/client/src/client.ts
```diff
@@ -16,13 +17,25 @@ export class Client {
   static *create(url: string): Operation<Client> {
     let socket = new w3cwebsocket(url) as WebSocket;
```
cowboyd, line 18: Out of scope for this PR, but in order to make this work in the browser, we're going to have to loosely couple to our underlying websocket library. I wonder what is the best way to do that.

cowboyd (approved):
Looks great! effection code looking sharp.

pittst3r (approved):

PR #474, opened 2020-08-21
Consume websockets on the server using @effection/subscription

cowboyd:
Motivation
-----------
In order to write tests for the bigtest agent, we need to simulate an orchestrator agent handler that the agent can connect to. However, the @bigtest/effection-express package requires you to write your websocket handler with the deprecated mailbox handler interface.

Approach
----------
Rather than replace the mailbox interface, this adds the subscription interface along side. That way, consumers of effection-express can migrate over gradually over time.

In order to verify both interfaces, tests have been wrapped around the websocket API for both the Mailbox interface as well as the subscription interface.

### DIFF
--- a/packages/effection-express/test/websocket.test.ts
+++ b/packages/effection-express/test/websocket.test.ts
@@ -11,7 +11,7 @@ import { Socket, express } from '../src';
 import { Mailbox } from '@bigtest/effection/dist';
 
 
-describe.only('websocket server', () => {
+describe('websocket server', () => {
   let client: WebSocket;
   let connection: Socket;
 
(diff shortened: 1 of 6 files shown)


taras (approved):
👍

packages/effection-express/test/websocket.test.ts
```diff
@@ -0,0 +1,85 @@
...
+import { Channel } from '@effection/channel';
+import { subscribe, ChainableSubscription } from '@effection/subscription';
+import * as WebSocket from 'ws';
+
+import { run } from './helpers';
+import { Socket, express } from '../src';
+import { Mailbox } from '@bigtest/effection/dist';
+
+
+describe.only('websocket server', () => {
```
taras: There is an `only` here

taras (changes requested):

pittst3r (commented):
LGTM after removing the `.only`.

taras (approved):
👍

PR #475, opened 2020-08-21
make Bundler a top level entity

dagda1:
This is a follow up to #450  makes bundler a top-level entity in the graphql schema:

```javascript
t.field("bundler", {
  type: "Bundler"
});
.....
enumType({
  name: "BundlerType",
  members: [
    'UNBUNDLED',
    'BUILDING',
    'GREEN',
    'ERRORED'
  ]
}),
objectType({
  name: "Bundler",
  definition(t) {
    t.field("type", {
      type: 'BundlerType'
    });
    t.string("path", {
      nullable: true
    }),
    t.list.field("errors", {
      type: "Error",
      nullable: true
    }),
    t.list.field("warnings", {
      type: "Error",
      nullable: true
    })
  }
}),
```

### DIFF
--- a/packages/server/src/schema.ts
+++ b/packages/server/src/schema.ts
@@ -6,6 +6,7 @@ import {
   subscriptionField,
   stringArg,
   makeSchema,
+  enumType,
 } from "@nexus/schema";
 
 export const schema = makeSchema({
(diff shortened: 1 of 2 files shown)


packages/server/src/schema.ts
```diff
@@ -168,6 +173,34 @@ export const schema = makeSchema({
         })
       }
     }),
+    enumType({
+      name: "BundlerType",
```
cowboyd: Can we call this bundler status?

dagda1: we can and have

cowboyd (approved):
Very nice indeed!

PR #476, opened 2020-08-21
Modernize agent

cowboyd:
Motivation
-----------

In preparation of isolating test code, we'll need to make some changes to the agent. However, there is actually quite a bit of dead code in the agent and harness that can be simplified or removed altogether in preparation for making any needed changes. 

Approach
----------

remove the dead code, ensure that tests still pass.


packages/agent/app/test-frame.ts
```diff
@@ -9,19 +8,14 @@ export class TestFrame {
...
 
-  load(url: string): Operation {
-    return ({ resume, ensure }) => {
-      this.element.src = url;
-      this.element.addEventListener('load', resume);
-      ensure(() => this.element.removeEventListener('load', resume));
-    }
+  *load(url: string) {
+    this.element.src = url;
+    yield once(this.element, 'load');
```
pittst3r, line 18: So much easier to understand. 💯

pittst3r (approved):

PR #477, opened 2020-08-22
Replace Mailbox-based ConnectionServer with AgentHandler

cowboyd:
> depends on #474 

Motivation
-----------

We want to make sure that our effection code represents an reasonably up-to-date snapshot of the best practices. Having decide to deprecate Mailboxes in favor of channels and subscriptions, we want to remove them where they appear.

They were used quite extensively in the agent test suite, and by proxy in the general purpose "Connection Server" which was meant to be a general server mechanism for handling agent connections. In practice, this class was never adopted by the `@bigtest/server`. However, with a new approach this can be more feasible.

Approach
----------

The general purpose server handler has been refactored into a resource called an `AgentHandler` which lets you subscribe to agents as they come in and handle their events accordingly. This uses modern techniques and provides an example of how to use resource and channels in coordination.

Because also there are now two channels associated with handling agents, one to accept incoming connections, and one to actually handle the events of the agent, there is the possibility that we don't need to pass `agentId` around all the time in the messages since it will be implicitly associated with a specific channel.


jnicklas (approved):
Looks great!

Issue #437, opened 2020-08-03
Step timeouts and interactor timeouts should be configurable via the command line and in bigtest.json


cowboyd:
We hardcode these to 2 seconds, but they shouldn't be. It's probably a good idea to make sure that the step timeouts are longer than interactor timeouts. That way you don't have the step shutting down the interaction before it's ready.

jnicklas:
Our step timeout is now *very* big at 60s. It seems unlikely that someone would have to increase this timeout. I think we should close this issue until we actually get some feedback on this, but at the very least we should remove the beta milestone.

jnicklas:
Removed the beta milestone from this.

PR #479, opened 2020-08-24
Increase step timeout

pittst3r:
## Motivation

Partially addresses https://github.com/thefrontside/bigtest/issues/437.

A hard-coded two seconds does not work universally. For example, some apps may be slow to boot and the test will time out before anything happens.

Additionally, we want the step timeout to be longer than the interactor timeout such that we prioritize the interactor failure message ahead of the step timeout message. This also enables multiple interactor invocations within a step.

## Approach

- Set the step timeout to one minute.
- Make the agent warn (per lane) if the interactor timeout is longer than the step timeout
- Make the step timeout a parameter of the `run` command coming from the server (originally motivated to make it testable but I think it makes it easier for future configurability)

## To-do

Consider a nicer API for managing timeouts.

### DIFF
--- a/.changeset/compute-step-timeout.md
+++ b/.changeset/compute-step-timeout.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/agent": minor
+---
+
+Set step timeout to 60s and throw agent warning if interactor timeout is >= step timeout.
--- a/packages/agent/app/run-lane.ts
+++ b/packages/agent/app/run-lane.ts
@@ -27,11 +27,17 @@ export function* runLane(config: LaneConfig) {
   }
 }
 
+const stepTimeout = 60_000;
+
 function *runLaneSegment(testRunId: string, events: TestEvents, test: TestImplementation, context: TestContext, remainingPath: string[], prefix: string[]): Operation<void> {
   let currentPath = prefix.concat(test.description);
 
   console.debug('[agent] running test', currentPath);
   events.send({ testRunId, type: 'test:running', path: currentPath })
+
+  if (bigtestGlobals.defaultInteractorTimeout >= stepTimeout) {
+    console.warn(`[agent] the interactor timeout should be less than, but is greater than or equal to, the step timeout of ${stepTimeout}`);
+  }
 
   for(let step of test.steps) {
     let stepPath = currentPath.concat(step.description);
@@ -39,7 +45,7 @@ function *runLaneSegment(testRunId: string, events: TestEvents, test: TestImplem
       console.debug('[agent] running step', step);
       events.send({ testRunId, type: 'step:running', path: stepPath });
 
-      let result: TestContext | void = yield timebox(step.action(context), 2000)
+      let result: TestContext | void = yield timebox(step.action(context), stepTimeout)
 
       if (result != null) {
         context = {...context, ...result};
@@ -77,7 +83,7 @@ function *runLaneSegment(testRunId: string, events: TestEvents, test: TestImplem
           console.debug('[agent] running assertion', assertion);
           events.send({ testRunId, type: 'assertion:running', path: assertionPath });
 
-          yield timebox(assertion.check(context), 2000)
+          yield timebox(assertion.check(context), stepTimeout)
 
           events.send({ testRunId, type: 'assertion:result', status: 'ok', path: assertionPath });
         } catch(error) {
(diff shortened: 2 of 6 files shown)


jnicklas:
Would it make sense to add a new global setting `defaultStepTimeout` instead? We could default this to 10s (though I could also see this being even higher), which would leave us with identical functionality, but it seems a bit more intuitive rather than adding this arbitrary multiplier.

pittst3r:
@jnicklas Would that mean people would potentially need to manage the two timeouts on their own if they want to deviate from the defaults?

cowboyd:
@jnicklas Unfortunately, this is a bit of a can of worms. First, we want to ensure that the the step timeout is always more than the interactor timeout so that you get "couldn't find element X within Y matching Z" rather than "step took too long". In fact, it would be even better if the step timeout would actually be dynamic depending on the interactions contained within the step. So if I had a step that looked like like:

```ts
step("login", async () => {
    await App.visit('/');
    await Input('username').focus();
    await Input('username', { hasFocus: true }).type(username);
    await Input('password', { hasFocus: true }).type(password);
    await Button('Log in').click();
});
```

It would know that it was comprised of 5 interactions, and should be something like 6x the interactor timeout. This doesn't even account for the idea that some interactions might take a long time like the visiting of the original URL.

My suspicion is that our user's intuition will not make an immediate distinction between interactor timeout and step timeout and even when they do, if they do, they won't understand what is a "good" value for a step timeout in relation to an interactor timeout and why your step timeout must be significantly longer than interactor timeout I agree that this is very blunt instrument (although slightly less blunt than what we have now which is 2s hard-coded timeouts for both), but given that you will almost always need to adjust your step timeout whenever you adjust your interactor timeout, it does seem that one is constrained by the other.... shouldn't we help our users by at least setting a sane default that avoids the pitfalls?

Would this be a potential solution that would make computing the step timeout from the interactor timeout more palatable?

1. Allow for explicitly setting both step timeout and interactor timeout
2. If step timeout is smaller than interactor, timeout, it is an ERROR
3. If step timeout is within a certain threshold, say < 2x of interactor timeout, it is a WARNING
4. If no step timeout is specified, then it is defaulted to 5x interactor timeout.


I predict the subject of optimal timeouts is going to be an entire research project spanning months, if not years, but what is the way we can get folks up and running most quickly that will give them the least flake, and yet always give them an escape hatch?

jnicklas:
A really simple solution to this is making the step timeout very big, so that we are unlikely to hit it under normal conditions. Something like 60s or even 120s maybe. The conditions for creating a step which times out on its own should be pretty unlikely if users end up using interactors mostly. This would mean that while there is an adjustable timeout for steps, the user should very rarely have to concern themselves with it.

pittst3r:
@jnicklas That seems sensible to me. What do you think @cowboyd? 60s seems good to me.

I'm trying to think of conditions that would cause a step to timeout with a high value like 60-120s without any interactor timeouts. I can't think of anything specific but it seems like maybe it would only be exceptional circumstances and would indicate something is not right.

cowboyd:
I'm ok with this. However, we still run into the issue of what happens when someone sets the interactor timeout to 60s...

pittst3r:
This is truly a can of worms.

> what happens when someone sets the interactor timeout to 60s

That smells too long to me in absolute terms. Maybe we should update the `defaultInteractorTimeout` setter to warn if it exceeds some value (or maybe update @bigtest/interactor instead). Maybe like 20s. Something is very wrong with your app if a single action legitimately takes that long. Here we are still hard-coding arbitrary values but until we can come up with a better pattern maybe this is okay? I mean, 2s still seems too long to me but we've already seen that some apps are incapable of making that deadline without lots of work.

cowboyd:
Do we have a consensus with this one? configurable interactor timeout, absolute, large step timeout, and warning if interactor timeout exceeds it?

jnicklas:
Sounds great to me!

pittst3r:
Sounds good, I will make these changes.

cowboyd:
@pittst3r heads up: In order to get past the new MacOS test run introduced with #486, you'll have to rebase onto origin/v0 and then close and re-open this PR to pull in the new checks.

pittst3r:
This is ready for another review.

cowboyd (approved):

taras (approved):
🚢

PR #481, opened 2020-08-25
Upgrade @effection/subscription

jnicklas:
Also fixes all warnings created through deprecation of `Subscribable.from`


cowboyd (dismissed):
love it!

jnicklas:
Didn't add a changeset, because there is no change in functionality. This only updates internal dependencies.

cowboyd:
> Didn't add a changeset, because there is no change in functionality. This only updates internal dependencies.

Ah, I see. I think that in this case that it should be a `patch` then. If we don't indicate that the PR includes a changeset, then it won't actually create a release for this package in the release PR.

I'd see changes that don't require a changeset as things like adding tests, that don't change any runtime operation.

cowboyd (approved):

PR #483, opened 2020-08-25
Better interactor errors

### LINKED ISSUE #455: When filter value matches, show what it was, not just what it wasn't
>  FilterNotMatchingError: Key "Permanent location" within Card "Item title:  A semantic web primer" does not match filters: with value "Main Library"

In this example, there was a double space after the `:` character in "Item title: A semantic web primer".

It would be _really_ awesome to see what the value of the filter was.

jnicklas:
Improve error messages when an interactor fails to match.

Examples:

```
did not find link "Blah", did you mean one of "Foo Bar", "Quox"?
```

```
did not find text field "Password" which is not enabled and with value "incorrect", did you mean one of:

| text field   | enabled: false | value: "incorrect"    |
| ------------ | -------------- | --------------------- |
| ✓ "Password" | ✓ false        | ⨯ "test1234"          |
| ⨯ "Email"    | ⨯ true         | ⨯ "jonas@example.com" |
```

Closes #455


cowboyd (approved):
This is so cool! I'd love to see this tabular data attached to the error as metadata so that we can work with it from GraphQL

PR #484, opened 2020-08-27
Resolve source maps in error stack traces

### LINKED ISSUE #431: Resolve sources in stack traces that are generated on remote JavaScript runtime
We generate sourcemaps for each test bundle which means that in the browser where an exception occurs, the sources are resolved. However, when the same exception is transmitted across the network, logging it out to the console does not resolve the sources because 

1. it isn't an `Error` class
2. the stack trace was generated by code on a remote JavaScript vm and so the sourcemap is not intrinsically associated with the code.

We need to first figure out how to do this period, possibly by manually loading the sourcemap, and and then figure out how to integrate it into our output.

### Example

```
 AssertionError: expected 2 == 3
|       at _callee$ (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:66693:23)
|       at tryCatch (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:196:23)
|       at Generator.invoke [as _invoke] (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:413:28)
|       at Generator.next (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:249:27)
|       at asyncGeneratorStep (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:861:28)
|       at _next (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:883:13)
|       at http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:890:11
|       at new Promise (<anonymous>)
|       at http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:879:16
|       at _assertEqual (http://localhost:24005/manifest-44d22a51a12a34996032ee8496870689be93fe591ea90818f31fe310be8dfdfc.js:6670
```

jnicklas:
Closes #431

Use the error-stack-parser and get-source packages to extract the source location from stack traces and transform these sources. This gives us vastly improved stack traces.

Also adds support for filtering the stack trace to hide lines from bigtest itself, as well as dependencies from node_modules. The resulting stack traces are very clean. In the future we can make these configurable via options to the CLI.

Example:

```
⨯ [step]       Failing Test -> child -> child second step
|    ERROR: moo
|      test/fixtures/failing.test.ts:14:54 @ _callee2$
```

### DIFF
--- a/packages/agent/app/lane.ts
+++ b/packages/agent/app/lane.ts
@@ -1,16 +1,9 @@
-import { timebox } from './timebox';
-
-import { Agent } from '../shared/agent';
-import { TestImplementation, ErrorDetails, Context as TestContext } from '@bigtest/suite';
 import { Operation, fork } from 'effection';
+import { TestImplementation, Context as TestContext } from '@bigtest/suite';
 
-const serializeError: (error: ErrorDetails) => ErrorDetails = ({ message, fileName, lineNumber, columnNumber, stack }) => ({
-  message,
-  fileName,
-  lineNumber,
-  columnNumber,
-  stack
-});
+import { timebox } from './timebox';
+import { serializeError } from './serialize-error';
+import { Agent } from '../shared/agent';
 
 export function *runLane(testRunId: string, agent: Agent, test: TestImplementation, path: string[]): Operation<void> {
   return yield runLaneSegment(testRunId, agent, test, {}, path.slice(1), []);
@@ -50,7 +43,7 @@ function *runLaneSegment(testRunId: string, agent: Agent, test: TestImplementati
           type: 'step:result',
           status: 'failed',
           timeout: false,
-          error: serializeError(error),
+          error: yield serializeError(error),
           path: stepPath
         });
       }
@@ -71,7 +64,7 @@ function *runLaneSegment(testRunId: string, agent: Agent, test: TestImplementati
           agent.send({ testRunId, type: 'assertion:result', status: 'ok', path: assertionPath });
         } catch(error) {
           console.error('[agent] assertion failed', assertion, error);
-          agent.send({ testRunId, type: 'assertion:result', status: 'failed', error: serializeError(error), path: assertionPath });
+          agent.send({ testRunId, type: 'assertion:result', status: 'failed', error: yield serializeError(error), path: assertionPath });
         }
       });
     }
--- a/packages/agent/app/serialize-error.ts
+++ b/packages/agent/app/serialize-error.ts
@@ -8,6 +8,7 @@ interface SourceLocation {
   name?: string;
   line?: number;
   column?: number;
+  sourceLine?: string;
   sourceFile: {
     path: string;
   };
@@ -24,6 +25,7 @@ export function *resolveStackFrames(stackFrames: StackFrame[]): Operation<ErrorS
         return {
           name: location.name || stackFrame.functionName,
           fileName: stackFrame.fileName,
+          code: location.sourceLine,
           line: stackFrame.lineNumber,
           column: stackFrame.columnNumber,
           source: {
--- a/packages/agent/app/serialize-error.ts
+++ b/packages/agent/app/serialize-error.ts
@@ -1,0 +1,40 @@
+import * as getSource from 'get-source';
+import * as ErrorStackParser from 'error-stack-parser';
+import { Operation } from 'effection';
+import { ErrorDetails } from '@bigtest/suite';
+
+interface SourceLocation {
+  name?: string;
+  line?: number;
+  column?: number;
+  sourceFile: {
+    path: string;
+  };
+}
+
+export function *serializeError(error: Error): Operation<ErrorDetails> {
+  let parsedError = yield Promise.all(ErrorStackParser.parse(error).map(async (stackFrame) => {
+    if(stackFrame.fileName && stackFrame.lineNumber && stackFrame.columnNumber) {
+      let source = await getSource.async(stackFrame.fileName);
+      let location: SourceLocation = await source.resolve({ line: stackFrame.lineNumber, column: stackFrame.columnNumber });
+      return {
+        name: location.name || stackFrame.functionName,
+        fileName: stackFrame.fileName,
+        line: stackFrame.lineNumber,
+        column: stackFrame.columnNumber,
+        source: {
+          fileName: location.sourceFile.path,
+          line: location.line,
+          column: location.column
+        }
+      };
+    } else {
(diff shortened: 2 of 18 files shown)


packages/agent/app/lane.ts
```diff
@@ -71,7 +64,7 @@ function *runLaneSegment(testRunId: string, agent: Agent, test: TestImplementati
           agent.send({ testRunId, type: 'assertion:result', status: 'ok', path: assertionPath });
         } catch(error) {
           console.error('[agent] assertion failed', assertion, error);
-          agent.send({ testRunId, type: 'assertion:result', status: 'failed', error: serializeError(error), path: assertionPath });
+          agent.send({ testRunId, type: 'assertion:result', status: 'failed', error: yield serializeError(error), path: assertionPath });
```
cowboyd, line 67: What do we do if an error occurs in error serialization? I think we need to make sure that this is a very safe thing, because as-is, it could blow up the entire test run.

packages/agent/app/serialize-error.ts
```diff
@@ -0,0 +1,40 @@
...
+interface SourceLocation {
+  name?: string;
+  line?: number;
+  column?: number;
+  sourceFile: {
+    path: string;
+  };
+}
+
+export function *serializeError(error: Error): Operation<ErrorDetails> {
```
cowboyd: Even though our type is `Error` here, I think we can enter this function with values that are not, in fact, errors because the default type of a catch block is `any`. While it should never happen that someone does something like `throw 73;` inside their test cases, I think we should abide by the heuristic: "if something can be thrown, it will be"

packages/server/src/manifest-server.ts
```diff
@@ -7,12 +7,17 @@ interface ManifestServerOptions {
...
+  proxyPort: number;
 };
 
 export function* createManifestServer(options: ManifestServerOptions): Operation {
   let app = express();
 
-  app.raw.use(staticMiddleware(options.dir));
+  app.raw.use(staticMiddleware(options.dir, {
+    setHeaders(res) {
+      res.setHeader('Access-Control-Allow-Origin', `http://localhost:${options.proxyPort}`);
```
cowboyd, line 18: Is this required so that the stacktrace resolver can fetch the sourcemap?

packages/cli/src/query.ts
```diff
@@ -1,8 +1,8 @@
-import { ResultStatus } from '@bigtest/suite';
+import { ResultStatus, ErrorDetails } from '@bigtest/suite';
 
 export function run() {
   return `
-    subscription {
+    subscription($showInternal: Boolean! = true, $showDependencies: Boolean! = true) {
```
cowboyd: while `showInternal` makes sense as the parameter to the `stack()` property on `ErrorDetails`, it threw me for a loop when I was reading it at the subscription level, can we call these variables `stacktraceShowInternal` and `stacktraceShowDependencies` ?

cowboyd (commented):
A few suggestions, but I love it how BigTest does the heavy lifting here to resolve stack traces and provide a convenient way of filtering them over GraphQL. This provides the best of both worlds, where building the best error message is easy, and working with complex stacks is possible.

packages/agent/app/lane.ts
jnicklas: Yes, you're right, we should be careful here!

packages/agent/app/serialize-error.ts
jnicklas: Probably true, but what is a good way of disambiguating here? Should we just check `typeof`?

packages/server/src/manifest-server.ts
jnicklas: Exactly.

taras (commented):
Does it make sense to remove the invocation function because it's not very helpful at the moment: `@ _callee2$`

jnicklas:
@taras yeah, I've been unable to get better information for the function name, but I still think that it could be useful sometimes, e.g. if the error is thrown in a helper method or something. Maybe we could make this nicer to read through using terminal colors, we could de-emphasize the function name, but still render it.

taras:
@jnicklas is it possible to get the actual code at that line?

packages/agent/app/serialize-error.ts
cowboyd: Maybe we could duck type on  a `stack` property or whatever is used by `ErrorStackParser.parse()`?

jnicklas:
@taras yes it is, `getSource` makes it pretty simple to get the code at the location.

packages/agent/app/serialize-error.ts
dagda1: You could have a type guard

```typescript
function isError(error: Error | any): error is Error {
  return error instanceof Error;
}
```

That is where the line below that was queried in one of my previous PRs comes in for `BundlerError`

```typescript
Object.setPrototypeOf(this, new.target.prototype);
```

All custom errors should do this IMO.  It should be added to `AmbiguousElementError` and friends

type guards also type narrow so you could  have it

`*serializeError(error: unknown)`

If the type guard returns true then the type is narrowed to the Error class

jnicklas: @dagda1 I think pretty much all the errors we use inherit from `error` via `class SomeError extends Error`, this should satisfy the `instanceof` check, no prototype trickery necessary.

dagda1: @jnicklas the reason I mention it is [this](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#example).

I've always believed it was necessary as according to the link, the prototype chain is broken and needs to be restored.

Definitely not saying you must do this

jnicklas:
@taras I added support for getting the source at the error location, but it is disabled by default. We should add some parameters to the CLI to show these, but I didn't want to bloat this PR up even further.

packages/cli/src/query.ts
```diff
@@ -11,10 +11,18 @@ export function run() {
...
+            fileName
+            code @include(if: $showStackTraceCode)
+            line
+            column
+            source {
+              fileName
+              line
+              column
+            }
+          }
```
taras, line 25: WOW

taras (approved):
👍

PR #485, opened 2020-08-27
Run all tests in a lane inside their own separate JavaScript context (iframe)

### LINKED ISSUE #422: Should we dispose of each test's JavaScript context before the next lane is run?
We currently have two JavaScript context's at play when running tests: The `Agent` JS context, and the `Harness` JS context. The `Agent` JS context is initialized once when the agent itself is connected to the orchestrator, whereas the `Harness` JS context is created once _per test_ by clearing the contents of the test iframe. The following psedo-html demonstrates this relationship.

```html
<html>
  <body>This is the Agent frame which is loaded once</body>
  <iframe src="app.src.html">This is the Harness Frame into which the app is loaded<iframe>
</html>
```

The harness JS context is only responsible for housing the application under test, whereas the agent context is the js context where the tests actually run. So when you write:

```ts
test("homepage")
  .step("load the app", App.visit("/")
  .step("click on a button", () => Button("menu").click())
```

The interactor code is evaluated in the _agent_ frame, but because the agent frame and harness frame are loaded from the same origin, JavaScript in the agent frame has permission to access the `documentElement` of the harness frame. Thus it uses the harness `documentElement` when evaluating interactors.

The rub is that the scope of the agent frame can encompass not only multiple lanes in a single test, but in the case of a long lived server process used during development, it can span many many test runs. This is a problem because it [violates the principle of isolation and reliability][1] which is one of the pillars of BigTest. Anything that any test cases writes to the global scope will persist across every other test case. That's bad. We don't want to depend on library code, or test code being "well-behaved" and ensuring to clean up after itself. A much safer strategy is to discard it altogether after all side-effects have been run, and then start fresh again from the top.

> Note: this isn't an abstract concern, we encountered this when starting services that track global state like [miragejs][2] and [msw][3]

One thing we could do is to introduce a test frame along side the harness frame. In pseudo html this would look like this:

```html
<html>
  <body>This is the Agent frame which is loaded once</body>
  <iframe src="test.src.html">This frame is responsible for evaluating the test code</iframe>
  <iframe src="app.src.html">This is the Harness Frame into which the app is loaded<iframe>
</html>
```

Then, before each test case, the agent frame would tell the "runner" frame to clear, load the test bundle, and run the lane. The runner frame would be passed the document element of the harness frame so that all of the interactors would target it.

Drawbacks:

1. Instead of loading the test bundle once before the test run inside the agent frame, we'd have to load it once per lane. This could add a performance overhead.
2. Setting breakpoints and doing inspections on the test suite would now have to happen on the evaluation frame, and not the agent frame, so it might become more awkward.

[1]: https://frontside.com/blog/2020-triple-threat-to-testing-part-2-reliability/#environment-and-preconditions
[2]: https://miragejs.org
[3]: https://mswjs.io

cowboyd:
> resolves #422 

Motivation
-----------

We currently have two JavaScript context's at play when running tests: The Agent JS context, and the Harness JS context. The Agent JS context is initialized once when the agent itself is connected to the orchestrator, whereas the Harness JS context is created once per test by clearing the contents of the test iframe. The following psedo-html demonstrates this relationship.

```ts
<html>
  <body>This is the Agent frame which is loaded once</body>
  <iframe src="app.src.html">This is the Harness Frame into which the app is loaded<iframe>
</html>
```
The harness JS context is only responsible for housing the application under test, whereas the agent context is the js context where the tests actually run. So when you write:

```js
test("homepage")
  .step("load the app", App.visit("/")
  .step("click on a button", () => Button("menu").click())
````

The interactor code is evaluated in the agent frame, but because the agent frame and harness frame are loaded from the same origin, JavaScript in the agent frame has permission to access the documentElement of the harness frame. Thus it uses the harness documentElement when evaluating interactors.

The rub is that the scope of the agent frame can encompass not only multiple lanes in a single test, but in the case of a long lived server process used during development, it can span many many test runs. This is a problem because it violates the principle of isolation and reliability which is one of the pillars of BigTest. Anything that any test cases writes to the global scope will persist across every other test case. That's bad. We don't want to depend on library code, or test code being "well-behaved" and ensuring to clean up after itself. A much safer strategy is to discard it altogether after all side-effects have been run, and then start fresh again from the top.

Approach
----------

This Draft PR spikes out running the actual test code for each lane inside its own completely separate iframe. This way, the agent never even loads the test bundle at all. It means that if there are any global variables or side effects inside of a test bundle or 3rd party code, they'll never actually make their way into the agent frame.

I'd originally spiked this out using `MessageChannel` objects to user inter-frame communication, but found that it introduced large amounts of latency between when an event was actually raised, and when it was transmitted back to the orchestrator over websockets. This is probably to do with the fact that the agent frame, not having any interactive elements, did not get many cycles from the browser to flush its queues.

Ultimately, I used a brute-force approach of passing a direct reference to the lane parameters (path, manifestUrl, appURL, and event reporter) from the agent frame to the test frame which we can do because it's running on the same origin. This ended up being pretty fast when running through the CLI. Anecdotally I didn't notice any difference when running via the CLI, however also anecdotally, it did seem like the agent tests took a few more milliseconds on average to run. Still, this feels much safer in the long run since we now don't have to worry about side-effects inside our tests.

### CODE REVIEW

File: packages/agent/app/lane.ts [hunk 1]
```diff
@@ -12,40 +11,44 @@ const serializeError: (error: ErrorDetails) => ErrorDetails = ({ message, fileNa
   stack
 });
 
-export function *runLane(testRunId: string, agent: Agent, test: TestImplementation, path: string[]): Operation<void> {
+interface TestEvents {
+  send(event: TestEvent): void;
+}
+
+export function *runLane(testRunId: string, agent: TestEvents, test: TestImplementation, path: string[]): Operation<void> {
```
pittst3r: Should `agent` be renamed `events`?

```suggestion
export function *runLane(testRunId: string, events: TestEvents, test: TestImplementation, path: string[]): Operation<void> {
```

packages/agent/app/runner.ts
```diff
@@ -0,0 +1,46 @@
+import { Operation } from 'effection';
+import { Test } from '@bigtest/suite';
+
+import { Agent, Run, TestEvent } from '../shared/agent';
```
pittst3r: Looks like `Run` and `TestEvent` are in `../shared/protocol`?

pittst3r (commented):
Looks great, just a couple questions. Also, needs a changeset.

packages/agent/app/test-frame.ts
```diff
@@ -1,30 +1,48 @@
...
+    if (!element) {
+      throw new Error(`CRITICAL: unabled to find the test-frame! It should be findable with id: ${id}`);
+    }
+    if (!(element instanceof HTMLIFrameElement)) {
+      throw new Error(`CRITICAL: expected the test frame to be an <iframe>, but it was ${element}`);
+    }
 
-    return new TestFrame(element);
+    return new TestFrame(element, src);
   }
```
jnicklas: Does this even need to be a class? Can we just have `element` be a getter which throws an error if the frame is absent?

packages/agent/app/runner.ts
```diff
@@ -0,0 +1,46 @@
...
+import { TestFrame } from './test-frame';
+
+export interface Runner {
+  run(command: Run): Operation<void>;
+}
+
+export function* createRunner(agent: Agent): Operation<Runner> {
+
+  let testFrame: TestFrame = TestFrame.setup('test-frame', 'test-frame.html');
```
jnicklas: If testFrame is not a class then this wouldn't retain anything, meaning we wouldn't need the runner.

cowboyd:
🤔 It looks like the checks that get run are still the same even after rebasing. I may have to close this PR and open a new one.

jnicklas (approved):
Looks great!

Issue #432, opened 2020-08-03
Run Tests on Windows


cowboyd:
Bigtest does not currently work on Windows because of path separator and module separator issues. We should start running our test suite on windows.

PR #486, opened 2020-08-28
chore: test on multiple os

jbolda:
This updates the test script, bumps the version of a bunch of the official actions and prepares to run on ubuntu, mac and windows. There is also a step to cache the node_modules, but this will generally not trigger in PRs (as each PR has it's own cache) unless there a multiple commits. I switched to the node_module cache as that is more consistent between OS. It also sets a timeout to catch situations where the build hangs (like what is currently happening on windows).

This is the first step to address #432. We can enable windows once they pass in CI to keep the builds green.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,7 +59,7 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
       - uses: actions/download-artifact@v2
         with:
-          name: ${{ matrix.package }}/bigtest.dist.${{ matrix.platform }}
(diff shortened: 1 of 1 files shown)


jbolda:
I believe you may need to adjust the settings slightly that the old jobs are not required anymore?

cowboyd:
@jbolda Do we need to remove the old required checks?

jbolda:
I think you may need to for now? The names changed because it's appending the OS to the names now.

cowboyd (approved):

pittst3r (approved):

PR #489, opened 2020-08-31
Better error message when matches are ambiguous

### LINKED ISSUE #454: Better messaging for AmbiguousElementError
When there are more than one element on the page that an interaction matches, we should somehow at least give an indication of which elements are present that are causing the ambiguity. Otherwise, you're left to resolve it yourself.

jnicklas:
When a match in an interactor is ambiguous, we show a list of elements that matched by formatting their opening tag as suggested by @cowboyd in #454.

Closes #454

### DIFF
--- a/packages/interactor/src/match.ts
+++ b/packages/interactor/src/match.ts
@@ -26,6 +26,14 @@ export class Match<E extends Element, S extends InteractorSpecification<E>> {
   get sortWeight(): number {
     return this.matchLocator.sortWeight + this.matchFilter.sortWeight;
   }
+
+  elementDescription(): string {
+    let tag = this.element.tagName.toLowerCase();
+    let attrs = Array.from(this.element.attributes).map((attr) => {
+      return `${attr.name}="${attr.value}"`
+    });
+    return `<${[tag, ...attrs].join(' ')}>`;
+  }
 }
 
 export class MatchLocator<E extends Element> {
--- a/packages/interactor/src/match.ts
+++ b/packages/interactor/src/match.ts
@@ -1,6 +1,7 @@
 import { Locator } from './locator';
 import { Filter } from './filter';
 import { InteractorSpecification } from './specification';
+import { escapeHtml } from './escape-html';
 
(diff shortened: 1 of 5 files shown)


packages/interactor/src/match.ts
```diff
@@ -26,6 +26,14 @@ export class Match<E extends Element, S extends InteractorSpecification<E>> {
...
   }
+
+  elementDescription(): string {
+    let tag = this.element.tagName.toLowerCase();
+    let attrs = Array.from(this.element.attributes).map((attr) => {
+      return `${attr.name}="${attr.value}"`
+    });
+    return `<${[tag, ...attrs].join(' ')}>`;
+  }
 }
```
jnicklas: This is a bit of a brute-force implementation. Is there a better alternative?

cowboyd: Not that I can find 🤷

cowboyd (approved):
Looks good 👍

packages/interactor/src/match.ts
```diff
@@ -26,6 +26,14 @@ export class Match<E extends Element, S extends InteractorSpecification<E>> {
   get sortWeight(): number {
     return this.matchLocator.sortWeight + this.matchFilter.sortWeight;
   }
+
+  elementDescription(): string {
+    let tag = this.element.tagName.toLowerCase();
+    let attrs = Array.from(this.element.attributes).map((attr) => {
+      return `${attr.name}="${attr.value}"`
```
cowboyd: @jnicklas Just occured to me. Do we need to HTMLEncode the attribute value here? It's purely for debug purposes, but it would look weird if the `.value` had a `"` character for example

cowboyd:
@jnicklas heads up: In order to get past the new MacOS test run introduced with #486, you'll have to rebase onto `origin/v0` and then close and re-open this PR to pull in the new checks.

packages/interactor/src/match.ts
jnicklas: Nice catch!

PR #495, opened 2020-09-02
Use `match` instead of `filter` in bundler tests

jnicklas:
Because we can...


cowboyd (approved):
⚡ True ultimate power ⚡

PR #496, opened 2020-09-02
chore(test-workflow): switch cache to v2

jbolda:
This enables just the cache change per #492.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,4 +50,10 @@ jobs:
         run: yarn
       - name: Run tests
         run: yarn workspace @bigtest/${{ matrix.package }} test
+
+  fail:
+    runs-on: ubuntu-18.04
+    name: ${{ matrix.package }}
+    needs: test
+    steps:
       - run: boop to fail
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
         # which does not effect the node_modules themselves
         with:
           path: '**/node_modules'
-          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+          key: cachekey-v1-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
       - name: Install dependencies
         run: yarn
       - name: Run prepack
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,10 +50,3 @@ jobs:
         run: yarn
       - name: Run tests
         run: yarn workspace @bigtest/${{ matrix.package }} test
-
-  fail:
-    runs-on: ubuntu-18.04
-    name: ${{ matrix.package }}
-    needs: test
-    steps:
-      - run: boop to fail
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,14 +11,10 @@ jobs:
     name: prepack
     steps:
       - uses: actions/checkout@v1
-      - name: cache node_modules
+      - name: cache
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -8,13 +8,18 @@ on:
...
+      - name: cache node_modules
+        uses: actions/cache@v2
+        # cache node_modules based on lockfile which can change
+        # when there is no explicit change in package.json
+        # and prevents cache bust when scripts or metadata is changed
+        # which does not effect the node_modules themselves
         with:
-          path: /home/runner/.cache/yarn/v6
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          path: '**/node_modules'
```
cowboyd: Shouldn't we have _both_ the `.cache` dir _and_ `node_modules`?

.github/workflows/test.yml
```diff
@@ -8,13 +8,18 @@ on:
...
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
```
cowboyd: I like to have a key version as well which lets us just arbitrarily blow away the cache by swapping out keys in the event that we want to change our caching strategy without actually changing our `yarn.lock`, e.g.

```suggestion
          key: cachekey-v1-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
```

.github/workflows/test.yml
jbolda: does `.cache` provide anything extra from `node_modules`? I was hesitant to add more as there is a size limit to the overall cache and dropping too much in the cache will start to bust things unintentionally. As I say this though, there is a random timeout so maybe the cache is the source of this?

.github/workflows/test.yml
jbolda: entirely reasonable ❤️

.github/workflows/test.yml
cowboyd: @jbolda If I understand correctly, the `.cache` directory contains the downloaded tarballs, so in the event that you need to re-install a package, it doesn't have to download it. Actually, you're right. If we want to benefit fom this, we'll need to store it in a _separate_ cache, not tied to `yarn.lock`, basically one with a manually maintained cache key of `yarn-cache-v1` or something. That way, in the event that we do need to re-run a yarn install, we can skip the downloading step.

Maybe a separate PR though?

cowboyd:
@jbolda Interesting. It looks like caching might be the cause of the intermittent failures? 😢

cowboyd:
I wonder what happens if we even further limit our free variables by using the v1 cache action with `node_modules` and the `v2` cache action with `.cache`.

jbolda:
I have been getting similar errors in windows which makes me think `node_modules` myself. I wonder if it's worth just trying v2 with `.cache`. I can add a later command that fails, and that cheeky fail will let us rerun it a handful of times.

jbolda:
Looks like I can't rerun on my end, but that was successful so far.

cowboyd:
> Looks like I can't rerun on my end, but that was successful so far.
I re-ran four times in a row without failure which seems pretty likely then that this is the problem.

jbolda:
yeesh, maybe we skip that for now and merge everything else in over a few PRs and come back around to investigate?

cowboyd:
Sounds like a plan!

cowboyd:
Sounds like a plan!

cowboyd (approved):
Nice! I found out yesterday that our yarn.lock was out of sync. Wonder if that has anything to do wiith the failures we were seeing.

PR #497, opened 2020-09-02
Improve agent test build

jnicklas:
Previously we've taken an approach of relying on the fact that `prepack` is run before the tests are run, in combination with a `manifest:build` script, to build the fake manifest for the agent tests. This is a bit brittle, and it is also problematic in that source maps are not resolved correctly. I wanted to add some tests for source map resolution to the agent package itself, and this turned out to necessitate these changes. The new approach is to have a `pretest` script which runs before the tests and packages both the app and the fake manifest into the same hidden folder. This should be less surprising and it also allows us to test the source mapping properly.


jnicklas:
The `pretest` step takes a few seconds the first time around, but it is reasonably fast if the cache is primed. Less than 2s on my machine.

cowboyd:
TIL [`package.json#pretest`](https://docs.npmjs.com/misc/scripts) is a thing. 👍

cowboyd (approved):

packages/agent/package.json
```diff
@@ -16,8 +16,12 @@
     "test": "mocha -r ts-node/register test/**/*.test.ts",
     "mocha": "mocha -r ts-node/register",
     "start": "parcel serve --out-dir dist/app app/index.html app/harness.ts",
-    "prepack": "tsc --outdir dist --declaration --sourcemap && parcel build --public-url ./ --no-minify --out-dir dist/app app/index.html app/test-frame.html app/harness.ts",
-    "manifest:build": "parcel build test/fixtures/manifest.src.js --out-dir test/fixtures --out-file manifest.js --no-minify --global __bigtestManifest"
+    "prepack:app": "parcel build --public-url ./ --no-minify --out-dir dist/app app/index.html app/test-frame.html app/harness.ts",
+    "prepack:tsc": "tsc --outdir dist --declaration --sourcemap",
+    "prepack": "yarn prepack:app && yarn prepack:tsc",
```
cowboyd, line 21: There is one danger of making the prepack script use yarn, and that is if someone, using `npm` or `pnpm` or some other package were to want to install this package from a `git` or a `github` url, it would run prepack after installation and potentially blow up. I can't see that as a likely scenario, but wanted to point it out.

jnicklas:
> TIL [`package.json#pretest`](https://docs.npmjs.com/misc/scripts) is a thing. 👍

I didn't know that either, and I initially was very confused why it ran twice, because I'd added it to the `test` script 😂

packages/agent/package.json
jnicklas: Is there a better way of invoking one script from another? We don't *have* to do that, but it's convenient/

cowboyd: I've never found a good way, and I don't think this should block it. Just pointing it out in case it bites anybody in the future

PR #498, opened 2020-09-02
pass unwrapped bundle from manifest-builder to bundler

dagda1:
Currently, the `manifest-builder` passes an array of one element when creating the bundler.

```typescript
  let bundler: Bundler = yield Bundler.create(
    [{
      entry: options.srcPath,
      globalName: bigtestGlobals.manifestProperty,
      outFile: path.join(options.buildDir, "manifest.js")
    }],
  );
```

We can simplify the code by just dealing with the unwrapped object.

### DIFF
--- a/.changeset/one-bundle.md
+++ b/.changeset/one-bundle.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/bundler": minor
+---
+
+pass unwrapped bundle from manifest-builder to bundler
--- a/.changeset/one-bundle-to-bundler.md
+++ b/.changeset/one-bundle-to-bundler.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/bundler": minor
+---
+
+pass bundle and not array of bundles from manifest-builder to bundler
--- a/packages/bundler/src/bundler.ts
+++ b/packages/bundler/src/bundler.ts
@@ -21,51 +21,49 @@ interface BundlerOptions {
   mainFields: Array<"browser" | "main" | "module">;
 };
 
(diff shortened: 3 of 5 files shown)


.changeset/one-bundle.md
```diff
@@ -0,0 +1,6 @@
+---
+"@bigtest/bundler": patch
```
cowboyd: Since this is a breaking change, I'd make it a minor. This will actually automatically cause server to both bump a patch, and have it update its dependencies.

dagda1: very true.  Done

.changeset/one-bundle.md
```diff
@@ -1,6 +1,6 @@
 ---
-"@bigtest/bundler": patch
-"@bigtest/server": patch
+"@bigtest/bundler": minor
+"@bigtest/server": minor
```
cowboyd: Because "@bigtest/server" depends on "@bigtest/bundler" it will bump the patch version and add the changelog without even being part of the changeset.

cowboyd (approved):

pittst3r (approved):

.changeset/one-bundle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigtest/bundler": minor
```
pittst3r, line 2: Try this

```suggestion
"@bigtest/bundler": "minor"
```

pittst3r (changes requested):

.changeset/one-bundle.md
pittst3r: nm

pittst3r (approved):

PR #501, opened 2020-09-03
Migrate client from Mailbox -> Subscription

cowboyd:
Motivation
-----------
As part of moving towards a general reporter api, we need to incorporate formally the concept of "done-ness" to the stream of events coming from the server. That way, an entire test run can be represented as a single effection subscription.

The first piece in getting this to work is to make the client speak natively in subscriptions to begin with.

Approach
----------
This has been something that we've been meaning to do anyhow since it makes the code a lot nicer, and so this takes that first step by converting the client over to use subscriptions natively.

Instead of returning a mailbox from the `subscription` method, a subscription is returned instead. As an added bonus, the client methods are now parameterized, so that you can map, filter, match etc on a client query in a strongly typed fashion.

### DIFF
--- a/.changeset/subscription-based-client.md
+++ b/.changeset/subscription-based-client.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/client": minor
+---
+update Client.subscription() and Client.liveQuery() methods to return
+an effection `Subscription` instead of a Mailbox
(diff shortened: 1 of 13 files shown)


jnicklas (approved):
Awesome work! Let's work to get rid of Mailbox entirely! 🙌

PR #502, opened 2020-09-04
test file default export validation

dagda1:
This is my proposed way to validate that a test file has a default export of type

`export default {`
`module.exports =`
or
`exports =`

I have created an [eslint package](./packages/eslint-plugin-bigtest).  There currently only 1 rule in the package.

In `manifest-generator`, a `validator` object runs the [eslint CLIEngine](https://eslint.org/docs/1.0.0/developer-guide/nodejs-api) every time the chokidar watcher is triggered.

If an error is found, the file is not added to the `manifest.children` property but is instead added to a new errors property.

I am not failing the build but that is maybe something I should add.


packages/server/src/logger.ts
```diff
@@ -7,7 +7,7 @@ export interface LoggerOptions {
   out: <A extends unknown[]>(...args: A) => void;
 }
 
-export function* createLogger({ atom, out }: LoggerOptions) {
+export function* createBundlerLogger({ atom, out }: LoggerOptions) {
```
cowboyd, line 14: Let's keep this as-is. I actually want to move more things in here than just the bundler output so that we can start a server in our `ci` command without printing that stuff to the console.

packages/bundler/src/types.ts
```diff
@@ -3,8 +3,29 @@ import { RollupWarning, RollupError } from 'rollup';
 export type BundlerError = RollupError;
 export type BundlerWarning = RollupWarning;
 
+// TODO: All errors could implement at least these fields?
+export type ValidationException = {
```
cowboyd, line 6: Does this assume that all validators are source-based? For example, what if we determine that the test tree is too deep, something that should be done at runtime. There we won't have anything related to the source

packages/server/package.json
```diff
@@ -35,6 +36,8 @@
     "@types/rimraf": "3.0.0",
     "@types/websocket": "^1.0.0",
     "abort-controller": "^3.0.0",
+    "eslint": "6.8.0",
+    "eslint-plugin-bigtest": "^0.1.0",
```
cowboyd: Shouldn't these be runtime dependencies? How will the arrive onto the end-users system?

packages/eslint-plugin-bigtest/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "eslint-plugin-bigtest",
```
cowboyd: We should stick with the `@bigtest/eslint-plugin` convention

packages/server/src/validators/eslint-validator.ts
```diff
@@ -0,0 +1,85 @@
...
+  baseConfig: {
+    root: true,
+    plugins: ['bigtest'],
+    rules: {
+      'bigtest/require-default-export': 'error',
+    },
+  },
+  extensions: ['.ts', '.tsx', '.js', '.jsx'],
+};
+
```
cowboyd: What if we externalize this configuration into a `.babelrc` in the root of the `test/` directory. We can use our init script to deposit that boilerplate on installation. 

That way, users can tweak it to their hearts content, and really all we're doing is running eslint

packages/eslint-plugin-bigtest/src/rules/require-default-export.ts
```diff
@@ -0,0 +1,86 @@
...
+  AST_NODE_TYPES
+} from '@typescript-eslint/experimental-utils';
+import { createRule } from './utils';
+
+export const requireDefaultTextExport = createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Possible Errors',
+      description: 'Require a bigtest test to have only 1 default export',
```
cowboyd:
```suggestion
      description: 'Require a BigTest test file to have a default export',
```

packages/server/src/manifest-generator.ts
```diff
@@ -5,59 +5,86 @@ import { throwOnErrorEvent } from '@effection/events';
...
 interface ManifestGeneratorOptions {
   delegate: Mailbox;
   files: string[]; 
   destinationPath: string;
+  atom: Atom<OrchestratorState>;
 };
 
-function* writeManifest(options: ManifestGeneratorOptions) {
-  let files = yield globby(options.files);
+function* writeManifest({ bundlerSlice, destinationPath, ...options }: ManifestGeneratorOptions & { validator: Validator; bundlerSlice: Slice<BundlerState, OrchestratorState> }) {
```
cowboyd: Can we just add these to the ManifestGeneratorOptions and slice off the bundler state inside this method?

packages/server/src/manifest-generator.ts
```diff
@@ -5,59 +5,86 @@ import { throwOnErrorEvent } from '@effection/events';
...
 
-  let manifest = 'let load = (res) => res.default || res;\n';
-  manifest += 'const children = [\n';
+  let files: string[] = yield globby(options.files);
 
-  for(let file of files) {
-    let filePath = "./" + path.relative(path.dirname(options.destinationPath), file);
-    manifest += `  Object.assign({}, load(require(${JSON.stringify(filePath)})), { path: ${JSON.stringify(file)} }),\n`;
-  }
+  let validState = options.validator.validate(options.files);
```
cowboyd: Should validation be an `Operation` since it could be asynchronous? If not now, then in the future. As we add validators (typescript for example) then we can compose them into a single validation that runs in parallel with the others.

[1]: https://github.com/thefrontside/bigtest/issues/279

cowboyd (commented):
I must admit I hadn't thought of using eslint, but I like this approach since it will be automatically integrate with the user's IDE.


I'm curious. What is the state diagram of the bundle now?

```
UNBUNDLED -> VALIDATING --> VALID -> BUNDLING -> GREEN
                                                  |
                                                  +-> INVALID
```

Also, If we have the validation in-place, and represented in the `BundlerState`, do we now need to have the `errors: []` property in the manifest?

packages/server/src/logger.ts
dagda1: this was always a really quick fix as there was nothing.  

The logging is a ticket all on its own

packages/bundler/src/types.ts
dagda1: I mean any bundling/building errors, next we will have type checking is added and any other build-time errors

packages/server/src/logger.ts
dagda1: As far as the state goes

The validation happens outside of the bundling in the manifest-generator which is like a preprocessor.  I did think about putting it in rollup but I think it would complicate things with little benefit.  The bundler works on the single entry point of `manifest.js` and you would need a complicated plugin to make it work for multiple entry points or test files.

So the state goes from `'GREEN'` back to `'VALIDATING'` when the `chokidar` watcher gets triggered.

Be great to have a finite state machine around this.  Might give an extra layer of structure to the code also.

packages/server/src/validators/eslint-validator.ts
dagda1: I think that was the case I was trying to avoid because this is only validating the things that will stop bigtest working and not jumbled up with the user's wide range of formatting and whatever else they may have.

If we start inheriting rules from userland then it becomes more complicated.

Another way of doing this is to use [acorn](https://github.com/acornjs/acorn) or [estree-walker](https://github.com/Rich-Harris/estree-walker) which are lightweight parsers.

But I thought it would be good to kill the eslint rules and the file format check at the same time.

I think you can do the max depth rule with an AST also either eslint or one of the parsers above.

packages/server/src/manifest-generator.ts
dagda1: now, this is a good question.  Does type checking lie here or in rollup.  My initial thoughts were rollup.  Writing your own typescript checker that is performant is a lot of code and it is pretty much solved by one of the many rollup plugins.  I think I want to try that approach first.  The rollup plugins do a lot of caching etc and I don't think we want to go down that road just yet.

I did not want to go overboard on validators for now as there is no requirement for multiple validators yet and I might be over-engineering for something that might never happen.

If there are extra validators then absolutely they should be operations and parallelised.

packages/server/package.json
dagda1: definitely.  Moved

packages/eslint-plugin-bigtest/package.json
dagda1: renamed

packages/server/src/manifest-generator.ts
dagda1: `validator` and `bundlerSlice` are only used for `writeManifest`.  The `ManifestGeneratorOptions` is used in orchestrator so I created a new type.

packages/server/src/manifest-generator.ts
dagda1: having thought about it, I think you are right.  I've had a go.  Might be wrong to do this with a 1 element array.

cowboyd (changes requested):
@dagda1 Can we break this down into smaller component PRS. I think that will make it easier to review how all these pieces fit together. For example, I think having the ESLint plugin as it's own completely self-contained PR would be hugely helpful, and then maybe anotehr PR to introduce the bundle validation architecture that maybe does some lightweight checks first. With those in place, it will make it easier to see how and if to integrate heavier checks like typescript compilation and eslint running.

dagda1:
> @dagda1 Can we break this down into smaller component PRS.

@cowboyd no problem, I will do that

dagda1:
going to close this and split it up.

#512 has already been merged with at least 1 more to follow

PR #503, opened 2020-09-04
fix: windows prepack bugs and path issues

### LINKED ISSUE #434: test bundle entry points are incorrectly generated on windows
Inside the manifest generator, we find all of the test files, and then require all of them into a single entry point which we then point our bundler at. On unix systems, [this code][1] works because the separator for node modules, `'/'` is the same as the file separator, `'/'` So when we use the operating system path generator to generate the name of the test module on node, we get something like `"./../../my-case.test.js"` which is a valid unix path _and_ a valid node module.

However, when we use this logic to generate the test module name on Windows, it produces `"./..\\..\\..\\my-case.test.js"` which is a valid Windows path, but _not_ a valid node module name.

Instead, we need to always generate a correct node module name from the filesystem regardless of whether it is on Windows or Unix.

[1]: https://github.com/thefrontside/bigtest/blob/45ec886b51fbad1ccf153d4a80abfb342a24c41a/packages/server/src/manifest-generator.ts#L25-L26

jbolda:
This fixes windows related bugs in the prepack step for all packages. It also addresses path issues and other configuration issues that prevent `yarn test` from running. It _does_ not address `child_process` which I thought made sense to open in another PR (less surface area to review).

This closes #434.


package.json
```diff
@@ -22,7 +22,6 @@
       "packages/project",
       "packages/bundler",
       "packages/todomvc",
-      "packages/convergence",
```
jbolda: This package doesn't exist anymore?

packages/cli/package.json
```diff
@@ -13,7 +13,7 @@
   ],
   "scripts": {
     "lint": "eslint '{src,test,bin}/**/*.ts'",
-    "test": "mocha -r ts-node/register 'test/{,!(fixtures)/**}/*.test.ts'",
+    "test": "mocha -r ts-node/register \"test/{,!(fixtures)/**}/*.test.ts\"",
```
jbolda, line 16: Single quotes do not work on windows so we need escaped double quotes.

packages/server/package.json
```diff
@@ -15,7 +15,7 @@
     "test": "mocha ./test/setup test/**/*.test.ts",
     "mocha": "mocha ./test/setup",
     "test:app:start": "ts-node ./bin/todomvc.ts",
-    "generate:schema": "BIGTEST_GENERATE_SCHEMA=true ts-node ./src/schema.ts",
+    "generate:schema": "env BIGTEST_GENERATE_SCHEMA=true ts-node ./src/schema.ts",
```
jbolda, line 18: Prepended `env` in a node script is afaik the most cross compatible way to inline an env var.

packages/todomvc/package.json
```diff
@@ -16,7 +16,7 @@
     "lint": "eslint 'src/**/*.ts' 'test/**/*.ts'",
     "test": "mocha -r ts-node/register test/**/*.test.ts",
     "build": "yarn prepack",
-    "prepack": "rm -rf dist && mkdir -p dist && cd app && cp app.package.json package.json && react-scripts build && cd .. && mv app/build dist/app && tsc --outdir dist --module commonjs --declaration --sourcemap && chmod a+x dist/start.js"
+    "prepack": "rm -rf dist && yarn mkdirp dist -p && cd app && cp app.package.json package.json && react-scripts build && cd .. && mv app/build dist/app && tsc --outdir dist --module commonjs --declaration --sourcemap && chmod a+x dist/start.js"
```
jbolda, line 19: Windows misbehaves with the `-p` and creates a folder for `-p` instead. So using the `mkdirp` node script to help us out here.

.github/workflows/test.yml
```diff
@@ -31,6 +31,7 @@ jobs:
     name: ${{ matrix.package }}
     needs: prepack
     strategy:
+      fail-fast: false
```
cowboyd: This is something we want, but isn't really part of this PR scope.

package.json
cowboyd: I don't think so no. We may me keeping it around so that we can eventually deprecate it with the release of 1.0? /cc @minkimcello In any event, let's address it separately.

cowboyd (commented):
Nice... not only will we be able to run on windows, but we can develop on it too!

.github/workflows/test.yml
jbolda: Aye, I dropped it in there for testing on my fork and forgot to delete it 🤦‍♂️

jbolda:
@cowboyd cheers, all set!

cowboyd (approved):

PR #504, opened 2020-09-04
Allow DSL to consume units of work (like interactor) directly

cowboyd:
> This is another stab at #311 

If we use interactors with our DSL we end up repeating a lot of information:

```ts
.step("I click the 'next' link", async () => {
  await Link('next').click();
});
```

It would be amazing if we could simply infer this information, and write this more succinctly like this:

```ts
.step(Link('next').click())
```

To accomplish this, we do the simplest thing that could possibly work: we first allow the DSL to consume the units it represents directly, and then second, make interactors compatible with those unit interfaces.

For example, the `step()` function now has an overload to pass description and action as a single object:

```ts
.step({
  description: "this is a step provided as a single object",
  action: () => Promise.resolve({ hello: 'world' })
})
```

Now, we can simply make our interactions implement that interface and we're off to the races.

As an added bonus, interactions are now split into `Interaction` and `ReadonlyInteraction` so that checks like `exists()` or `absent()` canbe used inside `steps`, but actions like `click()` and `fillIn()` _cannot_ be used inside assertions. In typescript this will fail to compile, and in JS it will throw a runtime error.

### DIFF
--- a/packages/suite/src/dsl.ts
+++ b/packages/suite/src/dsl.ts
@@ -40,7 +40,7 @@ export class TestBuilder<C extends Context> implements TestImplementation {
   step<R extends Context | void>(descriptionOrStep: StepDefinition<C,R> | string, action?: Action<C,R>): TestBuilder<R extends void ? C : C & R> {
     let step = typeof descriptionOrStep !== 'string' ? descriptionOrStep : {
       description: descriptionOrStep,
-      action: action ? action : async () => {}
+      action: action ? action : async () => undefined
     };
 
     return new TestBuilder({
@@ -54,7 +54,7 @@ export class TestBuilder<C extends Context> implements TestImplementation {
   assertion(descriptionOrAssertion: string | AssertionDefinition<C>, check?: Check<C>): TestBuilder<C> {
     let assertion = typeof descriptionOrAssertion !== 'string' ? descriptionOrAssertion : {
       description: descriptionOrAssertion,
-      check: check ? check : async () => {}
+      check: check ? check : async () => undefined
     };
 
     return new TestBuilder({
--- a/packages/suite/src/dsl.ts
+++ b/packages/suite/src/dsl.ts
@@ -1,4 +1,4 @@
-import { TestImplementation, Context, Step, Assertion, Check, Action } from './interfaces';
+import { TestImplementation, Context, Step, Assertion } from './interfaces';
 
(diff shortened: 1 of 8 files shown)


packages/suite/src/dsl.ts
```diff
@@ -11,8 +11,10 @@ export function test<C extends Context>(description: string): TestBuilder<C> {
 
 export class TestBuilder<C extends Context> implements TestImplementation {
   public description: string;
-  public steps: Step[];
-  public assertions: Assertion[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public steps: Step<any, any>[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public assertions: Assertion<any>[];
```
cowboyd: I couldn't figure out how to get the collection of steps to type correctly without putting an any in here 🤷

packages/suite/src/interfaces.ts
```diff
@@ -30,7 +30,7 @@ export interface TestImplementation extends Test {
   children: TestImplementation[];
 }
 
-export type Action = (context: Context) => Promise<Context | void>;
+export type Action<C extends Context, R extends Context | void> = (context: C) => Promise<R>;
```
jnicklas: I'm assuming you made this change because you want to use `Action`, `Step`, etc... as the parameters for the DSL methods. It's worth thinking about if it'd actually better to add separate types for these. The type parameters added here serve a purpose when used from the test builder DSL, but they do not anywhere else, and since we do reuse these types in other places, it's worth considering if the added complexity here is worth it. I think it'd be better to just copy these types into `./dsl.ts` and leave these ones as is.

packages/suite/src/dsl.ts
jnicklas: As mentioned [here](https://github.com/thefrontside/bigtest/pull/504/files#r484054190) if we don't actually modify these types, we don't need to invent type parameters here.

jnicklas (changes requested):
Marking this as "changes requested" for clarity, see my previous comments re type parameters.

packages/suite/src/interfaces.ts
cowboyd: Makes sense. Moved all of the parameterized types directly into the dsl.

jnicklas (approved):
Looks great! Killer feature and very nicely implemented 😎

PR #507, opened 2020-09-06
equality check on Atom#set to stop updates of equal states

dagda1:
this is a quick fix to stop state updates of similar values causing updates to all slices

I've used [ramda equals](https://ramdajs.com/docs/#equals) because the check is always against 2 different objects with potentially the same values.

You could also use [fast-deep-equal](https://github.com/epoberezkin/fast-deep-equal) which benchmarks `R.equals` as one of the slowest ways.

I think the real issue though is that `Slice#update` ends up calling `Atom#set` which replaces the entire tree causing all slices to be notified of the change.

### DIFF
--- a/.changeset/atom-equality-check.md
+++ b/.changeset/atom-equality-check.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/atom": major
+---
+
+equality check in Atom#set to stop equal state subscription updates
--- a/packages/atom/src/atom.ts
+++ b/packages/atom/src/atom.ts
@@ -3,6 +3,7 @@ import { Channel } from '@effection/channel';
 import { subscribe, Subscribable, SymbolSubscribable, Subscription } from '@effection/subscription';
 import { Slice } from "./slice";
 import { Sliceable } from './sliceable';
+import { equals } from 'ramda';
 
 export class Atom<S> implements Subscribable<S,undefined> {
   private readonly initial: S;
@@ -22,6 +23,13 @@ export class Atom<S> implements Subscribable<S,undefined> {
   }
 
   set(value: S) {
+    // need to use a deep equality check on the object's values
+    // because this.state and value
+    // will always be 2 different objects but will often have the 
+    // same values
+    if (equals(this.state, value)) {
+      return;
+    }
     this.state = value;
     this.states.send(value);
   }
--- a/packages/atom/test/atom.test.ts
+++ b/packages/atom/test/atom.test.ts
@@ -163,7 +163,7 @@ describe('@bigtest/atom', () => {
       });
     });
 
-    describe('idempotent set', () => {
+    describe('prevent identitcal state notifications', () => {
       let subscription: Subscription<string, undefined>;
 
       beforeEach(async () => {
@@ -173,7 +173,7 @@ describe('@bigtest/atom', () => {
         subject.update(() => 'bar');
       });
 
-      it('iterates over emitted states', async () => {
+      it('only one notification for identical states', async () => {
         await expect(spawn(subscription.next())).resolves.toEqual({ done: false, value: 'bar' });
         await expect(spawn(subscription.next())).rejects;
       });
--- a/packages/atom/test/atom.test.ts
+++ b/packages/atom/test/atom.test.ts
@@ -162,5 +162,21 @@ describe('@bigtest/atom', () => {
         await expect(spawn(subscription.next())).resolves.toEqual({ done: false, value: 'quox' });
       });
     });
+
+    describe('idempotent set', () => {
+      let subscription: Subscription<string, undefined>;
+
+      beforeEach(async () => {
+        subscription = await spawn(subscribe(subject));
+
+        subject.update(() => 'bar');
+        subject.update(() => 'bar');
+      });
+
+      it('iterates over emitted states', async () => {
+        await expect(spawn(subscription.next())).resolves.toEqual({ done: false, value: 'bar' });
+        await expect(spawn(subscription.next())).rejects;
+      });
+    })
   });
 });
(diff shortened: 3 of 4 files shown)


jnicklas:
> I think the real issue though is that Slice#update ends up calling Atom#set which replaces the entire tree causing all slices to be notified of the change.

This is very much by design. The whole point of `Atom` is to centralize state management, it has a similar sort of purpose to and constraints as a Redux state manager. Yes, this is somewhat ineffecient, but the trade-off is that we do not need to keep track of which state is changed where and who needs to be notified of it. We chose this design pattern very deliberately and with full knowledge of the advantages and disadvantages.

I don't think this change as implemented buys us much at all. If we want to reduce the number of events, we should do so at the *subscription point*, not at the *generation point*. The state tree is large, and so checking deep equality is potentially *very* expensive, also it only means that we skip events where the state tree is *exactly the same*, which isn't a majority of cases. On the other hand at the subscription end, we might be looking only at a single text string (for something like a status), it is much easier to compare for equality here and debounce, though it's questionable how useful this actually is, since we'd likely debounce at the point of rendering anyway.

dagda1:
@jnicklas 

> > This is very much by design. The whole point of Atom is to centralize state management, it has a similar sort of purpose to and constraints as a Redux state manager.

I think I agree that deep equality check is probably not the right thing here.


So the code I noticed the problem was here

```typescript
  while(true) {
    yield timeout(100);

    if (yield isReachable(options.url)) {
      appStatus.set('reachable');
    } else {
      appStatus.set('unreachable');
    }
  }
```

calling `      appStatus.set('reachable');`

causes this code

```typescript
  set(value: S) {
    this.state = value;
    this.states.send(value);
  }
```

To run everytime and not just the slice getting updated.

Every slice gets notified of a state change whenever 1 slice changes.  Is this correct?

I noticed

```typescript
    capabilities.slice('alwaysMatch', 'moz:firefoxOptions', 'args')
      .over(args => args.concat(['--headless']))
```

Seemed to be working just on the slice

taras:
> causes this code
> 
>  set(value: S) {
>    this.state = value;
>    this.states.send(value);
>  }
>
> To run everytime and not just the slice getting updated.

Does this run only on this slice or all slices?

> Every slice gets notified of a state change whenever 1 slice changes. Is this correct?

It should. 

> I noticed
> 
>    capabilities.slice('alwaysMatch', 'moz:firefoxOptions', 'args')
>      .over(args => args.concat(['--headless']))
> Seemed to be working just on the slice

That would be surprising. And perhaps incorrect?

dagda1:
> It should.

That is surprising.  If we are using redux as an example, it has reducers that are like slices and it does a lot of diffing on old state and new state to ensure that only changed reducers are notified.

One of the problems people run into with the new React.context is not knowing that every change to the context causes a rerender.

I'm not going to labour the point, if this is by design then I am probably missing something.

I can close the PR.  I think either way, an expensive equality check is definitely not the right answer.

cowboyd:
@dagda1 @jnicklas @taras 

Digging into this I think one of the reasons that signals were getting crossed (certainly on my end) is that the Ramda lens library does not actually behave like every other lens library that I've used, including the one that I rolled by hand for Microstates.

When we use lenses, it is the lens's responsibility to determine what constitutes a change and what does not, and to return an "equal" object if the focus is being set to the same value. Apparently, ramda _always_ returns a new value for objects, even when the focus is the same. I don't know why they do this since, [as you can see in my homegrown version][1], it's super easy to do a strict equals check, and if it succeeds, we return the reference to the previous value unchanged since there is literally no more work to do. My guess is that when you are in "Ramda World", you are expected to use `equals` everywhere to determine equality, and so you can technically say that you're obeying the lens laws assuming everyone uses `equals()` always. I think this is a bad assumption, but I didn't design ramda, so maybe there's a deeper reasoning there.

Here's a [comparison with how monocle-ts behaves][2] when the focus of a lens is set to the same value, and as you can see, it just returns a reference to the previous value. This is the behavior that I expected from ramda, and would expect from any JavaScript lens implementation. My opinion is that strict equality checks at the root should be sufficient in deciding whether to publish a new state.

The good news is that if we use a lens library like monocle, we can do just that. And, we can still have the best of both worlds because lenses are really cool. Not only do we get strict equality checking out of the box, but we could, if we wanted to, implement custom lenses that do deep equality checks _at the leaf nodes_, and then return strict references to the previous value in the case that they're deeply equal. Imagine something like:

```ts
slice.set({ appStatus: 'reachable' });
```
where we only do the deep equals comparison on `{ status: 'reachable' }` which is cheap and yet resolve to the exact same strict equals reference as before.

In summary, I think we should

1. adopt monocle as a more sound lens implementation
2. add a strict equality check at the root in `Atom#set` not to publish identical values
3. filter subscriptions to slices so that only unique values are published
4. explore using deep-equal comparisons at the lens foci as a potential optimization if performance becomes an issue

[1]: https://github.com/thefrontside/microstates/blob/master/packages/microstates/src/lens.js#L67-L68
[2]: https://runkit.com/cowboyd/strict-equality-for-monocle-vs-ramda

dagda1:
@cowboyd wow, great answer.  I was a bit worried I had been way too annoying about this.  Irish people speak how they see it which is fine in Ireland but often does not travel well. 

> where we only do the deep equals comparison on { status: 'reachable' }

I think that is definitely what I would have expected

> filter subscriptions to slices so that only unique values are published

sounds great

> explore using deep-equal comparisons at the lens foci as a potential optimization if performance becomes an issue

Totally agree that you don't do anything until you actually know there is a problem to fix.  

> monocle-ts

definitely worth a punt or a spike. 

I got [a PR](https://github.com/gcanti/monocle-ts/pull/139) merged to update their examples and had a quick look.

They don't infer the types is the only downside so you have to explicitly add the type argument but I wonder if that could be added.

`const company = Lens.fromProp<Employee>()('company')`

cowboyd:
@dagda1 is migrating to monocle something you'd be interested in trying?

dagda1:
> @dagda1 is migrating to monocle something you'd be interested in trying?

Absolutely love to!  Sounds a hell of a lot of fun.

I still have  #502 outstanding, which is ready for review again whenever you have time.

Issue #379, opened 2020-07-02
Add built in interactors


jnicklas:
Once #359 is merged, we will want to add a suite of built in interactors to ship out of the box.

I've compiled a list of interactors here which will we can fill out as we come up with more ideas. Potentially useful filters are in parentheses.

Almost all interactors should have some visibility filter.

- [x] Link (href,target)

### Form elements

All form elements should have an `enabled` filter.

- [x] Select
- [x] TextArea
- [x] Checkbox (checked)
- [x] Button *should allow `<button>`, `<input type="submit">`, `<input type="button">`, `<input type="reset">` and `<input type="image">`*
- [x] TextField (value) *should use `<input type="text">` but also `<input>` without a type*
- [x] ColorField (value) 
- [x] DateField (value) 
- [x] DateTimeField (value) *should use `<input type="datetime-local">` since `datetime` is obsolete*
- [x] EmailField (value) 
- [ ] FileField (accept,capture,multiple) *how do we upload files?! this one is tricky*
- [x] HiddenField *not really normally available to users, so we maybe we should omit this?*
- [x] MonthField (value)
- [x] NumberField (value)
- [x] PasswordField (value)
- [x] RadioField (value,checked)
- [ ] RangeField
- [x] SearchField
- [x] TelephoneField
- [x] TimeField
- [x] UrlField
- [x] WeekField

## Acceptance Criteria

Must work and be tested with Popular Frameworks, including
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Svelte
- [ ] Ember

jnicklas:
There are a lot of fields that are "text-like", should there be some common interactor for all of these?

Namely text areas and also input without type and with type `text`, `number`, `email`, `password`, `search`, `tel` and `url`.

cowboyd:
Also, all form elements should have a `focus` action.

jnicklas:
So, one thing I just ran into as I was looking into building a `Link` interactor: ARIA roles.

Maybe this is not something we should worry about right now, but I thought it'd be worth mentioning. A link with `role="button"` should be treated as a button, not as a link, so should the `Link` interactor find such a link? Should the `Button` interactor? I think it would probably be good to respect ARIA attributes, but it does complicate things.

jnicklas:
Another wrinkle I ran into:

[...]

PR #508, opened 2020-09-07
Interactor definitions

jnicklas:
This adds `Link` and `Headline` as the first built-in interactors. See #379.

This implementation is quite minimal for now. It's meant to serve as a guideline for how and where interactors are added and tested, to serve as a jumping-off point.

### DIFF
--- a/packages/interactor/package.json
+++ b/packages/interactor/package.json
@@ -16,7 +16,7 @@
   "scripts": {
     "lint": "eslint '{src,test}/**/*.ts'",
     "mocha": "mocha -r ts-node/register",
-    "test:unit": "mocha -r ts-node/register test/**/*.test.ts",
+    "test:unit": "mocha -r ts-node/register 'test/**/*.test.ts'",
     "test:types": "dtslint types --localTs ../../node_modules/typescript/lib --expectOnly",
     "test": "yarn test:unit && yarn test:types",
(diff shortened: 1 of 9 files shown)


cowboyd (commented):
Looks good to me. @pittst3r Would you want to change anything about this before integrating it into the ui-users test suite?

packages/interactor/package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
     "lint": "eslint '{src,test}/**/*.ts'",
     "mocha": "mocha -r ts-node/register",
-    "test:unit": "mocha -r ts-node/register test/**/*.test.ts",
+    "test:unit": "mocha -r ts-node/register 'test/**/*.test.ts'",
```
cowboyd, line 19: I think single quotes might be a no-no in windows shell. /cc @jbolda

cowboyd (changes requested):
Might have some windows incompatibility

pittst3r (approved):

packages/interactor/package.json
jbolda: Aye, the shell processes the glob first with single quotes on windows. Escaped double quotes is pretty much always going to be the preferred option here.

cowboyd, line 19: @jbolda So something like this?
```suggestion
    "test:unit": "mocha -r ts-node/register \"test/**/*.test.ts\"",
```

jbolda: Yep! I'm sure there are so a handful of single quotes hanging around after my PR. I was focusing on just getting tests running (and linting was likely ignored).

cowboyd (approved):

PR #509, opened 2020-09-08
Remove mailbox API from effection-express

jnicklas:
Get rid of further usages of `Mailbox`.

### DIFF
--- a/packages/server/src/connection-server.ts
+++ b/packages/server/src/connection-server.ts
@@ -1,11 +1,12 @@
-import { Operation, fork } from 'effection';
+import { Operation, spawn } from 'effection';
+import { subscribe, ChainableSubscription } from '@effection/subscription';
 import { Mailbox } from '@bigtest/effection';
 import { Atom } from '@bigtest/atom';
 import { OrchestratorState } from './orchestrator/state';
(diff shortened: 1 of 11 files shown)


cowboyd (approved):
Adios 👋

jnicklas:
Made some more substantial changes to this, so would be good with a re-review. The connection server now uses the agent handler internally, so we're no longer duplicating this code.

packages/server/src/connection-server.ts
```diff
@@ -1,71 +1,46 @@
...
+      agent.set({ ...connection.data, agentId: connection.agentId });
 
-      yield fork(function*() {
+      yield spawn(function*(): Operation<void> {
         while (true) {
-          console.debug('[connection] waiting for message', agentId);
-          let message = yield options.inbox.receive({ agentId: agentId });
-
-          yield socket.send(message);
+          let message = yield options.inbox.receive({ agentId: connection.agentId });
```
jnicklas, line 32: We could take this even further here and make the inbox a Channel, but I wanted to leave this to a future PR.

cowboyd (approved):
The progress is real ❤️

PR #510, opened 2020-09-08
Prep for init command

### LINKED ISSUE #276: Better errors when an invalid launch configuration is specified.
The `--launch DRIVER_NAME` flag that gets passed to the `bigtest server` command tells the orchestrator server to automatically start and manage the process described by `DRIVER_NAME` where drivers are name value pairs that match a driver configuration to nickname. By default, the following drivers are configured for a bigtest project:

```js
    drivers: {
      chrome: {
        browserName: "chrome",
        headless: false
      },
      "chrome.headless": {
        browserName: "chrome",
        headless: true
      },
      firefox: {
        browserName: "firefox",
        headless: false
      },
      "firefox.headless": {
        browserName: "firefox",
        headless: true
      },
      "safari": {
        browserName: "safari",
        headless: false
      },
    },
```

So you can say `--launch chrome` or `--launch firefox.headless`, but it would be totally invalid to say `--launch floppy-nooby`. We _do_ some very basic validation of this, but not ahead of time. The `--launch` option should really be a `choice` in the yargs nomenclature where the values to choose from are actually sourced from the loaded config.

If we can do that, then we get some nice stuff for free:

1. `bigtest server --help` will show the choice of driver available
2. `bigtest server --launch INVALID-DRIVER` will print an informative error to the console.
3. `bigtest server --launch ch` can show completion options for both "chrome" and "chrome.headless"

jnicklas:
This does a few things:

- Change the structure of how we load CLI options so we don't have to load project config for all commands. The init command will not want to load project config.
- Change format of command line args to the standard dasherized form, reduce weirdness tax
- Remove default app, instead we will provide a sensible default in the `init` command.

Closes #276


packages/project/src/index.ts
```diff
@@ -47,16 +47,12 @@ export type ProjectOptions = {
...
-  url: "http://localhost:24000"
-};
-
 export function defaultConfig(configFilePath: string): ProjectOptions {
   return {
     port: 24002,
-    app: defaultApp,
+    app: {
+      url: "about:blank"
+    },
```
cowboyd: Let's just allow the app service to error if no url is provided, and make sure that we initialize the boilerplate with one in our `init` command.

cowboyd (changes requested):
See comment regarding default app options

packages/project/src/index.ts
jnicklas: Fixed!

cowboyd (approved):

PR #511, opened 2020-09-08
Support nullary interactors

### LINKED ISSUE #505: Nullary Interactors
## Motivation

Interactors currently require that one passes a string value as the first argument, which gets used with the `defaultLocator`:

```ts
const Button = createInteractor('button')({
  selector: 'button',
  defaultLocator: element => element.textContent
});

Button('Click me').click();
```

What if we are trying to make an interactor for a component there is only ever one of on a page though? Let's also say that this component doesn't have a way to identify it other than, for example, an `id` or `data-test` attribute, which we may already be using in the `selector` property of the interactor definition. This is very common:

- Navigational menus
- Modals/dialogues
- Large tables that take up the whole page
- etc.

It is currently possible to circumvent the string argument requirement by making the `defaultLocator` return an empty string and passing an empty string to the interactor:

```ts
MainNav('').click('Widgets');
```

This is confusing and does not scale well.

## Approach

Allow nullary interactors:

```ts
const MainNav = createInteractor('button')({
  selector: '#main-nav'
});

MainNav().click('Widgets');
```

This assumes the `selector` matches a single element, otherwise we will get an `AmbiguousElementError`.

pittst3r:
## Motivation

Closes #505 

Interactors currently require that one passes a string value as the first argument, which gets used with the `defaultLocator`:

```ts
const Button = createInteractor('button')({
  selector: 'button',
  defaultLocator: element => element.textContent
});

Button('Click me').click();
```

What if we are trying to make an interactor for a component there is only ever one of on a page though? Let's also say that this component doesn't have a way to identify it other than, for example, an `id` or `data-test` attribute, which we may already be using in the `selector` property of the interactor definition. This is very common:

- Navigational menus
- Modals/dialogues
- Large tables that take up the whole page
- etc.

It is currently possible to circumvent the string argument requirement by making the `defaultLocator` return an empty string and passing an empty string to the interactor:

```ts
MainNav('').click('Widgets');
```

This is confusing and does not scale well.

## Approach

Allow nullary interactors:

```ts
const MainNav = createInteractor('button')({
  selector: '#main-nav'
});

MainNav().click('Widgets');
```

This assumes the `selector` matches a single element, otherwise we will get an `AmbiguousElementError`.

### DIFF
--- a/.changeset/nullary-interactors.md
+++ b/.changeset/nullary-interactors.md
@@ -1,0 +1,13 @@
+---
+"@bigtest/interactor": minor
+---
+
+Support nullary interactors for singleton elements:
+
+```
+const MainNav = createInteractor('button')({
+  selector: '#main-nav'
+});
+
+MainNav().click('Widgets');
+```
--- a/packages/interactor/src/create-interactor.ts
+++ b/packages/interactor/src/create-interactor.ts
@@ -1,4 +1,4 @@
-import { InteractorSpecification, FilterImplementation, InteractorInstance, InteractorType, LocatorFn } from './specification';
+import { InteractorSpecification, FilterImplementation, InteractorInstance, InteractorType, LocatorFn, InteractorConstructor } from './specification';
 import { Locator } from './locator';
 import { Filter } from './filter';
 import { Interactor } from './interactor';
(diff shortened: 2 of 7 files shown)


cowboyd:
I've run into this as well, especially around singleton UI items like the the "Home Link" that usually appear at the top left of a page as is the name and/or logo of the website. 

Do you see the possibility of having nullary interactors that also have other locators?

pittst3r:
> Do you see the possibility of having nullary interactors that also have other locators?

Like this?

```ts
const Table = createInteractor('table')({
  selector: 'table',
  locators: {
    byId: element => element.id
  }
});

Table().has({ rowCount: 1337 });
// or
Table.byId('another-table').has({ rowCount: 1337 });
```

Totally possible. Here we are just saying that if you don't pass in a value for the default locator, then the default locator will let through all elements found by the selector.

jnicklas (approved):
This is a great feature, really, really love this!

PR #512, opened 2020-09-09
add @bigtest/eslint-plugin

dagda1:
The first part of validating that the test files are in a correct structure.

Firsts and foremost, we need eslint to check for things like a bigtest test file has a default export as eslint is an integral part of most devs and most CI pipeline's workflow.

I personally would prefer static analysis for a default export rather than requiring an object and looking for a default property.

This check also has to happen before the bundler as the bundler deals with the manifest that is generated from the manifest-generator.  We want the user to know as soon as possible that their test file is not in the correct structure.

Waiting for the bundler is a bad experience in my opinion and also too late.

In a later PR, I run the eslnit `CLIEngine` at runtime using the same rule and the user does not progress to the bundling state unless their test files are correct.

## Eslint Ruleset

There is currently only 1 rule  `@bigtest/require-default-test-export` that checks for a default export like the following:

```typescript
export default test('a test')
```
or
```javascript
module.exports = test('a test'); 
```
or even
```javascript
exports = test(' a test');
```

### DIFF
--- a/.changeset/eslint-plugin.md
+++ b/.changeset/eslint-plugin.md
@@ -1,5 +1,5 @@
 ---
-"@bigtest/eslint-plugin": major
+"@bigtest/eslint-plugin": minor
 ---
 
 add @bigtest/eslint-plugin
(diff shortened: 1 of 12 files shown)


taras (approved):
👍 this is great. Nice place to start.

jnicklas:
Great job, @dagda1!

PR #519, opened 2020-09-11
fix: remove all single quotes in package.json scripts

jbolda:
This is for windows compatibility reasons. It can only run scripts with the double quotes otherwise the terminal processes the globs which typically breaks the commands.


jnicklas (approved):

jnicklas:
Great job, thanks @jbolda!

PR #522, opened 2020-09-11
Add Gherkin-style given-when-then as an alternative DSL

jnicklas:
The current nomenclature of `step`/`assertion` is a bit technical, this is a slightly friendlier alternative, which should be familiar to anyone who has used cucumber, or something like rspec-given.

One downside of this is that the lexical difference between `when` and `then` is rather small, so it can be pretty hard to spot which is a step and which is an assertion. This could be solved via editor plugins and syntax highlighting, but it is definitely not ideal.

This may be less of a problem with Gherkin, since there is no functional difference between `When` and `Then`, it only affects how a person reads the test, not how the test runner does, so getting it wrong has lessened consequences.

### DIFF
--- a/packages/suite/src/dsl.ts
+++ b/packages/suite/src/dsl.ts
@@ -22,6 +22,37 @@ export interface AssertionDefinition<C extends Context> {
   check: Check<C>;
 }
 
+type StepArgs<C extends Context, R extends Context | void> = [StepDefinition<C,R>] | [string, Action<C,R>];
+
+function normalizeStepArgs<C extends Context, R extends Context | void>(stepArgs: StepArgs<C, R>): Step {
+  if(typeof(stepArgs[0]) === 'string') {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    let action = stepArgs[1] || ((async () => {}) as () => Promise<R>);
+
+    return { description: stepArgs[0], action } as Step;
+  } else {
+    return stepArgs[0] as Step;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


jnicklas:
This was pretty easy to implement, so I figured it would be a nice thing to add, but I'm not sure about the resulting DSL.

packages/suite/src/dsl.ts
```diff
@@ -35,31 +66,38 @@ export class TestBuilder<C extends Context> implements TestImplementation {
     this.children = test.children;
   }
 
-  step<R extends Context | void>(step: StepDefinition<C,R>): TestBuilder<R extends void ? C : C & R>;
-  step<R extends Context | void>(description: string, action: Action<C,R>): TestBuilder<R extends void ? C : C & R>;
```
cowboyd, line 39: The value, imho in having these overloads is so that the different ways of invoking the function are apparent from intellisense in a way that wouldn't be true if it's just `...args: StepArgs<C,R>`. Can we refactor to have the base method to be `...args: StepArgs` and also keep the overloads?

cowboyd (commented):
I'm all about the `given/when/then` syntax and would love to see how other folks perceive it. My only question really is should we make this as an alternative to `step/assertion` or should we make it a separate DSL. My greatest concern is that folks will be confused when they see their completion popup show.

```
test
   - assertion(....
   - child(...
   - step(...
   - given(...
   - when(...
   - then(...
```

jnicklas:
Let's close this for now and revisit it if we feel like it's a good idea.

Issue #487, opened 2020-08-28
Proposal: Reporter API


cowboyd:
As part of the [work][1] to [enhance the default CI output][2], we wanted to add a formatter that did more fancy things, like rendering per-agent progress bars, etc. At the same time, we need to think ahead to the time when we're reporting results as part of a Web UI, or to a cloud-hosted service.

Right now the CLI `runTest()` function [does the heavy lifting][3] when it comes to tracking summary information about a test run such as calculating performance times, counting the number of steps and assertions, etc..

I propose we move this work onto the shoulders of the orchestrator, and then provide a very thin API for handling the entire stream of a test run asynchronously, so that non-console based reporters and console based reporters / formatters don't have to do any work.

### Interfaces

The core interface will rely on `@effection/subscription` which is now becoming a very handy utility to model processes exactly like a test run:

1. do something at the start
2. do something for each item
3. do something with a return value modeling the totality of the computation

The reporter would be nothing but a function that takes some metadata about the test, and a subscribable to receive every event. The subscription would return a summary similar to the one currently computed by the CLI.

```ts
export interface Reporter {
  (info: TestRunInfo, run: Subscribable<RunResultEvent, TestRunSummary>): Operation<void>;
}
```

For example, to couple to the [current "lines" streaming formatter][4], we would write a reporter like this:

```ts
export const lines: Reporter = function*(_info, events) {
  formatter.header();

  let summary: TestRunSummary = yield subscribe(events).forEach(function*(event) {
    formatter.event(event);
  });

  formatter.footer(summary);
}
```

By assuming nothing about how it is reported, this lets us write literally any type of reporter that can do arbitrary operations asynchronously, including streaming to file, network, or stdout. And what's really nice is that we can combine reporters easily using function composition.

```ts
export function combine(a: Reporter, b: Reporter): Reporter {
  return function* compositeReporter(info, events) {
    yield fork(a(info, events));
    yield fork(b(info, events));
  }
}
```

### GraphQL changes

The key to super-flexible reporters is that the entire test run is represented almost entirely as a subscription, so that the representation can be literally anything. In order to allow reporting to happen from any context, not just the CLI, then, we want to do as much summarizing as possible in the orchestrator. That would mean enhancing the `Subscription { run }` API  to have:

- the first event contain a preview of the test run including which agents it will be run on.
- the last event contain a complete summary about the test run including counts, failures, or other reports broken down by agent.

### Caveats and Open Questions

[...]

PR #523, opened 2020-09-11
Prepare for reporter API

### LINKED ISSUE #525: Display assertion errors
With #462, we added sweet and fresh error output that diagnosed exactly what was going on in steps, but we forgot to display the error output for assertions. Right now, when an assertion fails, we don't have any idea why.

We should add error display logic to our assertions as well.

jnicklas:
In #487, we proposed a reporter API for pluggable test result reporters. In order to get there, we'll need to do some cleanup work.

Closes #525 

Output looks like this:

Screenshot 2020-09-11 at 18 44 13


jnicklas:
I found it really hard to pick good icons for the ok/failed/disregarded states. It was hard to find three characters which individually look good, also look good together, are easily distinguishable from each other, and convey their intended meaning. I ended up with `✓◦⨯`. A less unicode option may be to go with the traditional `.SE` which for example RSpec uses.

I also tried using `↪` for steps and `✓` for assertions in the progress line as well, but that looked very messy.

RobbieTheWagner:
@jnicklas any particular reason for removing ololog and manually using `console.log`? Ololog provided a lot of nice things for us, handled indentions without manually calculating the spaces, etc.

jnicklas:
@rwwagner90 I did not really feel that it was needed. ololog seems great, but it's also a huge library with many features. At this point in time we really only need colors (already provided by chalk), and indentation (trivial to implement ourselves), so it felt a bit overkill to me.

cowboyd (approved):
Looks great. 👍

RobbieTheWagner:
@cowboyd @jnicklas I disagree with the removal of ololog. It handles so much for us without a lot of manual calculations like this. We lose a lot with manual console.log usage. We shouldn't care about the size of the library, since we're writing a testing framework here and not shipping bytes with people's apps. Really wish we could have discussed this more before merging, as I would love to simplify things and not rip out dependencies others are adding without talking about it.

PR #524, opened 2020-09-11
Update return type of Interactor#has to match Interactor#is

pittst3r:
## Motivation

Return type is annotated incorrectly for `Interactor#has`. This appears to just be a mistake.

## Approach

Update the return type to match `#is`, a `ReadonlyInteraction`.


taras (approved):
👍

cowboyd (approved):

PR #528, opened 2020-09-11
Clear persistent storage

### LINKED ISSUE #447: Cleanup localstorage before test runs?
The bigtest agent and harness are running on the same origin, usually `localhost:24001`. This means that any session storage and localstorage persists between testcases in the browser runtime which violates the principle of isolation.

Imagine you have two test cases that log in, but if the authentication token is written to local storage in the first test case, the second will see it and skip the login step.

cowboyd:
> Closes #447 

Motivation
-----------

Imagine you have two test cases that log in, but if the authentication token is written to local storage in the first test case, the second will see it and skip the login step. The violation of isolation between test cases (shared persistent storage) means that you get unexpected and confusing consequences and that's not what we want with bigtest. Things should be in your test environment because you put them there not because they fell from the sky.

Approach
----------

This calls a `clearPersistentStorage()` operation before each lane is run that clears local storage, session storage, and indexedDB. In oder to workaround the fact that there is no way to enumerate indexedDB databases, the harness overrides `indexedDB.open()` method in the app frame to delegate to the test frame, and then the test frame, overrides it to stash the
name of the opened database in the agent frame so that it can persist between lane runs.

We could have cleared storage _after_ the lane run, and not had to go through the shenanigans of stashing the set of opened database names on the agent frame, but this way has the advantage of leaving all of the state in tact after a test run in the event that it needs to be inspected.

Todos & Open Questions
-------------------------
- **Triangulate on a plugin architecture** How could we implement this as a plugin... something that can run as part of each lane, and also coordinate between data raised from the harness, but stashed in the `test-frame`?


jnicklas:
It could be a bit problematic to add the clearing of IndexDB in a non-portable way. If someone starts off writing their tests using Chrome and then later on runs them in FF it seems like a bad experience to suddenly start failing there, especially since failure conditions would be subtle.

I think it'd be better to do this in a portable way, or not at all. I think the better solution here might be to simply leave IndexDB be. It's not very commonly used anyway, and we can add it in the documentation that it should be cleared manually.

cowboyd:
>I think it'd be better to do this in a portable way, or not at all. I think the better solution here might be to simply leave IndexDB be. It's not very commonly used anyway, and we can add it in the documentation that it should be cleared manually.

@jnicklas I agree. I forgot to open this PR as a "draft". I have a strategy that I think will work to make clearing indexedb work on all implementations, but it will require some research. As far as I can tell, you can't change the draft status of a PR after the fact, but I'll update the description to put WIP in the title.

jnicklas:
It's actually possible to convert back to draft now!

Screenshot 2020-09-14 at 15 59 45

jnicklas:
@cowboyd would it make sense to ship just local and session storage for now, and work on indexedb separately.

cowboyd:
> It's actually possible to convert back to draft now!

So cool!

As far as breaking it up. I can definitely do that. I do feel, however, that indexedDB should be part of the beta if we can since FOLIO uses indexedDB for its token storage, and anybody who uses `localforage` default configuration will use indexedDB if it is available.

cowboyd:
@jnicklas I was able to get this to work in a cross-platform way for all implementations of indexedDB by using the harness-override strategy. I tested it on firefox, but noticed that we actually have other failures on firefox for the agent (they don't seem like actual failures, but rather slight differences in the format of the stack traces generated by v8 and spidermonkey.

I've opened up a ticket to [track that](https://github.com/thefrontside/bigtest/issues/533)

jnicklas (approved):

packages/agent/app/clear-persistent-storage.ts
```diff
@@ -0,0 +1,34 @@
...
+  yield clearIndexedDB();
+}
+
+function* clearIndexedDB() {
+  let { openedDBNames } = getIndexedDBConfig();
+
+  // delete any database that were opened in previous lanes
+  for (let dbName of openedDBNames) {
+    let request: IDBRequest = window.indexedDB.deleteDatabase(dbName);
+    yield idbRequest(request);
```
jnicklas: This ends up being serial, I think? Probably not really worth it to parallelize this, right?

cowboyd: Hmm... deletes are pretty fast, but then again, why not make it parallel? My main reason for not wanting to do parallel was not having to use `fork()` which is quasi-deprecated, but we don't have a great replacement.

cowboyd: I went ahead and parallelized. There's no reason we should expose ourselves to a performance hit when it really makes no difference either way. The goal of effection after all is to make parallel so cheap and risk free that these decisions are a no-brainer, is it not?

Issue #506, opened 2020-09-05
Ensure no two test have the same path


cowboyd:
We encountered the following test files:

// companies.test.ts
```ts
import { test } from '@bigtest/suite';

export test('Admin UI')
  .step()
```

And also in

// accounts.test.ts
```ts
import { test } from '@bigtest/suite';

export test('Admin UI')
  .step()
```

As a result, both of these tests had a path of `['All Tests', 'Admin UI']` so the steps associated with these tests ran twice. In our case, Mirage was installed twice. We should invalidate a bundle that has duplicate paths of tests.

related:  #294

jnicklas:
I've been thinking about this issue a bit and I wanted to point a flaw in our thinking here, and possible alternatives to how we might think about it:

The following is a pretty realistic test scenario:

``` typescript
test("go to third page")
  .step(Link("Next").click())
  .step(Heading("Page 2").exists())
  .step(Link("Next").click())
```

While there is no duplication among *tests* the two *steps* that have click the `Next` link have the same path. We currently assume this cannot happen, which is incorrect. IMO it should be perfectly fine to have duplicate descriptions among steps.

So why do tests need to be unique, but steps don't? Simply because step paths are only relevant in the context of one particular test run, whereas test paths should be stable across test runs. It would be nice to be able to re-run a particular test based on its path. But running an individual step doesn't make sense. Steps are only ever run as part of an entire lane.

Two questions:

1) How about assertions? Should assertions be unique? I can't see a reason for having multiple assertions with the same name, and it would be quite convenient to be able to re-run them by path, even if for example the order changes or the steps before it change, or whatever.

2) How should we express the path to a step? Currently path are arrays, this has the advantage of not having to parse the path, and not leaving any holes where using some special character in a description could cause bugs. Should we just use an array index for steps, like this:

``` javascript
["All tests", "go to third page", 0]
["All tests", "go to third page", 2]
```

cowboyd:
For (1), I agree with your initial inclination that we should disallow duplicate assertions. Any assertion that changed the outcome of a test by duplicating or not duplicating would, by definition be an invalid assertion.

For (2) An array does make sense since steps are implicity ordered. That's kinda their whole deal. And in fact, on the _runner_ side of things, we only use the test path to run steps, and not the step path https://github.com/thefrontside/bigtest/blob/fillin-react/packages/agent/app/run-lane.ts#L67

[...]

Issue #294, opened 2020-05-18
Verify assumptions in test suite


jnicklas:
There are two invariants in our test suite format that we should verify are met, and if they are not, the test suite should not be accepted:

1. All paths of descriptions should be unique. That is two steps/assertions cannot have the same description within the same test and two sibling tests cannot have the same description.

2. There should be a maximum depth that we allow for tests. This is mainly so we can express recursive queries in our graphql without running into problems.

cowboyd:
@dagda1 @pittst3r Given recent upgrades in the bundler architecture, any idea how we can validate this?

dagda1:
could these be linting rules?  

Should we have an `eslint-plugin-bigtest`?

something like [this](https://eslint.org/docs/rules/max-depth)

cowboyd:
Yes, that is actually a great idea, and more evidence for establishing the bigtest tests as a separate bundle in order to make #443

Issue #280, opened 2020-05-13
Type-check test bundle


cowboyd:
We use parcel to bundle the manifest which means that if you're writing your tests in TypeScript, everything will "Just Work" ™️.  But parcel uses babel typescript to just strip the type annotations away from your javascript code and not actually type-check. This means that invalid code can get through and try to be required by the bundle.

We need to typecheck the test code before hand and communicate the error to the user rather than crash the server which is what happens now.

jnicklas:
This issue is really such a bummer. I still can't understand how the parcel folks feel that this behaviour is good :(. I almost feel like we *shouldn't* do anything about this though. This is really parcel's problem to fix, and I don't love doing something like invoking tsc if we see any TypeScript files. It feels to me like everything we do in bigtest to fix this problem is going to be a really nasty hack which might break in unexpected ways.

pittst3r:
We are now using rollup to build the test files. However, we are still using babel to handle TS, which does not typecheck. That said, I think it is a better DX to have typechecking and building as separate steps because:

- TSC is slow
- Sometimes trivial type errors get in my way when I'm moving quickly and I want to defer fixing the type error
- We can introduce the idea of Validators which is a pluggable way to handle linting and type checking

Validators have the following attributes:

- There can be one or more of them
- They are pluggable
- They are kicked off when the test files change
- They only validate test files
- They do not block the test run following the rebuild, even if they fail

Questions about Validators:

- Do they use config files in the root of the project or the test directory?
- How should they report results? After the test run?

cowboyd:
> Do they use config files in the root of the project or the test directory?

For all intents and purposes, I think we should treat the test bundle as its own self-contained javascript module, which means that it should not automatically pull in things from the ambient environment. This will let us embed a BigTest suite into _any_ project no matter what it looks like. If we don't, we're going to be playing an endless game of whack-a-mole with the fractal complexity of potential JavaScript configuration.

I like the validator pattern you've laid out because it should also allow us to store bundle validation state in the `Atom` so that when building stateful ui's and inpsectors, this information is readily available. A good heuristic when designing these APIs could be "what information do I need to collect to build a stateful UI on top of this"

jnicklas:
So I think I have a way forward for this:

I've experimented a lot with the `@rollup/plugin-typescript` plugin and how it works, and I think we can make it work and this is how I think we should proceed on integrating it:

1. We add a config option to `bigtest.json` to specify the path for a tsconfig file.
[...]

PR #530, opened 2020-09-12
runtime checking of eslint rules such as a default export

dagda1:
This PR is to add runtime checking of any eslint rules that would stop the user's build executing.

It contains 1 validator

[EslintValidator](./packages/bundler/src/validators/eslint-validator.ts).

This will call the `eslint cli engine` but only with the `@bigtest/eslint-plugin` rules.  We can't rely on the user's eslint files as rules can be disabled or removed and things like a default export are mission-critical.

For my money, any validation logic like this should utilise your bundler of choice and the thought of having this external to the bundler seems a waste of effort.

We should write plugins that utilise our bundler of choice but write them in such a way that the core logic knows nothing about the bundler.


dagda1:
@cowboyd is there any feedback on this PR?

cowboyd:
> @cowboyd is there any feedback on this PR?

Sorry, I haven't had a chance to get to this yet, but I'll make sure to take a look today

dagda1:
> > @cowboyd is there any feedback on this PR?
> 
> Sorry, I haven't had a chance to get to this yet, but I'll make sure to take a look today

No problem, whenever you have time. 

no need to apologise.

cowboyd:
I'm coming around to the idea of using eslint to validate the bundle, because the only way that we can reliably validate it at runtime is by evaluating it inside of context in which it will actually be used (in this case the web). The further we go, the more I expect that we won't even be able to load test trees inside Node like we do now simply because there will be modules that statically assume a browser context. I ran into this when trying to use the `localforage` library which assumes that `localStorage` is a global object and merely importing it caused a runtime exception. I had to work around it by using dynamic imports.

I suspect that we may eventually need to move the responsibility for bundling the test tree into the driver itself. 

For the moment though, it seems that the most robust way to do certain validations is with static analysis. What I'm unclear on how it all ties together is what our endgame is for validation. What I'd like to see before we add any validations to the bundle, eslint or otherwise is the design for how we account for

1. syntax errors in JS files
2. syntax errors in TS files
3. module-level evaluation errors in syntactically valid JS and TS
4. syntactically valid, module-level evaluation succeeds, but test is ill-formed
    * no default export
    * tests with same virtual path
    * tests beyond a certain depth
  
And having the simplest possible uniform representation for all of these error cases (as well as any others I may have missed). In other words, what's the optimal configuration of the GraphQL schema and the orchestrator state that supports it. We can then work backwards from there to fill in the missing pieces of what goes where.

dagda1:
@cowboyd I am basically saying as I have done in this PR, that all validators are XXXX plugins (XXX is currently rollup).

But we write them in such a way that the core functionality knows nothing about the particular bundler.

Errors go into the atom state just like any bundler errors that we get and this is already hooked up.

I was thinking of experimenting with the typescript typechecker of forking the process.  Or would effection be a good use case for this:

```typescript
    this.service = childProcess.fork(
```

I use a webpack plugin that does this and it does cut down on build time.

You can definitely do this from a plugin of any bundler I think.

cowboyd:
> I am basically saying as I have done in this PR, that all validators are XXXX plugins (XXX is currently rollup).
>
>But we write them in such a way that the core functionality knows nothing about the particular bundler.

What about errors that happen at a level above the bundler. For example, how do we represent it when we can't successfully load the bundle to harvest the metadata a bout it even when it bundles successfully and without errors?

Apart from that, how do you feel about making the actual eslint validator part of the eslint-plugin? It would validate the plugin architecture and decouple the core from eslint. If you're ok with that, we can tackle it in a separate PR.

dagda1:
@cowboyd I've added the validator and the rollup plugin into the eslint-plugin package.

dagda1:
So much better we can log now without fear of multiple identical events getting raised.

save

jnicklas:
I've spent a lot of time this week thinking about how we can move forward with this PR and with solving #294, which this PR is trying to address.

My proposal is that we should close this PR and solve the issue that this PR is trying to address in another way. I'll try to lay out why I think this is the best way forward.

Let's look at the classification that @cowboyd posted above:

1. syntax errors in JS files
2. syntax errors in TS files (see #280)
3. module-level evaluation errors in syntactically valid JS and TS (see #592)
4. syntactically valid, module-level evaluation succeeds, but test is ill-formed
    * no default export
    * tests with same virtual path (see #294, #506)
    * tests beyond a certain depth (see #294)

I believe that (1) and (2) should be surfaced as bundling errors. I have done some experiments on integrating the `@rollup/plugin-typescript` plugin in our build pipeline, and while there are [some issues with this](https://github.com/thefrontside/bigtest/issues/280#issuecomment-704845746), I think we can make it work.

As for (3), I believe that #593 is the right way forward towards handling these types of errors. Essentially, a module level error would be handled by the agent.

This way (4) would also be handled by the *agent* which would perform a set of validations against the loaded manifest to check whether it is valid, if they cannot be verified then an error is returned. If there is no export from a test file, the agent will see this and raise an appropriate error.

This way forward, we can verify all of our assumptions without having to involve eslint.

I do think that eslint still has a useful place. The eslint plugin will be really useful for validating that there is a default export, and also other lints we might want to add in the future (a duplicate path line would be great, if that's possible), however, we won't be running eslint as part of our build pipeline. Users of bigtest will have to run eslint themselves.

In the future, we could open up for extending our rollup pipeline, so users can add whatever plugin they like. It's worth mentioning that there seems to already exist a pretty solid [eslint plugin for rollup](https://github.com/TrySound/rollup-plugin-eslint), so we don't need to maintain this ourselves.

If we move forward in this direction and we close this PR I want to nevertheless thank you @dagda1 for all of the work you put into this, it's been very helpful in fleshing out where we want to go with this. I also want to apologize for being so slow on providing my feedback here, there've been so many threads with trying to close out beta issues that I haven't had the peace of mind to really sit down and think through this problem thoroughly enough to come up with a plan.

jnicklas:
One thing from this PR that we'll definitely want is that it surfaces the warnings generated during bundling. That is warnings generated primarily by rollup plugins. There is some code in there which stores and then displays the warnings, which is something we'll want IMO.

dagda1:
@jnicklas IMHO relying on a default export, in general, is a bad thing.

I think the real fix is to not have to rely on this.

I personally don't think static analysis is used enough in js or typescript land and if not eslint then a babel plugin could do the same job better than any run-time check.  There are other things about the structure of the code like max tree depth that are great uses of static analysis but I can appreciate the points you raised and will definitely bow to the greater knowledge of the fuller picture.

jnicklas:
> IMHO relying on a default export, in general, is a bad thing.

You mentioned this before, but I didn't fully understand why you were against the default export, could you elaborate on this? What would be the alternative?

> I personally don't think static analysis is used enough in js or typescript land and if not eslint then a babel plugin could do the same job better than any run-time check.

I totally agree that static analysis is good and useful, I'm a huge fan of Rust, after all.

Unfortunately while we could certainly use static analysis to give us an early warning that something is amiss, I don't think static analysis will be powerful enough to give us a definitive answer whether a test file is valid or not.

Consider the following example:

``` typescript
let t = test("foo")

if(navigator.userAgent.match(/Firefox/) {
  t = t.step("my step")
}

if(navigator.userAgent.match(/Chrome/) {
  t = t.step("my step")
}

export default t;
```

Admittedly this is a big far fetched, but note that we add steps depending on the userAgent, if we add both then we violate the "no duplicate paths" rule. This is probably perfectly fine, since we won't be adding both, but static analysis wouldn't be able to determine whether these conditions are mutually exclusive.

This example is solvable, but there are a ton of other patterns which are hard to see through with static analysis. For example, this is an evil way to get past static analysis to create an invalid test:

``` typescript
function recurse(test, number) {
  if(number === 0) {
    return test;
  } else {
    return test.child("child", (child) => recurse(child, number - 1));
  }
}

export default recurse(test("my test"), 20);
```

I think what will serve us best is a hybrid approach of having eslint rules which give an early warning in simple cases, which are easy to detect, and then having stronger validation which guards us against critical errors.

dagda1:
That is a good example of the static analysis with only the regex literal being different but I think you could still determine that they were not the same.  Might be wrong on this one but I'm slightly interested to find out if you could :). IIRC you can parse the RegExLiteral into the actual value.  

As far as the default export goes, there are several reasons and I'll do my best to explain them.

- Having a default export has nothing to do with writing a test.  Why do I have to have a default export? (I know this answer but a new user might not).  I might be quickly having a look and then get these weird errors about default exports.

- I honestly do believe that most people in js land and ts land are agreeing that default exports are a bad thing, more so in typescript land but we are making that an absolute rule here, couple of quick examples

```ts
// source.ts
export default {}

import whatever from './source';
```

as opposed to

```ts
// source.ts
export const whatEver = {};

// typo
import { whatever } from './source'; // no exported member whatever
```

I think most are realising that named exports give better type safety.
We are going against the grain by stipulating that we need a default export.

- we are having to write code to maintain this rule and rather than trying to enforce it we should allow the user to export what they want or don't export anything which sort of feels more natural.

cowboyd:
Does TypeScript have a way to declare an interface for a module? I mean, really a module is just a function whose input is a map of imports, and whose output is a map of exports. Seems like it should be a straightforward thing.

dagda1:
> Does TypeScript have a way to declare an interface for a module? I mean, really a module is just a function whose input is a map of imports, and whose output is a map of exports. Seems like it should be a straightforward thing.

As ever :) there is an existing issue for [this](https://github.com/microsoft/TypeScript/issues/420)

TLDR; not currently

PR #531, opened 2020-09-14
Force resolution to yargs-parser

taras:
## Motivation

image

## Approach

Add resolution for `yargs-parser` to "^13.1.2"


cowboyd:
Do we need a changeset here? I don't think the `resolutions` field in the root will effect the distributable artifacts in any way.

cowboyd (approved):

PR #534, opened 2020-09-15
chore: timeout tests and checkout@v2

jbolda:
This is the last of the workflow changes besides the `node_modules` cache. Upgrading the checkout which is faster and improves functionality on reruns. It checks out the sha instead. The tests shouldn't really on a specfic git fetch depth so it shouldn't break anything.

Also adding a timeout as the default is 6hrs and this makes it less possible to accidentally swamp CI.


cowboyd (approved):
💖

PR #538, opened 2020-09-17
Fail run on build failed

### LINKED ISSUE #279: Better error messaging if manifest fails to load.

### LINKED ISSUE #421: Test run hangs when there is a runtime error loading the bundle
The following error was occuring when `evaling` the test bundle in preparation for a test run:

```
instanceOf.js:14 Uncaught ReferenceError: process is not defined
    at instanceOf.js:14
    at createCommonjsModule (manifest-8bc262b6af9740b56f53146038705774960d60f436e124898df86782b9cea4cb.js:20)
    at keyValMap.js:26
    at manifest-8bc262b6af9740b56f53146038705774960d60f436e124898df86782b9cea4cb.js:4
    at manifest-8bc262b6af9740b56f53146038705774960d60f436e124898df86782b9cea4cb.js:5
```

because there was a module that was assuming to be inside of `node`. The expectation is that it should report failure immediately since the test run could not even be attempted due to a bad bundle. However, what actually happens is that the test hangs forever.

jnicklas:
If there are any build failures in the test files causing the manifest build to fail, fail the test run and print the failure that caused the build error in the first place.

Closes #279
Closes #421

### DIFF
--- a/packages/cli/src/format-helpers.ts
+++ b/packages/cli/src/format-helpers.ts
@@ -112,7 +112,13 @@ export function standardFooter() {
         recursiveChildrenResults(result);
       }
     });
-    console.log(chalk.grey('────────────────────────────────────────────────────────────────────────────────'));
+    if(testRun.agents.length) {
+      console.log(chalk.grey('────────────────────────────────────────────────────────────────────────────────'));
+    }
+    if(testRun.status === 'failed' && testRun.error) {
+      errorLines(testRun.error).forEach((line) => console.log(chalk.red(line)));
+      console.log('');
+    }
     console.log(
       testRun.status === 'ok'
         ? chalk.green('✓ SUCCESS')
--- a/.changeset/rare-snails-chew.md
+++ b/.changeset/rare-snails-chew.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/atom": minor
+---
+
+Atom and Slice's `once` never returns undefined. The subscription never terminates, so an undefined value will never happen in practice.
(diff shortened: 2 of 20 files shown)


jnicklas:
Seems like the build is failing, but I don't have any more time to check out why this week.

packages/cli/src/format-helpers.ts
```diff
@@ -112,7 +112,13 @@ export function standardFooter() {
         recursiveChildrenResults(result);
       }
     });
-    console.log(chalk.grey('────────────────────────────────────────────────────────────────────────────────'));
+    if(testRun.agents.length) {
+      console.log(chalk.grey('────────────────────────────────────────────────────────────────────────────────'));
+    }
+    if(testRun.status === 'failed' && testRun.error) {
+      errorLines(testRun.error).forEach((line) => console.log(chalk.red(line)));
```
cowboyd, line 119: We should probably output an error here indicating that the test run couldn't be attempted because the bundle had an error in it. Otherwise, it just blops the error the console without context and that's pretty confusing:

image

cowboyd (commented):
Not sure of the best way to accomplish this, but running the test command with a broken bundle felt a bit jarring since I was expecting a successful test run, bun instead got a syntax error printed to the console without any context.

jnicklas:
@cowboyd I added a better explanation when a failure occurs:

Screenshot 2020-09-22 at 12 23 33

cowboyd (approved):
Major improvement 💯

PR #540, opened 2020-09-17
✨Add lists of steps in one go

cowboyd:
Motivation
-----------

One of the things we've noticed develoing large test suites is that it is very common to have sequences of three to four steps. For example, here is a real-life example of a multi-step interaction that represent navigation to a page, then a sub-page, and then setting some filters:

```js
.step(App.visit('/users/preview/1'))
.step(Button('Fees/fines').click())
.step(Link('Create fee/fine').click())
.step(Select('Fee/fine owner*').select('testOwner'))
.step(Select('Fee/fine type*').select('testFineType'))
```

the fact that we have to add each step with a `step` invocation is not
only annoying to the author of the step, it more importantly obscures
what is happening to the reader of the test case. 

Approach
----------
We can more simply express a list of steps using a `stepList` method where we add all the steps in a single shot:

```js
step(
  App.visit('/users/preview/1'),
  Button('Fees/fines').click(),
  Link('Create fee/fine').click(),
  Select('Fee/fine owner*').select('testOwner'),
  Select('Fee/fine type*').select('testFineType'));
```

Alternatives
------------

One thing to consider would be to just add yet another overload to the `step` method that would expect a list of steps. That way, there's less surface API and it's clear that a list of steps is just another type of step.

### DIFF
--- a/packages/suite/types/dsl.ts
+++ b/packages/suite/types/dsl.ts
@@ -38,10 +38,10 @@ test('a test')
   })
 
 // Context Async
-test('a test')
+let t1 = test('a test')
   // should not be able to return non-object
   // $ExpectError
-  .step('return nonsense', async () => {
+  t.step('return nonsense', async () => {
     return "foo";
   })
 
@@ -53,10 +53,10 @@ test('a test')
   .step({ description: "consume from context", action: async () => {} })
 
 // Context Sync
-test('a test')
+let t2 = test('a test')
   // should not be able to return non-object
   // $ExpectError
-  .step('return nonsense', () => {
+  t.step('return nonsense', () => {
     return "foo";
   })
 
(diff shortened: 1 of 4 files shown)


jnicklas:
Some comments:

1) How about `steps` instead of `stepList`? Reads a bit nicer in my opinion.
2) If we make the function variadic we don't need to wrap the arguments in an array
3) Should we add `assertions` as well, for symmetry?

cowboyd:
> Some comments:
> 
> 1. How about `steps` instead of `stepList`? Reads a bit nicer in my opinion.

I really wanted to do that at firt, but `steps` is already a variable which is required to implement the `Test` interface. We could get around this with a `[Symbol.toTest]()` method, but that felt like a lot of complexity to bite off for a relatively simple feature  😭 😭 😭 


> 2. If we make the function variadic we don't need to wrap the arguments in an array

We could also just generalize the current `step()` function which already takes a single step definition as an object to just take many.

> 3. Should we add `assertions` as well, for symmetry?

Probably?

jnicklas:
I like the idea of generalizing `step`, if it's possible to type properly, than I think that'd be the nicest solution.

jnicklas (changes requested):
Some tests would be good.

packages/suite/types/dsl.ts
```diff
@@ -38,10 +38,10 @@ test('a test')
   })
 
 // Context Async
-test('a test')
+let t1 = test('a test')
   // should not be able to return non-object
   // $ExpectError
-  .step('return nonsense', async () => {
+  t.step('return nonsense', async () => {
```
cowboyd, line 44: I had to break these expressions apart since with the overloaded `step` method, typescript was putting the location of the error on the `test()` invocation and not the `.step()` invocation and so in order to ensure that the correct error was being tested, I needed to force the error to only possibly be where expected

cowboyd:
@jnicklas revamped and tests added to use the varargs approach.

jnicklas (approved):

jnicklas:
Looks great!

PR #542, opened 2020-09-18
chore: test on mac

jbolda:
This is a strict copy of our workflow to run on Mac. Copying because, as previously discussed, the prepack job can finish and continue to the test job separately for each OS.

### DIFF
--- a/.github/workflows/test-mac.yml
+++ b/.github/workflows/test-mac.yml
@@ -1,0 +1,59 @@
+name: Test Mac
+
+on:
+  pull_request:
+    branches:
+      - 'v[0-9]+'
+
+jobs:
+  prepack:
+    runs-on: ${{ matrix.platform }}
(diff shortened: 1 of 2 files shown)


cowboyd:
@jbolda thanks! 🍻 

In order to get this in smoothly, how do we roll this out to the project? The main issue is that we have outstanding PRS against these checks which are changing names. Basically how do we order merging down of existing PRS versus introducing the new checks and making them required.

jbolda:
As far as I can tell, the checks work based on the job name and not the check group name, e.g. it checks for `agent` and not `Test Ubuntu / agent`. I think this means you won't need to adjust the required checks, but I am not certain if that means _all_ `Test X / agent` need to pass or just one 1 of X. I haven't tried setting required checks where names overlap like this.

cowboyd (approved):
Let's go for it then 👍

PR #543, opened 2020-09-20
Dockerisation experiment

pittst3r:


pittst3r:
This worked, but I'm going to turn this into an issue so we can discuss the design now that I've learned about many of the requirements.

PR #549, opened 2020-09-21
Hide splash page when app frame is active

### LINKED ISSUE #435: Agent Frame sometimes interferes with test Frame
Depending on the CSS of the app under test, the agent frame doesn't play nicely with it:

BigTest agent clobbering application frame

jnicklas:
Closes #435

### DIFF
--- a/.changeset/silver-apricots-sleep.md
+++ b/.changeset/silver-apricots-sleep.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/agent": patch
+---
+
+Hide splash page when app frame is active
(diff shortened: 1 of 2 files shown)


cowboyd (commented):
How does this even work? Will it work on a dark themed desktop?

jnicklas:
The iframe is always in front of the splash page. The problem only occurs when the app does not have a background color on its body, so the frame shows through the page. By adding a background color to the iframe we simply paint over it ;) The default background color in browsers is always white as far as I know, so it does look correct, regardless of nightmode or not.

jnicklas:
And the iframe does not initially have a `src` attribute, so the splash page is still shown before we visit the first time.

cowboyd (approved):
👍

PR #550, opened 2020-09-21
Time out if server start takes too long

### LINKED ISSUE #405: Fail if orchestrator doesn't start up within a specified about of time.
Before the `ci` command can run it has to start a server. However, the server can hang on its startup for many reasons (including but not limited to)

* If the app command runs successfully, but the app never successully starts (because of a build error for example) 
* some problem building the manifest.
* inability to connect an agent.

In those cases, we would like the `ci` command to fail with an error indicating what was ready and what wasn't

jnicklas:
Closes #405

This takes the approach of timing out in the CLI, rather than in the orchestrator.


packages/cli/src/cli.ts
```diff
@@ -93,5 +110,7 @@ function parseOptions(argv: readonly string[]): CLIArguments {
     .command('ci', 'start a server and run the test suite', (yargs) => runOptions(startOptions(yargs)))
     .demandCommand()
     .help()
-    .parse(argv) as unknown as CLIArguments // types generated by yargs are inadequate
+    .parse(argv)
+
+  return { command: parsed._[0], ...parsed } as unknown as Options // types generated by yargs are inadequate
```
jnicklas, line 115: TypeScript is unable to narrow the enum types when looking at the array value, which is why I've extraced it into a property here so the types are properly narrowed in the `if` statements above.

cowboyd (approved):
Two minutes seems ample. One we can get the readiness state reflected in the Atom, It would be *really* nice to understand which services are preventing you from proceeding. E.g.

```
`Timed out waiting for server to start after ${options.timeout}ms. Still waiting for:
  - manifest builder
  - app server
```

It would also be awesome, in CI mode to fail immediately if, for example, the bundler errors out.

jnicklas:
That would definitely be a nice future enhancement :)

Issue #448, opened 2020-08-06
specify root of test tree in with custom tests


cowboyd:
Right now, we hard-code the root of the test suite to the following:

```js
{
  description: "All tests",
  steps: [],
  assertions: [],
  children
}
```

where `children` is all of the test suites that we find by loading the files. This is great for making sure that our entire suite is recursive and accessible at any point. However, what if we want to change the name from "All Tests" to something else? More importantly, what if we want to add steps that are executed for _every single test_ in the suite? Right now, you would have to write a higher-order function to pass to the builder:

```ts
import { test } from '@bigtest/suite';
import { addGlobalSetup } from './helpers';

export default addGlobalSetup(test("my test"))
  .step()
  .assertion()

```

What we'd really like to do is specify in the configuration which root suite to use in your `bigtest.json`

```json
{
  "tests": {
    "files": ["test/**/*.test.ts"],
    "root": "test/all-tests.ts",
  },
   "app": { "url": "http://foo.bar.com"},
  "launch": ["chrome.headless"],
}
```

Our boilerplate installed by the `init` command would just drop the `all-tests.ts` into your `tests/` directory with the default value being:

```ts
// this is the root of your test tree
// steps added here will be run as part of every single test
export default test("All Tests")
```

That way, adding global steps are a snap, putting global things in the context are a snap, and if you want to configure the root node in any other way, it is transparent to you.

cowboyd:
A way we can type everything and magic everything:

```ts
import { test as testOG } from '@bigtest/suite';
import { startBrowserSimulation } from '@my/simulation';


// add steps to this test to run for every test in the test suite.
const allTests = testOG("All tests")
  .step("install simulated backend", async () => {
    return { server: await startBrowserSimulation() };
  })

export type AllTests = typeof allTests;

export function test(description: string): AllTests {
  return testOG(description);
}

export default allTest;
```

PR #551, opened 2020-09-21
Filtered run

### LINKED ISSUE #386: Enable running a subset of the test suite
Currently it's all or nothing. Every test run will run the entire suite. We should make it possible to run only part of the suite, ideally both by file and line. This will also be useful since it will make it far easier to write tests which test the entire stack, since we don't need a separate suite for each set of successes/failures.

jnicklas:
Limit test run to subset of test suite

Closes #386 
Depends on #538

### DIFF
--- a/packages/server/src/filter-test.ts
+++ b/packages/server/src/filter-test.ts
@@ -1,0 +1,33 @@
+import { Test } from '@bigtest/suite';
+import * as path from 'path';
+import * as fs from 'fs';
+
+interface Options {
+  files: string[];
+}
+
+function comparePaths(a: string, b: string) {
+  return path.resolve(a) === path.resolve(b);
+}
+
+export function filterTest(test: Test, options: Options): Test {
+  let children: Test[] = [];
+
+  if(options.files.length) {
+    for(let file of options.files) {
+      let child = test.children.find((c) => c.path && comparePaths(c.path, file));
+
+      if(child) {
+        children.push(child);
+      } else if(fs.existsSync(file)) {
+        throw new Error(`file with path '${path.resolve(file)}' exists, but is not part of your test files, if you want to run this test, adjust the 'testFiles' setting in 'bigtest.json'`)
+      } else {
+        throw new Error(`file with path '${path.resolve(file)}' does not exist`);
+      }
+    };
+  } else {
+    children = test.children;
+  }
+
+  return { ...test, children };
+}
--- a/packages/server/src/filter-test.ts
+++ b/packages/server/src/filter-test.ts
@@ -4,6 +4,7 @@ import * as fs from 'fs';
 
 interface Options {
   files: string[];
+  testFiles?: string[];
 }
 
 function comparePaths(a: string, b: string) {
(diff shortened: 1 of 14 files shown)


packages/server/src/filter-test.ts
```diff
@@ -0,0 +1,33 @@
...
+      } else {
+        throw new Error(`file with path '${path.resolve(file)}' does not exist`);
+      }
+    };
+  } else {
+    children = test.children;
+  }
+
+  return { ...test, children };
+}
```
jnicklas: I wasn't entirely sure how to write this. Fundamentally all `Test` nodes are the same, and it shouldn't really affect the filtering at which *level* of nesting a test occurs. However, the `path` property only exists at the *second* level of nesting (just below the root). I thought about various ways of trying to detect whether there is a `path` property, but this all felt pretty brittle, so in the end I opted to just hard-code that we're only filtering the second level. Still not super happy with this.

cowboyd: It's really cool that since the orchestrator and agent only think in terms of a single tree node, we can implement this feature without changing the agent at all.

Note that with the advent of #448, the root node _will_ actually have a file path. I guess this means that we won't run any assertions from the root test if it isn't in the list?

Anyway, I think we will be "eventually consistent" 😆

packages/server/src/filter-test.ts
```diff
@@ -0,0 +1,33 @@
...
+  let children: Test[] = [];
+
+  if(options.files.length) {
+    for(let file of options.files) {
+      let child = test.children.find((c) => c.path && comparePaths(c.path, file));
+
+      if(child) {
+        children.push(child);
+      } else if(fs.existsSync(file)) {
+        throw new Error(`file with path '${path.resolve(file)}' exists, but is not part of your test files, if you want to run this test, adjust the 'testFiles' setting in 'bigtest.json'`)
```
cowboyd: Might be worth printing out what the test files pattern is so when the error is encountered it can be diagnosed quickly.

packages/server/src/filter-test.ts
jnicklas: This is only partially the case. It works for this feature, since we're only filtering by test, so the set of lanes that we run actually changes. However, if we were to filter by step or assertion, then we would need to make some changes to the agent, because we currently run all of the steps/assertions in a lane, and we take this information from the manifest, not from the tree provided as part of the test run command.

packages/server/src/filter-test.ts
jnicklas: I'm unsure of what you would like for this to say. Can you suggest a wording for this?

packages/server/src/filter-test.ts
cowboyd: Ah yes. Well, it feels like the right change to make to the agent to run the lane according to the tree and only use the manifest as an index to look up steps and assertions.

packages/server/src/filter-test.ts
cowboyd: I'm mainly thinking of ways I've been bitten by this over the years. One would be trying to run `tests/foos/foo.test.ts`, but the pattern in `bigtest.json` is `tests/*/*.test.ts` that doesn't expand subdirectories. What would have helped in that instance is something like:

```
The file `tests/foos/foo.test.ts` exists, but does not match the "files" pattern 'tests/*/*.test.ts`. If you want to run this test, you will need to adjust the "files" field in bigtest.json.
```

cowboyd (approved):

packages/server/src/filter-test.ts
jnicklas: Improved error message:

Screenshot 2020-09-24 at 12 50 12

jnicklas:
🥳

Issue #446, opened 2020-08-06
Coverage reporting


cowboyd:
One of the key use-cases is to be able to collect coverage data. In JavaScript, this is done by running the tests against an instrumented build. Because BigTest makes no assumptions about the build, this part needs to be handled by the actuall app.

So if you need to change your webpack config to add instrumentation to your app bundle, then you would have to make that change yourself.

However, once you have an app url that has intrumented code, we can tell bigtest to harvest the coverage data after each test lane is run, and then merge

The convention (in javascript) which is what we're focused on is for instrumentation to write metrics to the `window.__coverage__` object which is a POJO containing all of the coverage data. We can check for this object after the lane run, and send it over the wire to the orchestrator with the `lane:end` event. As each event is received, we can [merge that lane's coverage][1] data into a total coverage object associated with the test run on that agent. The CLI can then take that coverage object and write it to disk in `.bigtest/coverage.json`

In summary

1. each lane will report its coverage
2. lane's coverage will be merged into an aggregate coverage on the `TestRunAgent`
3. CLI can write coverage data to `.bigtest/coverage.json`

[1]: https://github.com/ljharb/istanbul-merge

cowboyd:
For JavaScript code, the coverage library of note is https://github.com/instanbuljs/istanbuljs

If I understand correctly, it appears that what we need to do is to get an istanbul `CoverageMap` object from [`instanbul-lib-coverage`][1]. This is what will be generated inside each lane under the `window.__coverage__` object. Since we cannot depend on agents even having the concept of lanes, it feels most intuitive, for the agent itself to be responsible for assembling a fully merged coverage map for the test run, and only reporting the coverage at the end of its test run.

 That is, we can accumulate an agent level coverage map by merging each coverage map after each lane run. Then, after the entire test run is complete on the agent, we can send the fully merged coverage map with the `testRunAgent:end` event. The orchestrator state can store this in the `coverage` field of of the `TestRunAgentState`.

The alternative would be to send each lane coverage as it is received, but the problem is that it exposes the private structure of the agent's test run into lanes which doesn't feel like a safe assumption to make.

[...]

PR #554, opened 2020-09-21
Harvest coverage metrics

cowboyd:
Motivation
----------
> part of #446 

One of the key use-cases is to be able to collect coverage data to determine exactly how well your tests are doing, and what you might need to do to improve the reliability of it. Because BigTest makes no assumptions about how an app is the build, this part needs to be handled by the actual app. That said, once there _is_ instrumentation, BigTest can harvest the metrics for you and help you inject them into your build pipeline at the appropriate point.

Approach
----------

This PR is a first step in that process. It only handles harvesting of coverage metrics from the agent and making sure that they are sent back to to the orchestrator. It does not deal with integrating that coverage into the test run state, or writing it out to a file from the cli.

To do this, the `__coverage__` object is checked after each lane run, and if present, is merged into any prior coverage maps from previous lanes. Once the test run is completed, the coverage data is added to the `run:end`

### DIFF
--- a/.changeset/harvest-coverage-data.md
+++ b/.changeset/harvest-coverage-data.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/agent": minor
+---
+If the application under test generates a `__coverage__` object, then
+it will be harvested after each lane, and the aggregate coverage map
+will be sent back to the orchestrator inside the `run:end` event.
--- a/packages/agent/app/__coverage__.ts
+++ b/packages/agent/app/__coverage__.ts
@@ -1,0 +1,1 @@
+declare var __coverage__: unknown;
--- a/packages/agent/app/__coverage__.ts
+++ b/packages/agent/app/__coverage__.ts
@@ -1,1 +1,2 @@
+/* eslint-disable prefer-let/prefer-let, no-var */
 declare var __coverage__: unknown;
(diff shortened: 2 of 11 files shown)


jnicklas:
I'm not sure about this, the communication between the various frames is getting really very messy. It's already pretty bad, right now, and this is definitely in line with how we are approaching this problem in other places. So maybe this is not the right place to fix it?

On the other hand I'm having a really hard time understanding how all of these pieces fit together. I think we need to do some significant cleanup of the entire inter-frame communication. Maybe it's okay to merge this as is, and clean up afterwards?

cowboyd:
I'm with you in the sense that this definitely feels gross. I'm having a hard time seeing what all of the inter-frame communication use-cases are going to be. Sometimes you need to send messages that will be reported back to the orchestrator, and sometimes you need storage that is per-test run. I keep wanting to stop and refactor, but at the same time, I'm hesitant to do so because without the data of what an agent plugin will need, it's hard to say what the right abstraction is.

Some sort of picture is emerging where you have two storage cells, one scoped to the test case, and one scoped to the lane that can keep any type of data you want that are initialized at the beginning of each and passed to you on your various hooks. At the same time, I think we also want to have storage on our events that is scoped to a plugin, so for example we could have a `plugindata` on each event where. `plugindata.logevents` and `plugindata.coverage` would store all the out of band data in a single location. 

But then how does that thread through the entire stack all the way to the client? Almost like we'll need to [GraphQL Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/) to build the schema and implementation at run-time.



But these are just hunches, so really at this point, I'm wondering not _if_ we need to clean up and normalize these communication and storage protocols, but _when_.

jnicklas (approved):
Okay, approving this. Let's move forward and try to consolidate all the frame communication later.

PR #556, opened 2020-09-22
App interactor waits for application to load

### LINKED ISSUE #407: `App.visit()` interactors succeeds even if the app doesn't load.
The following test suite:

```ts
import { App } from '@bigtest/interactor';
import { Button } from './interactors';

export default test("Admin UI")
  .step("load the app", () => App.visit("/"))
  .child(
    "menu navigation", test => test
      .assertion("has a profile button", () => Button("Profile").exists())
      .assertion("has a refresh button", () => Button("Refresh").exists()))
```

Was passing the "load the app" step, and failing on the assertions by timing out, when in fact, what was happening was that latency in the devserver build step was making the iframe take a long time to load. In this case, where no valid html has been received at all, the `App.visit()` step should fail with an indicative error message like "app didn't load within the timeout period"

jnicklas:
Also adds a distinct timeout setting for this, since it's likely that the application build process is slow.

Closes #407

### DIFF
--- a/packages/interactor/src/app.ts
+++ b/packages/interactor/src/app.ts
@@ -8,14 +8,28 @@ export const App = {
       if(!appUrl) {
         throw new Error('no app url defined');
       }
-      let testFrame = bigtestGlobals.testFrame;
+      // eslint-disable-next-line prefer-let/prefer-let
+      const testFrame = bigtestGlobals.testFrame;
       if(!testFrame) {
         throw new Error('no test frame defined');
+      } else {
+        let url = new URL(appUrl);
+        url.pathname = path;
+        testFrame.src = url.toString();
+        await new Promise((resolve, reject) => {
+          let listener = () => {
+            clearTimeout(timeout);
+            testFrame.removeEventListener('load', listener);
+            resolve();
+          }
+          testFrame.addEventListener('load', listener);
+          let timeout = setTimeout(() => {
+            clearTimeout(timeout);
+            testFrame.removeEventListener('load', listener);
+            reject(new Error('timed out trying to load application'));
+          }, bigtestGlobals.defaultAppTimeout);
+        });
       }
-
-      let url = new URL(appUrl);
-      url.pathname = path;
-      testFrame.src = url.toString();
     });
   }
 }
--- a/.changeset/five-suits-sleep.md
+++ b/.changeset/five-suits-sleep.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/globals": minor
+"@bigtest/interactor": minor
+---
+
+App interactor waits for application to load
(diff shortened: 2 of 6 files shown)


jnicklas:
There is something in the tests of the `App` interactor which keeps node from shutting down properly. I spent about two hours trying to track down what it is before finally giving up. They do finish eventually, it just takes longer than it should.

Screenshot 2020-09-22 at 16 42 52

Compare the runtime 326ms with the total duration 10.84s

packages/interactor/src/app.ts
```diff
@@ -8,14 +8,28 @@ export const App = {
...
-      let testFrame = bigtestGlobals.testFrame;
+      // eslint-disable-next-line prefer-let/prefer-let
+      const testFrame = bigtestGlobals.testFrame;
       if(!testFrame) {
         throw new Error('no test frame defined');
+      } else {
+        let url = new URL(appUrl);
+        url.pathname = path;
+        testFrame.src = url.toString();
+        await new Promise((resolve, reject) => {
```
cowboyd, line 19: Rather than coordinate via on event, can we just converge on the document `readyState`?:

```ts
await converge(() => {
  if (testFrame.contentDocument.readyState !== 'complete') {
    throw new Error('timed out trying to load application');
  }
});
```

cowboyd (commented):
The application build process is slow, but we do currently [wait until the app url is reachable and returning success statuses](https://github.com/thefrontside/bigtest/blob/v0/packages/server/src/orchestrator.ts#L119-L122) before considering the server ready to run tests. Given this do you think it is still warranted to have a separate app load timeout?

It might still be, but I'm not sure. In FOLIO for example, the app bundle was 45M and so there was just a huge delay in downloading and parsing that javascript that other interactions just didn't have. So perhaps so?

packages/interactor/src/app.ts
jnicklas: I tried that, but it didn't seem to work, unfortunately. I think there is a race condition where the readyState is `complete` before actually flipping to `loading`.

Also, the polling approach of `converge` will be much more processor intensive during a time when the application will want any resources it can get, so this could also have a performance impact.

It's really telling here how useful effection is, the cleanup needed for just these TWO simulataneous async processes so they don't get messed up is pretty terrible.

jnicklas:
@cowboyd many build systems don't start building at all until a request hits the application, so the app startup in the orchestrator might be pretty fast, until that first request hits the app, and then we need to block and wait for the build process to complete. I think the kind of situation that Folio has is pretty common. I've seen this happen a lot with Capybara. Often the initial `visit` is very, very slow, because we need to wait for various things to build and load.

cowboyd (approved):

PR #558, opened 2020-09-23
Allow steps actions/assertion checks to be sync

### LINKED ISSUE #423: If a step returns an object, the runner barfs
I had the following step in my test suite:

```ts
  .step("start GraphQL simulator", () => ({
    simulator: startBrowserSimulation()
  }))
```

where `startBrowserSimulation()` returns a `Promise<Simulator>`. This was actually a bug in my code, where I should have more properly written:

```ts
  .step("start GraphQL simulator", async () => ({
    simulator: await startBrowserSimulation()
  }))
```

But as such it caused the following error in the runner:

```ts
lane.ts:38 [agent] step failed Object Error: cannot find controller for [object Object]
    at t.n.createController (context.js:217)
    at t.n.enter (context.js:96)
    at n.fork (context.js:79)
    at n.e.advance (control.js:84)
    at n.e.trapExit (control.js:102)
    at t.n.finalize (context.js:168)
    at t.n.resume (context.js:131)
    at t.call (control.js:133)
    at t.n.enter (context.js:101)
    at n.fork (context.js:79)
```

We should fix this bug, but my question is: how do we prevent this foot shot? of making sure that when you inject the resolution of a promise into a context, that you put the resolved value, not the promise... In other words, make sure you return a value that will result in:

```ts
{
  simulation: Simulation
}
```

not

```ts
{
  simulation: Promise<Simulation>
}
```

jnicklas:
Currently they must return a promise, this makes it possible for them to return sync values instead.

Closes #423

### DIFF
--- a/.changeset/lemon-boats-trade.md
+++ b/.changeset/lemon-boats-trade.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/suite": minor
+"@bigtest/agent": minor
+---
+
+Allow step actions and assertion checks to return values synchronously
--- a/packages/agent/app/run-lane.ts
+++ b/packages/agent/app/run-lane.ts
@@ -67,7 +67,7 @@ export function* runLane(config: LaneConfig) {
         originalConsole.debug('[agent] running step', step);
         events.send({ testRunId, type: 'step:running', path: stepPath });
 
-        let result: TestContext | void = yield timebox(step.action(context), stepTimeout)
+        let result: TestContext | void = yield timebox(Promise.resolve(step.action(context)), stepTimeout)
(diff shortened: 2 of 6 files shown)


cowboyd (commented):
Definitely an improvement. Since synchronous methods can raise exceptions rather than simply returning a promise that rejects, what do you think about wrapping the actions in a method that resolves to their return value and rejects with any errors they throw?

jnicklas:
@cowboyd the try/catch around it is going to catch it no matter whether it is handled as a promise or not.

cowboyd (approved):

PR #564, opened 2020-09-24
Prevent performing actions in assertions

### LINKED ISSUE #315: Prevent interactor actions when running an assertion

jnicklas:
Closes #315

### DIFF
--- a/packages/interactor/src/create-interactor.ts
+++ b/packages/interactor/src/create-interactor.ts
@@ -1,3 +1,4 @@
+import { bigtestGlobals } from '@bigtest/globals';
 import { InteractorSpecification, FilterImplementation, InteractorInstance, InteractorType, LocatorFn, InteractorConstructor } from './specification';
 import { Locator } from './locator';
 import { Filter } from './filter';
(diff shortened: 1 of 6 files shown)


packages/interactor/src/create-interactor.ts
```diff
@@ -18,8 +19,11 @@ export function createInteractor<E extends Element>(interactorName: string) {
           if(args.length) {
             actionDescription += ` with ` + args.map((a) => JSON.stringify(a)).join(', ');
           }
-          return interaction(`${actionDescription} on ${this.description}`, () => {
-            return converge(() => {
+          return interaction(`${actionDescription} on ${this.description}`, async () => {
+            if(bigtestGlobals.runnerState === 'assertion') {
+              throw new Error(`tried to ${actionDescription} on ${this.description} in an assertion, actions should only be performed in steps`);
+            }
```
jnicklas, line 25: Another option would have been to implement this in `Interactor#perform`, but I liked this option better.

packages/interactor/src/create-interactor.ts
```diff
@@ -18,8 +19,11 @@ export function createInteractor<E extends Element>(interactorName: string) {
           if(args.length) {
             actionDescription += ` with ` + args.map((a) => JSON.stringify(a)).join(', ');
           }
-          return interaction(`${actionDescription} on ${this.description}`, () => {
-            return converge(() => {
+          return interaction(`${actionDescription} on ${this.description}`, async () => {
+            if(bigtestGlobals.runnerState === 'assertion') {
```
cowboyd, line 23: Is it worth making a check for `runnerState === 'pending'`? Probably not, but at the same time it definitely would be an indicator that something is not right.

cowboyd (approved):

packages/interactor/src/create-interactor.ts
jnicklas: I thought about it, but this felt like it'd be enough.

PR #561, opened 2020-09-24
Don't watch test files when running CI command

jnicklas:
This adds an option to the bundler whether to watch the test files or not. It then threads this option through the project options. Finally we always set this option to `false` when running the CI command.

Closes #445

PR #565, opened 2020-09-24
chore: cache yarn dynamically

### LINKED ISSUE #562: Tests on Mac intermittently fail when installing dependencies
Getting these errors reasonably frequently. Oddly, this only seems to happen on the Mac actions, not on Ubuntu.

Screenshot 2020-09-24 at 14 04 41

jbolda:
This dynamically pulls the location of the yarn cache which should be more resilient to upstream changes. This also adds another restore key which should increase the odds of getting _some_ cache which should be safe with the yarn cache. See https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key for more info.

This closes #562.

### DIFF
--- a/.github/workflows/test-mac.yml
+++ b/.github/workflows/test-mac.yml
@@ -16,11 +16,15 @@ jobs:
         # keeping platform as a single item so test-*.yml files can be identical except for this value
     steps:
       - uses: actions/checkout@v2
-      - name: cache
-        uses: actions/cache@v2
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - uses: actions/cache@v2
         with:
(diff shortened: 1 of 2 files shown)


jnicklas (approved):
That set-output trick is very neat!

jnicklas:
@jbolda this just failed on the bundler tests, but this is a separate issue. That particular test seems to be flaky on macOS. I added a fix for this in #561.

jbolda:
This seems to be hitting the cache for Mac now too. Less than a minute install vs 13m in your screenshot in the issue.

jbolda:
@cowboyd we have our answer to the question about required checks. Bundler failed on Mac and passed on Ubuntu. It says all required checks must pass.

cowboyd:
@jbolda I temporarily disabled that check in order to get this through.

cowboyd:
Checks re-enabled.

PR #566, opened 2020-09-25
Add interactive init command

### LINKED ISSUE #156: Add an interactive init command to bootstrap an application

jnicklas:
This is a pretty basic init command, which only tackles the most basic configuration needed. It uses an interactive prompt created via the node `readline` package.

At first I attempted to do this using `ink`, but the results were not satisfying. Not only did it become much more complicated than it needed to be, but `ink` needs to emulate a proper terminal readline, which leads to very sub-par results. Not only do keybindings not work properly, but the resulting prompt also looks pretty janky.

If we do want to use `ink`, we can use it for those questions where it makes sense, but not for those that prompt for a string/number.

Closes #156

init


jnicklas:
Another thing we will want to add is detect whether the user is using yarn or npm.

packages/cli/src/init.ts
```diff
@@ -0,0 +1,82 @@
...
+  } catch {
+    options = {
+      port: 24000,
+      launch: ['chrome.headless']
+    };
+  }
+
+  if(!options.app) { options.app = {} }
+
+  options.port = yield prompt.number('Which port would you like to run BigTest on?', {
```
cowboyd, line 29: If this is your first exposure to BigTest it might help to provide some context.

```suggestion
  options.port = yield prompt.number('Customize the port on which BigTest runs?', {
```

cowboyd: Not sure if we want to contextutalize other responses, but other than that, I'm ready to move with this

jnicklas: I took a look at how other projects do this, e.g. `yarn`/`npm` and they don't provide any explanation or context for the various fields at all. I tried to find a balance between making this friendly for newbies while also keeping the question short and easily scannable.

cowboyd: Maybe we can go lite on the context, but I guesss the main thing to convey is that most of the time, this won't be important.

jnicklas: It could actually be important. If they are developing multiple apps using BigTest, then a port collision here would be a bit annoying. Maybe it would even make sense to randomly pick a port in a range as a default, say between 24000 and 28000 or whatever, to minimize the chance of a collision.

cowboyd: It's true, and we've actually run into this before. In fact, it might even be nice to run the range in order instead of randomly. So you always get 24000 if it's available, 24100 if not that, 24200 if not that, 24300 if not that, etc...

That would be way handier than what we have now. Question is, should we do that now?

cowboyd: We could also do a similar thing with the app and pass it into the config as a template variable:

```json
{
  "app": {
    "command": "yarn start --port ${port}",
    "url": "http://localhost:${port}"
  }
}
```
Where we'd choose 3000 by default, 3001 if that weren't available, 3002, etc... 

If they wanted to ignore the variable and just use hardcoded values, that would work too.

jnicklas: @cowboyd that's already kinda what we do, except via environment variables. The app port default to `port + 1`, and we set it as the `PORT` environment variable. This seems to be a common standard, and many frameworks will pick up that `PORT` env variable automatically. We don't have template variables right now, and I'm a bit wary of it, since it complicates the parsing of the configuration quite a bit.

cowboyd: @jnicklas Oh, I didn't realize that we already set the port environment variable. That's pretty cool.

Well then the question is, do we go ahead and draw from a port range, or use a fixed port. I'm kinda inclined to do the former which then sidesteps this entire issue and makes the configuration that much simpler.

dagda1: `react-dev-utils` do this [choosePort](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/WebpackDevServerUtils.js#L448)

But 9 times out of 10 it just means I have forgotten that the process is running on another tab or something that I have forgotten about.

So maybe a warning is enough

jnicklas:
This seems like it's good to go, aside from the discussion above about how to choose a default port and exact wording of the questions. These are improvements we can easily make in the future, if we want to make a better choice about default ports. It's worth noting that we currently have fixed defaults for all of the other ports, so making a better choice here does not buy us that much. I propose that we merge this as is, and leave these enhancements to a future PR.

cowboyd (approved):
Ok, let's go for it. I'll note this as a post-beta change https://github.com/thefrontside/bigtest/issues/576

PR #567, opened 2020-09-25
Default to launching headless chrome

jnicklas:
Our default settings currently do not specify a default for which browser to launch. While it would make sense to add an interface in the `bigtest init` command to prompt the user which browser they would like to launch (if any), actually building such an interface is a bit cumbersome, so let's punt on that.

It feels like a better default to be opinionated about a default browser to launch, rather than defaulting to not launching anything.

### DIFF
--- a/.changeset/real-toys-attend.md
+++ b/.changeset/real-toys-attend.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/project": minor
+---
+
+Default to launching headless chrome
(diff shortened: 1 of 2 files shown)


jnicklas:
Would like some feedback if this is a sensible choice or if I'm just being too lazy in not building an interface to select which browsers to launch. 😬

cowboyd:
If we're going to hardcode `chrome.headless` somewhere, wouldn't it make more sense to just hard-code it in the `init` command and write it into `bigtest.json` without prompting? That way there's a default configuration provided, but it's obvious where it's happening, and there's a place of record in the project connecting the `launch` field to the `--launch` command line option.

We can always make the command interactive in the future, no?

jnicklas:
@cowboyd that's a great idea!

PR #568, opened 2020-09-25
Reexport from `bigtest` package.

jnicklas:
This adds exports to the `bigtest` package, exporting the contents of the @bigtest/interactor and @bigtest/suite packages. This way, there is only one package to install and a simple `yarn add -D bigtest` is needed to get going.

### DIFF
--- a/.changeset/sixty-geese-mix.md
+++ b/.changeset/sixty-geese-mix.md
@@ -1,0 +1,5 @@
+---
+"bigtest": minor
+---
+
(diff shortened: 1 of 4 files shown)


jnicklas:
@cowboyd not sure if this is the right way of doing this. I think this would massively simplify using BigTest for new users.

cowboyd (approved):
😻

taras (approved):
👍

PR #571, opened 2020-09-25
Report coverage from the CLI

### LINKED ISSUE #446: Coverage reporting

cowboyd:
> closes #446 

Motivation
-----------
We need [coverage reporting][1] in BigTest, and this is the front half of the changes needed to implement it.

Approach
----------

With coverage data [being harvested][2] from the web agents, we can now take it and aggregate it on the server into a single coverage map that spans coverage of the same application over multiple browsers. This means that if you have code like the following:

```js
if (isFireFox()) {
  // some browser specific code.
}
```

The coverage metrics will change as you run begin to add different browsers to your test suite, which is a very good thing.

Once the aggregate coverage data is integrated into the atom and associated with the highest level `TestRunState`, it can be queried over graphql.

On the client end, this adds a `coverage` field to the project which has the output directory for the reports and the list of report formats to output. That way when you pass the `--coverage` option to the cli command, it knows what to do. For example, if your bigtest.json looks like:

```json
{
  "coverage": {
    "reports": ["lcov", "html"],
    "directory": "./coverage"
  }
}
```

Then running the tests with the `--coverage` option

```
$ bigtest test --coverage
```

will render both the `lcov` and the `html` report to the `./coverage` directory. Without specifically requesting coverage reports with the `--coverage` flag, the coverage data will not be requested via graphql at all.

One thing that did significantly change is the timing of when an agent run was considered complete. We had been "short circuiting" the run at the moment all of the steps and assertions were reported either passing, failing, or disregared. However, in order to make sure that we get the coverage data we cannot exit early. Instead, we have to explicitly wait for the `run:end` event to be sent by the agent since that is the event bearing the raw coverage metrics. This shouldn't be a problem, but it does pass control of "done-ness" of a test run over to the agent from the orchestrator which is a significant shift. On the whole this makes sense since the amount of metadata likely to be passed back about a test run from the agent is only likely to grow, and only the agent can determine when it is time to report that data.

Screenshots
-------------

Using the `--coverage` option on a test suite that has coverage metrics, and the reports are generated in the requested format

informational message is display showing coverage successfully reported

Using the `---coverage` option, but there are no coverage metrics generated by the test run (usually caused by the build being improperly instrumented)

A warning is show because coverage is requested, but no coverage metrics were present in the application

Open Questions
----------------

* We could add configuration flags for the `reports` and `directory` fields of the config, but this change is already significant, and while this PR adopts a "wait and see" approach, we could do either.

* I'm not super thrilled that the orchestrator and CLI are coupled directly to the instanbul coverage library since I don't know it to be a standard, and in order to generate coverage for different platforms, we'll need to make it generic. I feel like coverage is a pretty good use-case for having a plugin and it pretty much touches on every single thing where we need to have an extension point.
1. to stash cross-lane data on the agent
1. extend the run:end event to report metadata
1. to extend the orchestrator state to house the metadata
1. extend the run command to aggregate coverage data
1. extend the GraphQL schema to add the coverage field
1. extend CLI to add --coverage option
1. extend the project config to have a "coverage" section
1. add a custom reporter to handle test results and render reports to
1. report configuration errors and runtime errors to the user

If we can implement coverage and application logging events via a plugin, we've got something really powerful.

[1]: https://github.com/thefrontside/bigtest/issues/446
[2]: https://github.com/thefrontside/bigtest/pull/554

### DIFF
--- a/packages/server/test/command-processor.test.ts
+++ b/packages/server/test/command-processor.test.ts
@@ -48,7 +48,7 @@ describe('command processor', () => {
     }));
   });
 
-  describe('when sent a `run` message with a valid manifest', () => {
+  describe.only('when sent a `run` message with a valid manifest', () => {
     let pendingMessage: any; // eslint-disable-line @typescript-eslint/no-explicit-any
     let testRun: TestRunState;
 
@@ -56,6 +56,7 @@ describe('command processor', () => {
       atom.slice('bundler', 'type').set('GREEN');
       commands.send({ type: 'run', id: 'test-id-1', files: [] });
       pendingMessage = await actions.fork(delegate.receive({ type: 'run' }));
+      events.send({type: 'run:end', testRunId: 'test-id-1', agentId: 'agent-1' });
       testRun = await actions.fork(atom.slice('testRuns', 'test-id-1').once((testRun) => testRun?.status === 'ok'));
     });
 
--- a/packages/server/test/command-processor.test.ts
+++ b/packages/server/test/command-processor.test.ts
@@ -48,7 +48,7 @@ describe('command processor', () => {
     }));
   });
 
-  describe.only('when sent a `run` message with a valid manifest', () => {
+  describe('when sent a `run` message with a valid manifest', () => {
     let pendingMessage: any; // eslint-disable-line @typescript-eslint/no-explicit-any
     let testRun: TestRunState;
 
--- a/.changeset/coverage-reporting.md
+++ b/.changeset/coverage-reporting.md
@@ -1,0 +1,7 @@
+---
+"@bigtest/server": minor
+"@bigtest/cli": minor
+"@bigtest/project": minor
+---
+produce coverage reports by passing the `--coverage` option to the
+`test` and `ci` commands
--- a/packages/cli/.gitignore
+++ b/packages/cli/.gitignore
@@ -1,3 +1,4 @@
+/tmp
 /node_modules
 /yarn-error.log
 .node-version
--- a/packages/cli/bigtest.json
+++ b/packages/cli/bigtest.json
@@ -4,5 +4,9 @@
     "url": "http://localhost:36000"
   },
   "testFiles": ["./test/fixtures/*.test.{ts,js}"],
-  "launch": ["chrome.headless"]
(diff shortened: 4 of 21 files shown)


packages/server/test/command-processor.test.ts
```diff
@@ -48,14 +48,15 @@ describe('command processor', () => {
     }));
   });
 
-  describe('when sent a `run` message with a valid manifest', () => {
+  describe.only('when sent a `run` message with a valid manifest', () => {
```
jnicklas: Stray `only`

jnicklas:
```suggestion
  describe('when sent a `run` message with a valid manifest', () => {
```

jnicklas (changes requested):

jnicklas:
Totally agree that tying this to instanbul is not great at all, but at the same time I don't see a straightforward way of implementing this right now other than this kind of deep integration. Like you said, a potential plugin to do this would need to touch basically every part of our stack, and it would take a huge amount work to create all of the extension points to make it work. Since we need coverage *now* we basically have no choice but to move forward with this.

jnicklas (approved):

PR #572, opened 2020-09-26
Atom slice unique state publish

dagda1:
This is to stop consecutive identical state updates from both slice and atom:

In the logger for example, if identical bundler events are received:

```javascript
{ type: 'GREEN' }
{ type: 'GREEN' }
```

Then the responsibility of filtering identical events is pushed to the logger or any other component that is affected by similar events being raised.  

The fix is to apply a unique filter function at the subscription level of atom and slice:

```ts
export const unique = <S>(current: S) => (state: S) => {
  if (current !== state) {
    current = state;
    return true;
  } else {
    return false;
  }
}
```

```ts
  *[SymbolSubscribable](): Operation<Subscription<S, undefined>> {
    return yield subscribe(this.atom)
    .map((state) => view(this.lens)(state))
    .filter(unique(this.get()));
  }
```

This ensures that only unique state updates happen.

### DIFF
--- a/.changeset/unique-state-publish-from-atom-slice.md
+++ b/.changeset/unique-state-publish-from-atom-slice.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/atom": minor
+---
+
+only publish from atom or slice if the state has changed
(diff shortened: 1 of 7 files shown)


cowboyd (approved):
So beautiful 😍

PR #575, opened 2020-09-29
More built in interactors

jnicklas:
See #379

This adds two important interactors, `Button` and `TextField`. It also adds the `fillIn` function which can be used as an action to fill in a field, and can be re-used for the various types of field.

### DIFF
--- a/packages/interactor/src/fill-in.ts
+++ b/packages/interactor/src/fill-in.ts
@@ -1,0 +1,34 @@
+
+// do a best effort at determining the key code
+function guessCode(letter: string): string | undefined {
+  if(letter.match(/^[a-zA-Z]$/)) {
+    return `Key${letter.toUpperCase()}`;
+  } else if(letter.match(/^[0-9]$/)) {
+    return `Digit${letter}`;
+  }
+}
+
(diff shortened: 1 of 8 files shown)


packages/interactor/src/fill-in.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function fillIn(element: HTMLInputElement, value: string) {
+  let Event = element.ownerDocument.defaultView?.Event || window.Event;
+
+  element.focus();
+
+  clearText(element);
+  enterText(element, value);
+
+  element.dispatchEvent(new Event('change', { bubbles: true, cancelable: false }));
```
cowboyd: IIRC, the change event is only fired if the value is different from what it was before?

cowboyd (commented):
More interactors == more fun 😅

packages/interactor/src/fill-in.ts
jnicklas: Nice catch!

cowboyd (approved):
Looks good. I've noticed that the `cli` tests seem to be timing out with some regularity. Not exactly sure what's going on there, but it's something we need to look into.

PR #578, opened 2020-09-30
Add visibily filter using `element-is-visible`

jnicklas:
This sets up a filter for our built in interactors which checks element visibility. This filter is enabled by default, so our interactors default to only matching visible elements.

The advantage to this is that it makes our interactors more robust. It is quite common to hide elements via CSS, and while they then technically exist, they aren't usable by an actual user, and so filtering out these elements makes sense. Additionally, since we require our interactors to be unambiguous, it can sometimes help in matching an element if another match exists but is invisible.

There are some downsides, however. The first is obviously performance. Checking the visibility of every element we match is going to take up processing cycles. The algorithm in element-is-visible is pretty comprehensive and in my opinion is glaringly unoptimized in some spots.

It's worth noting though that Capybara defaults to doing this, and even actually does an additional request to fetch the visibility state for each found element sequentially, and it is *still* decently fast. I suspect that the performance hit this creates is only going to be visible in microbenchmarks, when running against an actual app it is going to get completely drowned out by the performance of the app under test.

Another downside is the nature of a filter which is enabled by default. If we do want to find an element which is hidden, we can do `Link("Foo", { visible: false })`, however there is currently no way to disable the filter entirely, that is there is no way to find an element regardless of visibility state. Capybara works around this by providing the rather awkward `Link("Foo", { visible: "all" })`. What we really want is the tri-state boolean yes/no/maybe ;)

Depends on https://github.com/thefrontside/element-is-visible/issues/1

### DIFF
--- a/.changeset/new-shoes-confess.md
+++ b/.changeset/new-shoes-confess.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/interactor": minor
+---
+
+Add visibility filter to interactors using element-is-visible
--- a/packages/interactor/package.json
+++ b/packages/interactor/package.json
@@ -24,7 +24,8 @@
   "dependencies": {
     "@bigtest/globals": "^0.7.0",
     "@bigtest/performance": "^0.5.0",
-    "change-case": "^4.1.1"
+    "change-case": "^4.1.1",
+    "element-is-visible": "^1.0.0"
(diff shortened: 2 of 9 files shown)


cowboyd (approved):
🍻

Issue #321, opened 2020-06-09
Better errors when app command fails


pittst3r:
This is what it looks like currently:

```
$ yarn test
yarn run v1.22.4
$ bigtest server
[orchestrator] starting
Error: 'yarn start' exited with non-zero exit code
    ... useless stack trace ...
error Command failed with exit code 255.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

We should surface the underlying error. In my case the error was this, which was an easy fix:

```
$ yarn start
yarn run v1.22.4
$ ts-node src/index.ts
Error: Cannot find module '/Users/rpitts/.../index.ts'
    ... useless stack trace ...
```

jnicklas:
YES! This is extremely frustrating and we have to fix it!!!

cowboyd:
Sketch of design:

appService has:

```ts
{
  stdout: { tail: string[], lastChunk: string };
  stderr: { tail: string[], lastChunk: string; };
}
```

`tail` is around 100 lines maybe? Or a hundred chunks? It's some buffer of some size.

jnicklas:
How did we feel about this, is the current solution good enough for the beta? Should we close this? Or remove it from the milestone?

cowboyd:
Yeah, I think it probably is good enough to add it for now. Output is ugly, but it does get printed to the console.

dagda1:
What is the current solution?

Could we not do something nicer.

Like in my spike?  [error](https://github.com/thefrontside/bigtest/blob/reporter/packages/reporter/src/out/console/console.ts#L110).

Not this exact code but something like it.

We could use the `ErrorDetail` type an `ErrorStack` types before it gets here.

I think we could do a quick facelift and add some colour and then think about the proper cli.

cowboyd:
Right now, we just include the output in the stack trace.

cowboyd:
@dagda1 Actually, I lied. We don't print anything 😊 

Can we take a lightest touch approach and just add a simple output to the `Logger` whenever the app transitions into the crashed state?

dagda1:
@cowboyd  how are you triggering the error?

cowboyd:
Right here https://github.com/thefrontside/bigtest/blob/graceful-agent-shutdown/packages/server/src/app-server.ts#L36

Basically we have most of  the information we need there in the exit status we just don't ever display a) the fact that the app crashed or b) any of the details around the app when it did. In fact, it would be nice to log that the app is started.

[...]

PR #580, opened 2020-09-30
abstract all output operations into @bigtest/reporter

dagda1:
aaa

This is not a complete PR, I would appreciate feedback first before proceeding.

I've not written any tests until I know that this is an agreed approach.

While looking at #321 it became apparent that we need a reporting or logging abstraction.

If we want to stream logs, buffer or output to things other than console then we need an abstraction around outputting info.

I've created a `Reporter` class

```ts
export class Reporter implements Pick<Out, LogCategory> {
  constructor(private out: Out) {}

  debug(...args: unknown[]) {
    this.out.debug(args);
  }

  info(...args: unknown[]) {
    this.out.info(args);
  }

  warn(...args: unknown[]) {
    this.out.warn(args);
  }

  success(...args: unknown[]) {
    this.out.success(args);
  }

  error(error: string | Error | BundlerError) {
    this.out.error(error);
  }

  clear() {
   this.out.clear(); 
  }
}
```

The `Reporter` uses containment and not inheritance for different output options such as a console outputter. 

The `Reporter` takes an `out` object in its constructor that can output to the console or a Buffer or a whatever.

Currently, there is only a `ConsoleOut` class that abstracts away `stdout` and `stderr`.

The reporter would have methods such as `progressBar` also for console, HTML etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "private": true,
   "workspaces": {
     "packages": [
+      "packages/reporter",
       "packages/eslint-plugin",
       "packages/effection",
       "packages/client",
--- a/packages/bundler/package.json
+++ b/packages/bundler/package.json
@@ -24,6 +24,7 @@
     "@babel/runtime": "^7.10.4",
     "@bigtest/effection": "^0.5.4",
     "@bigtest/project": "^0.9.0",
+    "@bigtest/reporter": "^0.1.0",
     "@effection/channel": "^0.6.7",
     "@effection/events": "^0.7.8",
     "@effection/node": "^0.8.0",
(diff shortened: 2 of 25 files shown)


jnicklas:
This is great, I've been thinking about something like this as well. One thing that I would really like would be the option of setting a log level on the reporter, so we can do something like `new ConsoleReporter({ logLevel: 'warn' })` and it'd only output `warn` and `error` logs.

dagda1:
@jnicklas absolutely.  Anything like that belongs behind the abstraction.

Also, good error messages and test failure messages in the console and browser are so important.

I'd definitely like @cowboyd's and anyone else's feed back before turning this into a full-blown PR.

PR #583, opened 2020-10-01
Restructure interactor definitions, remove custom locators

### LINKED ISSUE #513: Interactor specification types are too permissive
As pointed out by @dagda1.

When creating an interactor via `createInteractor`, the given options are properly type checked, but the set of options is not closed, so this is allowed, but it should not be:

``` typescript
createInteractor('foo')({ doesNotExist: "blah" });
```

### LINKED ISSUE #573: Are locators redundant?
I've been working on some built in interactors, and one thing that struck my is that custom locators feel kind of redundant to me, with some simple changes we could replace them with filters entirely. Let's look at an example:

This is how we would use a custom locator today:

``` typescript
Button.byId('my-button').click()
```

We already allow the locator to be omitted, which will find any element which matches the selector. So for example, this will match any button, regardless of its text:

``` typescript
Button().click()
```

If we combine this features with passing filters, then we can emulate the behaviour of a custom locator:

``` typescript
Button({ id: 'my-button' }).click()
```

This should match any button which has the id `my-button`, this is exactly the same as the custom locator.

Currently this is not supported, filters cannot be passed if the locator is omitted, but IMO this *should* be supported, and if we do, we might as well get rid of custom locators entirely.

There might be some slight complications with generating error messages, but I think overall this is a worthwhile simplification.

The nice effect of this is that it will simplify the interactor types quite a bit :)

Thoughts, @pittst3r @minkimcello @cowboyd?

jnicklas:
Closes #573
Closes #513

This makes a number of changes to interactors:

- Removes support for custom locators
- Rename `defaultLocator` to `locator`
- Filters can be passed to interactor even if locator is omitted, e.g. `Button({ id: "foo" })`
- InteractorSpecification is a concrete type, parameterized over its actions and filters, rather than an interface, disallows incorrect options in interactor definition
- "did you mean" suggestions always render as a table, rather than an inline comma separated list

### DIFF
--- a/packages/interactor/src/create-interactor.ts
+++ b/packages/interactor/src/create-interactor.ts
@@ -1,5 +1,5 @@
 import { bigtestGlobals } from '@bigtest/globals';
-import { InteractorSpecification, FilterImplementation, InteractorInstance, InteractorType, LocatorFn, InteractorConstructor } from './specification';
+import { InteractorSpecification, FilterParams, Filters, Actions, InteractorInstance, LocatorFn } from './specification';
 import { Locator } from './locator';
 import { Filter } from './filter';
 import { Interactor } from './interactor';
@@ -9,8 +9,8 @@ import { converge } from './converge';
 const defaultLocator: LocatorFn<Element> = (element) => element.textContent || "";
 
 export function createInteractor<E extends Element>(interactorName: string) {
-  return function<S extends InteractorSpecification<E>>(specification: S): InteractorType<E, S> {
-    let InteractorClass = class extends Interactor<E, S> {};
+  return function<F extends Filters<E> = {}, A extends Actions<E> = {}>(specification: InteractorSpecification<E, F, A>) {
+    let InteractorClass = class extends Interactor<E, F, A> {};
 
     for(let [actionName, action] of Object.entries(specification.actions || {})) {
       Object.defineProperty(InteractorClass.prototype, actionName, {
@@ -34,27 +34,20 @@ export function createInteractor<E extends Element>(interactorName: string) {
       });
     }
 
-    let result: InteractorConstructor<E, S> = function(value?: string, filters?: FilterImplementation<E, S>): InteractorInstance<E, S> {
-      let locator = value ? new Locator(specification.defaultLocator || defaultLocator, value) : new Locator();
-      let filter = new Filter(specification, filters || {});
-      let interactor = new InteractorClass(interactorName, specification, locator, filter);
-      return interactor as InteractorInstance<E, S>;
+    function initInteractor(filters?: FilterParams<E, F>): InteractorInstance<E, F, A>;
+    function initInteractor(value: string, filters?: FilterParams<E, F>): InteractorInstance<E, F, A>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    function initInteractor(...args: any[]) {
+      let locator, filter;
+      if(typeof(args[0]) === 'string') {
+        locator = new Locator(specification.locator || defaultLocator, args[0]);
+        filter = new Filter(specification, args[1] || {});
+      } else {
+        filter = new Filter(specification, args[0] || {});
+      }
+      return new InteractorClass(interactorName, specification, filter, locator) as InteractorInstance<E, F, A>;
     }
 
-    for(let [locatorName, locatorFn] of Object.entries(specification.locators || {})) {
-      Object.defineProperty(result, locatorName, {
-        value: function(value: string, filters?: FilterImplementation<E, S>): InteractorInstance<E, S> {
-          let locator = new Locator(locatorFn, value, locatorName);
-          let filter = new Filter(specification, filters || {});
-          let interactor = new InteractorClass(interactorName, specification, locator, filter);
-          return interactor as InteractorInstance<E, S>;
-        },
-        configurable: true,
-        writable: true,
-        enumerable: false,
-      });
-    }
-
-    return result as InteractorType<E, S>;
+    return initInteractor;
   }
 }
--- a/packages/interactor/types/filters.ts
+++ b/packages/interactor/types/filters.ts
@@ -72,5 +72,5 @@ TextField('foo').has({ value: 123 });
 TextField('foo').has({ blah: 'thing' });
 
 // cannot use filter on interactor which has no filters
-// FIXME: this should be disallowed!
+// $ExpectError
 Div('foo').has({ blah: 'thing' });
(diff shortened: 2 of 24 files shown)


packages/interactor/types/filters.ts
```diff
@@ -63,3 +70,7 @@ TextField('foo').has({ value: 123 });
 
 // $ExpectError
 TextField('foo').has({ blah: 'thing' });
+
+// cannot use filter on interactor which has no filters
+// FIXME: this should be disallowed!
+Div('foo').has({ blah: 'thing' });
```
dagda1, line 76: change `FilterParams` in specification to this should make it work

```ts
export type FilterParams<E extends Element, F extends Filters<E>> = keyof F extends never ? never : {
  [P in keyof F]?:
    F[P] extends FilterFn<infer TArg, E> ?
    TArg :
    F[P] extends FilterObject<infer TArg, E> ?
    TArg :
    never;
}
```

packages/interactor/src/create-interactor.ts
```diff
@@ -9,8 +9,8 @@ import { converge } from './converge';
 const defaultLocator: LocatorFn<Element> = (element) => element.textContent || "";
 
 export function createInteractor<E extends Element>(interactorName: string) {
-  return function<S extends InteractorSpecification<E>>(specification: S): InteractorType<E, S> {
-    let InteractorClass = class extends Interactor<E, S> {};
+  return function<F extends Filters<E> = {}, A extends Actions<E> = {}>(specification: InteractorSpecification<E, F, A>) {
```
dagda1, line 12: this is excellent and really cures the typing issues by making `InteractorSpecification` a concrete type.

packages/interactor/src/create-interactor.ts
```diff
@@ -9,8 +9,8 @@ import { converge } from './converge';
 const defaultLocator: LocatorFn<Element> = (element) => element.textContent || "";
 
 export function createInteractor<E extends Element>(interactorName: string) {
```
dagda1, line 11: I think this is the last domino to fall :).  If we could get the inference working here:

One way of doing it is like this:

```ts
export function createInteractor<N extends keyof HTMLElementTagNameMap>(name: N) {
  type Tag = HTMLElementTagNameMap[N];
  return function<F extends Filters<Tag> = {}, A extends Actions<Tag> = {}>(specification: InteractorSpecification<Tag, F, A>) {
    return {} as unknown as InteractorType<Tag, F, A>;
  }
}

const Link = createInteractor('link')({
  selector: 'a',
  actions: {
    click: perform(element => { element.click() }),
    setHref: perform((element, value: string) => { element.href = value })
  }
});

Link.click();
Link.setHref('http://whatever');
```
[here is a playground](https://www.typescriptlang.org/play?ts=4.0.2#code/KYDwDg9gTgLgBDAnmYcBiBLANjYUDyARgFbADGMAPACoA0cAonKLgHYAmAzo1sALbBWMAHxwAvHADeAKDhwAhmDBZEALnTZcUNKxr0GwgNyy47YADN5AVxwB+ddWMBfadNCRYCZKkw48nSiYWQS4efkERcTkAJXJodkpOGCgMVgBzel8tHUorVgBrVggAd1Z9UQAfDT8CEnIqPMKSssZhYVd3aHgkFDgAQQoMCFYcoJA2UIZeASFRCQAKVK15Cmh1AEkhPBWYaE2k+VYyYED6SSczp2F6ADo7+Sg0znVDxABtAF0ASnFRRqLSsYOuAunAyFh5JxuJtlqsoIFmOMQtwpuEhJlERNuFl-IFrv1Mcj+oNhgEDKIZHIyKTklY4fM4GArIQsBgyHBWPIBOokil0rQTD9ztIXMCPN1vHAYds4QBlFBkDDmNnyGBDXRjLFhGYwDHBDjYzS4gz0PqEg3EtWkvFzKQmTjAXhw+xwXmpNLGOQ7dWcF19T1wZU1X3qNDOMWgnqoAZW1gAWWAMAAFhAuAj9ZNphFTebQjGfTaopS4G8AApwVJwfLARAQcz9D7qPplj657jzRZbKA7NZSrs9qD7GCHY6nKQXcf4u43B5PdSpcx4ODUPqPTg-MSiBdL6ixGBWKCsL4mOS2OAd6ez57L1dPDeiXeJg9Hk9wdSsYAANzw4bcIM8UZ9rCuyDqwBxHCcmpEqiOp6kiFo4lAZL4maGbcPm1rklE0rdnCY5oKaRgRgBko4QO1DeOm8GZmiuroG21RaMhOZoZaBZYRIZF4QwmSEXAABkbHDAmyapmShFAp0JG9IhOR0IwDEwREtrzI6tHqAw97LpJ-7wOYeQkqwjJ4OY0B8FRWpKei9GsYhzEEqxGFgXi9DUAxryfMI8zmKw6iqVmQgabc9xrg4WmfhAGDsEKJgAPSxcwnCskIAC07AYJw8gssAKUfuMKXJagAACUacGQKRgDAKXAElSyxUU1XgKyipVa8JhQE+h7nksMogRs-bcbx-RTiFc7Llp8znMeopSRKvROWSikBZEEixNSUAJG6-JCSMGptDp4qBgZsZgh1qrAFxIGUAAcgx1a1vWAAS1BxgAMlZMDUPIaTXVywBxooXmctycDXTFciAV9aRRM9b0fVDv0CADYBvNdHwBh1+5dfpRyxpQaAMXZNDfba5wsdR6GGQEUOk1c8ycAqSoqrG-XAdA8rkEzZCquqxMZOghHg3IcCY8+44KNw-zNBLQG9dAFEoHzQ19ERcguKK1JgfAr2pPkUTlcA52XdA8wAOTJfkptfJN9pqXC6im-IpsCl6VPqMWVLNfk6goFAplQHw-m0b8UjMMtNzgmy+TzD8ThfC7wsOjAj0deYPsmWZHZqbBcCfvIWBWMAPLJO6WmSGHtE3EmqdRHnBeoHHJjq18QI6wUEdezHxht-kNxJynFhm0mMAwGAqjxcUSbnd+UBW4YQA)

[HTMLElementTagNameMap](https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L19263) will give you the element from a string.

I don't know enough about interactors to know if this will work but could

cowboyd (commented):
Structurally, I think this is a major upgrade. Since this is a breaking change, I'd love to see the FOLIO suite that @minkimcello developed upgraded to use the latest round of changes. That should give us a good fix on the benefits. Let's get together Monday to upgrade the test suite using the preview package?

cowboyd (approved):
Save any last minute feedback from @minkimcello, this seemed to work out very well.

minkimcello (approved):
@cowboyd @jnicklas No complaints here. I love these changes. 👍

PR #586, opened 2020-10-05
Improve formatting of stack traces

jnicklas:
Make stack traces easier to read by using better spacing and color.

Before

Screenshot 2020-10-05 at 13 46 30

After

Screenshot 2020-10-05 at 13 46 36


cowboyd (approved):
Very nice indeed.

PR #588, opened 2020-10-05
Show log events when running with `--show-log`

jnicklas:
We have all of the infrastructure for this in place, aside from actually printing them. The hardest part here is finding a design for this which is comprehensible and doesn't look like trash. Would love some feedback on this.

Screenshot 2020-10-09 at 11 40 43

The reason it prints "unknown error" is due to #589, if we fix that issue the error should be shown with its proper message and stack trace.

### DIFF
--- a/.changeset/few-rats-live.md
+++ b/.changeset/few-rats-live.md
@@ -1,0 +1,7 @@
+---
+"@bigtest/cli": minor
+"bigtest": minor
+"@bigtest/agent": patch
+---
+
+Print log output when running with --show-log
--- a/packages/agent/app/serialize-error.ts
+++ b/packages/agent/app/serialize-error.ts
@@ -64,10 +64,10 @@ export function *serializeError(error: unknown): Operation<ErrorDetails> {
       return { name: error.name, message: error.message, stack: resolvedStackFrames };
     } else {
       return {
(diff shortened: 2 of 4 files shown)


taras (commented):
Is this showing console from the browser in error feedback?

jnicklas:
@taras yes, exactly.

taras:
image

taras (approved):
👍

cowboyd (approved):
So cool! Just looking at the tree, I'm wondering if we should omit passing assertions along the path to a failed step or assertions. We could propose that change in a separate PR though.

cowboyd:
Should this be the default in CI?

pittst3r (approved):
This is great! What you have makes sense to me. We may find improvements after using it though.

pittst3r:
@cowboyd I don't think it should be default in CI because I often run `ci` locally (e.g. I want to check one last time before pushing that everything passes and I have already shut down the server), and when running locally I'm not sure we would want that to be the default. If it's the default we would need a way to turn _off_ console logs. And I think having that different default between the different commands could be confusing.

cowboyd:
@pittst3r makes sense. I guess you can just add the `--show-log` command to your CI config.

PR #590, opened 2020-10-06
Introduce Printer, an abstraction to simplify formatters

jnicklas:
Currently, all of the code in `format-helpers` is a bit of a mess, because it needs to manually keep track of prefixes, since we work very heavily with indenting code. This results in a lot of manual jiggling to get everything to print correctly. Also, there are probably a bunch of edge cases that we currently don't handle very well, where newlines show up in unexpected places.

This introduces the concept of a printer, which can write to a writable stream (usually stdout), and has an internal prefix. When we print to the printer it will prefix all printed lines. This also makes it very easy to write recursive code, or to reuse the same code in different context with different prefixes.

The resulting output looks completely identical as far as I can tell.

### DIFF
--- a/packages/cli/src/printer.ts
+++ b/packages/cli/src/printer.ts
@@ -1,0 +1,44 @@
+import * as os from 'os';
+import * as chalk from 'chalk';
+import { Writable } from 'stream';
+
+type Color = 'red' | 'grey' | 'white' | 'yellow' | 'green';
+
+export class Printer {
+  constructor(public stream: Writable, public linePrefix = '', public colorValue?: Color) {
+  }
+
+  write(...text: string[]) {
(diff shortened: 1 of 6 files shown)


packages/cli/src/printer.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  color(value: Color): Printer {
+    return new Printer(this.stream, this.linePrefix, value);
+  }
+
+  get red() { return this.color('red'); }
+  get grey() { return this.color('grey'); }
+  get white() { return this.color('white'); }
+  get yellow() { return this.color('yellow'); }
+  get green() { return this.color('green'); }
```
jnicklas, line 43: Would be nice to generate these methods automatically, but I'm not sure how to do this with TypeScript.

jnicklas:
Not sure why the publish preview is failing.

EDIT: did a `git commit --amend -CHEAD` and force push and it seems to have resolved itself 🤷‍♂️

cowboyd (approved):

PR #594, opened 2020-10-08
Fail build if there are typescript errors

### LINKED ISSUE #280: Type-check test bundle

jnicklas:
Closes #280

This adopts an approach of specifying the path to a tsconfig file. If this path is not set, we use the default tsconfig if it exists. TypeScript files need to be precompiled for the build to succeed, if they are not, compilation will fail. Unfortunately the error it fails with is a big cryptic, we can see if we can improve the error message here in a follow up.

Screenshot 2020-10-08 at 11 45 58


packages/bundler/src/bundler.ts
```diff
@@ -121,7 +128,14 @@ export class Bundler implements Subscribable<BundlerMessage, undefined> {
 
       this.channel.send({ type: 'UPDATE' });
     } catch(error) {
-      this.channel.send({ type: 'ERROR', error });
+      let message = error.message
+      if(error.loc) {
+        message += `\n\n${error.loc.file}:${error.loc.line}:${error.loc.column}`;
```
dagda1: I think you can get a better error message out of this and maybe this is not the right place to be doing this.

My take on this would be to put as much in the state and let whatever is being notified of the state change construct the error message.

I did [this](https://github.com/thefrontside/bigtest/blob/reporter/packages/reporter/src/out/console/console.ts#L129) on a spike branch which puts a nice error out to the console.

dagda1: a string message in the state gives less wriggle room.

jnicklas: Hmm, interesting. We might be able to render this error a littler nicer, on the other hand we make ourselves dependent on the format of an error generated in a completely different package. I'm a bit torn on this.

@cowboyd which option do you think is better?

dagda1: I think that error structure of `filename`, `linenumber` and `column` is pretty common even though it is coming from rollup in this instance

This [package](https://babeljs.io/docs/en/babel-code-frame) will print this nicely to the UI if you have the filename, line number and column.

It is a babel package so knows nothing about rollup.  Rollup has kindly retrieved this info for us.

cowboyd: My first inclination would be to do both. If we can provide it in a canonical string format, but also provide access to the underlying structure then we can have the best of both worlds. UI's that don't want to care have a quick fix, but those that want to take on the coupling to a richer data type are allowed to. 

If we could, for example, access the underlying errors emitted by the TS compiler, they seem like very detailed and unlikely to change. http://definitelytyped.org/docs/typescript-services--typescriptServices/classes/typescript.diagnostic.html Those seem really stable, and unlike rollup and babel, coupled to the right 3rd party library, TypeScript itself.

Alternatively, we could roll our own types for referencing source locations. That would give us the most control, but it would also mean that it was the most work for us :)

dagda1: > My first inclination would be to do both.

What a great answer!

There might be some benefit in having a generic error type and possibly generic stackframe.

The codeframe comes from a stackframe.

If you want a codeframe to display, then it will need a filename, line number and column.  That is pretty ubiquitous and unlikely to change but are called slightly different things in different packages.

jnicklas:
I've been investigating why the tests are failing on this PR and finally tracked it down to `rollup` not cleaning up after itself properly. I reported this issue to rollup here https://github.com/rollup/rollup/issues/3821

packages/cli/src/init.ts
```diff
@@ -64,6 +68,13 @@ export function* init(configFile: string): Operation<void> {
     });
   }
 
+  if(yield prompt.boolean('Do you want to set up a separate TypeScript `tsconfig` file for BigTest?', { defaultValue: true })) {
```
dagda1: one thing I would say is that when I was thinking about this, I was going to use either [typescript2's](https://github.com/ezolenko/rollup-plugin-typescript2) plugin's tsconfigDefaults or tsconfigOverride to ensure the tsconfig had the bits that were absolutely required

```ts
        tsconfigDefaults: {
          compilerOptions: {
            sourceMap: true,
            declaration: true,
            target: 'esnext',
            jsx: 'react',
          },
        },
       // or
        tsconfigOverride: {
          compilerOptions: {
            sourceMap: true,
            target: 'esnext',
          },
        },
```

that would negate the need to create a tsconfig for the user

packages/cli/test/cli.test.ts
```diff
@@ -201,15 +201,34 @@ describe('@bigtest/cli', function() {
...
+      });
+    });
+
+    describe('running the suite with type errors', () => {
+      let child: TestProcess;
+      let status: ExitStatus;
+
+      beforeEach(async () => {
+        child = await run('ci', '--test-files', './test/fixtures/typescript.broken.ts');
+        await World.spawn(child.stdout?.waitFor('[orchestrator] running!'));
```
dagda1, line 223: I just wonder if there is a better expectation we could use here.  Waiting for an exact string of a log message seems like something that could easily break, by removing an exclamation mark for example.

Could we use `Atom#once` or `slice#once` instead?

Having the build fail because a character from a log message is removed might not be obvious as to why the build has failed.

packages/cli/test/cli.test.ts
```diff
@@ -289,6 +308,9 @@ describe('@bigtest/cli', function() {
         await child.stdout.detect('Which URL do you use to access your application?');
         child.stdin.write('\n');
 
+        await World.spawn(child.stdout?.waitFor('Do you want to set up a separate TypeScript `tsconfig` file for BigTest?'));
```
dagda1: same as previous.  Could we use a better expectation than a log message?

cowboyd (commented):
Structurally, this seems sounds. From a UX perspective, I think that we should ask first the question of whether they want to use TypeScript at all first, since some folks will want to opt out, and in those cases, we won't need to bother with a tsconfig at all: something like

```
- Do you want to write your tests in TypeScript?
   -> What tsconfig file should be used to compile? (test/tsconfig.bigtest.json)
```

jnicklas:
@cowboyd we support TypeScript just fine now even without a tsconfig file, so the question isn't whether they want to use TypeScript (it'll work anyway), it's just whether they want to have their own tsconfig.

jnicklas:
@dagda1 I applied some defaults which I think make sense as you suggested in Discord.

dagda1:
Oh blast, I forgot to press the submit review button last night.  Apologies for that.

> @dagda1 I applied some defaults which I think make sense as you suggested in Discord.

Cheers, don't let anything I suggest hold up a release, I definitely don't have the full picture that you and @cowboyd have.

I do use `tsconfigDefaults` and `tsconfigOverrides` for things that I want to ensure are there in another project and it is not that controversial.

packages/cli/test/cli.test.ts
jnicklas: @dagda1 we're trying to test the CLI as a blackbox as much as possible here, so reaching into the internal state would be problematic, especially seeing as the CLI is actually running in a separate process here, so we don't even have access to this state. Pretty much all we know about this process is what it prints out, so there isn't really much else we could assert on.

dagda1: @jnicklas this is definitely just a suggestion.  Totally get the point about black box testing.

I could create a backlog ticket?  It worries me that deleting an apostrophe will fail the build.

jnicklas: @dagda1 feel free to create a ticket for it!

cowboyd:
> @cowboyd we support TypeScript just fine now even without a tsconfig file, so the question isn't whether they want to use TypeScript (it'll work anyway), it's just whether they want to have their own tsconfig.

This is definitely true, but I was more proposing it from a perspective of marketing/positioning. Because while we'll happily compile anything in the `testFiles`. Developers coming to BigTest will bring their feelings about TypeScript with them which can range from strongly positive to strongly negative. Given that the`init` command, especially in early days, is going to be focused on developers that are using BigTest for the first time to give it a spin on one of their projects. I'm suggesting that we use whatever those feelings are to draw them in and make them feel at home. 

I reckon folks will fall into roughly four buckets:

1. comfortable with TypeScript, and eager to use TypeScript tests on their TypeScript project.
2. comfortable with TypeScript, working with a JavaScript codebase, unsure if the can use TS.
3. curious about about TypeScript, but more comfortable with JavaScript.
4. more comfortable with JavaScript, and not interested in trying TypeScript for whatever reason.

For group (1), we want to say "Yes, our typescript support is 100% first class, and when you positively select `Yes`, you can feel good that your developer workflow is going to be great like you've come to expect."

For group (2), we want to say "application and associated build system is JavaScript-centric? That's totally fine, you can still write your tests in TypeScript with ZERO negative consequences"

For group (3): "You can write your tests in TypeScript, and it will be fine"

For group (4): "You don't have to use TypeScript if you don't want to, and you don't have to think about it again."

There may be more buckets, but the proposed changes in messaging was really about how to market to devs.

jnicklas:
@cowboyd that makes sense, I've added a question about TypeScript to the workflow.

dagda1 (commented):
nice job!  Looks great.  I think using `tsConfigDeafults` and `tsConfigOverrides` gives us a better certainty of what we are dealing with

cowboyd (approved):
Yassss! We're closing in on it.

PR #595, opened 2020-10-08
Catch errors when importing manifest

### LINKED ISSUE #592: Errors when loading manifest cause orchestrator to crash
When loading the manifest in the manifest builder, if there is some kind of error in a static context in the manifest file, the orchestrator will crash.

jnicklas:
Closes #592


cowboyd (approved):
Seems like a clear win 👍

PR #598, opened 2020-10-09
Improve interactor table formatting

jnicklas:
During our demo I was thinking about how the tables rendered by the interactors when something goes wrong look a bit *meh*, let's spruce them up a bit 💅

Before:

Screenshot 2020-10-08 at 22 29 46

After:

Screenshot 2020-10-08 at 22 29 52


cowboyd (approved):
Love to see it 😍

PR #602, opened 2020-10-09
🐛Workaround React > 15.6 monkey-patching input.value

### LINKED ISSUE #596: Should we work around React monkey-patching `HTMLInputElement.value`
While assembling a demonstration for BigTest using the [realworld e2e example](https://github.com/anishkny/realworld-e2e-test) I ran into a situation where React was not becoming aware of changes that the interactor was making in the DOM. While the typed text was successfully written to the input field, the application was not aware of it, and so was acting like the text field was blank.

This issue is documented here, https://github.com/cypress-io/cypress/issues/536, and we were able to apply the workaround [alluded to](https://github.com/cypress-io/cypress/issues/536#issuecomment-311694226) here. The workaround worked great, but the question is, do we account for it in the core interactors?

It was a surprising result, and definitely would have been difficult to track down without a core contributor at hand. Given that, it could be a real problem if there are a large number of applications on which our interactors won't work. React is a _very_ popular framework, and so we have to do some sort of estimation about which versions are effected so that we can understand what size of risk are we talking about. We should find out:

- [ ] Which versions of react behave in this way on a point release by point release basis
- [ ] What percentage of apps out there are within that range
- [ ] How to help anybody encountering this problem detect it, and overcome it (including just fixing it for them)

cowboyd:
> resolves #596

Motivation
-----------
In React > 15.6 an optimization implemented by monkey-patching the `value` property of HTMLInputElement. It's not exactly clear why not, but when setting the `value` property from JavaScript, it causes the DOM to be updated, but the internal React listener for the onChange event not to be fired. This results in the DOM value being correct, but the React component representing the DOM element to not be.

Approach
----------
This works around the issue by always invoking the setter from the HTMLInputElement.prototype. That way, no matter what monkey-patches are in place, the `value` property just magically appears. Similar to what happen when the browser performs the default action.

### DIFF
--- a/.changeset/fillin-react.md
+++ b/.changeset/fillin-react.md
@@ -1,0 +1,9 @@
+---
+"@bigtest/interactor": patch
+"bigtest": patch
+---
+
+Workaround the fact that React > 15.6 monkey-patches the
+HTMLInputElement `value` property as an optimization causing the
+`fillIn()` action not to work. See
+https://github.com/thefrontside/bigtest/issues/596
--- a/packages/interactor/src/fill-in.ts
+++ b/packages/interactor/src/fill-in.ts
@@ -12,7 +12,7 @@ function clearText(element: HTMLInputElement) {
   let InputEvent = element.ownerDocument.defaultView?.InputEvent || window.InputEvent;
 
   if(element.value.length) {
-    element.value = ""
+    setValue(element, "");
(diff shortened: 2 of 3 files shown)


jnicklas (approved):

jnicklas:
Great stuff!

pittst3r (approved):

PR #604, opened 2020-10-11
Improve build error formatting

jnicklas:
This is sort of a hybrid approach compared to what was discussed in #594. The `frame` part of the build error is included in the message, whereas the source file/line/column is added as a stack frame.

IMO the `frame` part doesn't really stand on its own. It's confusingly named and not formatted in a way which makes it easy to reuse. With this change we don't need to add this field to our GraphQL, which IMO feels better.

### DIFF
--- a/.changeset/purple-hornets-suffer.md
+++ b/.changeset/purple-hornets-suffer.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/server": patch
+---
+
+Improve build error formatting
--- a/packages/server/src/command-processor.ts
+++ b/packages/server/src/command-processor.ts
@@ -74,7 +74,21 @@ function* run({ id: testRunId, files }: RunMessage, options: CommandProcessorOpt
       testRunId: testRunId,
       status: 'failed',
       agents: {},
-      error: { name: 'BundlerError', message: 'Cannot run tests due to build errors in the test suite:\n' + bundler.error.message }
+      error: {
+        name: 'BundlerError',
+        message: [
+          'Cannot run tests due to build errors in the test suite:',
+          bundler.error.message,
+          bundler.error.frame,
+        ].filter(Boolean).join('\n'),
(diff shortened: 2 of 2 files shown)


dagda1:
> This is sort of a hybrid approach compared to what was discussed in #594. The `frame` part of the build error is included in the message, whereas the source file/line/column is added as a stack frame.
> 
> 
> 
> IMO the `frame` part doesn't really stand on its own. It's confusingly named and not formatted in a way which makes it easy to reuse. With this change we don't need to add this field to our GraphQL, which IMO feels better.

the reason it needs to be split and cannot be included at once is because you might want to colour the error message differently to make it stand out.  you might be doing this in html and the cli. 


for the cli i would use [@babel/code-frame](https://github.com/babel/babel/blob/master/packages/babel-code-frame/src/index.js#L123) which requires file, column and line number but for html you would do something different.

the client can decide the formatting.

dagda1:
i like the new stackframe part.  this is now good error state

dagda1 (approved):

jnicklas:
@dagda1 strangely the `frame` part already includes coloring codes, and since the innermost coloring code "wins" we can't actually format it as we want. The formatting of `frame` is really already decided by rollup.

dagda1:
> @dagda1 strangely the `frame` part already includes coloring codes, and since the innermost coloring code "wins" we can't actually format it as we want. The formatting of `frame` is really already decided by rollup.

Yes, I totally agree it should go.  That is no use in the state.

I totally missed [these error types](https://github.com/thefrontside/bigtest/blob/v0/packages/suite/src/interfaces.ts#L126).  

I think we should be using these everywhere and just transform errors into these.

A lot of the [rollup errors](https://github.com/rollup/rollup/blob/0ffbe94981a1201da851eb6001cce7bb5b8efb20/src/utils/error.ts#L74) are just a code and a message.

PR #606, opened 2020-10-13
Add checkbox interactor

jnicklas:
See #379

### DIFF
--- a/.changeset/dry-pants-share.md
+++ b/.changeset/dry-pants-share.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/interactor": minor
+"bigtest": minor
+---
+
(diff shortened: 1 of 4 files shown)


jnicklas:
The most controversial question here is probably whether it should be named `CheckBox` or `Checkbox` 😂

cowboyd (approved):
I must admit to being in the `Checkbox` camp, but my only reasoning is that the html type is `checkbox`, not `check-box`. 

Truth be told, if you hadn't invited the controversy, it would have slipped by unnoticed 😄

jnicklas:
No one voted, so I'm calling it in favour of CheckBox 😤

PR #607, opened 2020-10-13
TextField interactor matches wider range of inputs

jnicklas:
Currently out TextField interactor is pretty strict, it only matches `[type=text]` inputs, and those that do not have a type attribute at all. With this change it will also match all of the text-like fields, like `[type=email]` or `[type=password]`, it will also match an input with a completely made up type. This behaviour is more inline with how browsers display these fields.

Note that we do *not* match any of the date related fields, this is because those fields differ in their functioning across browsers, and simply entering text into them is unlikely to have the desired result. This is actually a major pain point when using Selenium, so we should avoid this problem.

### DIFF
--- a/packages/interactor/src/definitions/text-field.ts
+++ b/packages/interactor/src/definitions/text-field.ts
@@ -1,8 +1,13 @@
 import { createInteractor, perform, fillIn } from '../index';
 import { isVisible } from 'element-is-visible';
 
+const selector = 'input' + [
+  'button', 'checkbox', 'color', 'date', 'datetime-local', 'file', 'hidden',
+  'image', 'month', 'radio', 'range', 'reset', 'submit', 'time', 'datetime'
+].map((t) => `:not([type=${t}])`).join('');
+
 export const TextField = createInteractor<HTMLInputElement>('text field')({
-  selector: 'input:not([type]),input[type=text]',
+  selector,
   locator: (element) => element.labels ? (Array.from(element.labels)[0]?.textContent || '') : '',
   filters: {
     title: (element) => element.title,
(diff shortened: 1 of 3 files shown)


packages/interactor/src/definitions/text-field.ts
```diff
@@ -1,8 +1,13 @@
 import { createInteractor, perform, fillIn } from '../index';
 import { isVisible } from 'element-is-visible';
 
+const selector = 'input' + [
+  'button', 'checkbox', 'color', 'date', 'datetime-local', 'file', 'hidden',
+  'image', 'month', 'radio', 'range', 'reset', 'submit', 'time', 'datetime'
+].map((t) => `:not([type=${t}])`).join('');
```
cowboyd, line 7: Love it. What's a textfield? Well, a textfield is what it isn't...

cowboyd (approved):
Looks good. Should we add a test to make sure that it doesn't match something like `<input type="button">`?

packages/interactor/test/definitions/text-field.test.ts
```diff
@@ -16,13 +16,30 @@ describe('@bigtest/interactor', () => {
...
+    });
+
+    it('does not find non text field inputs', async () => {
+      dom(`
+        <label for="name-field">Name</label><input type="range" id="name-field"/>
+      `);
+
+      await expect(TextField('Name').exists()).rejects.toHaveProperty('name', 'NoSuchElementError');
+    });
+
```
jnicklas, line 42: @cowboyd here's a test which checks that it does not match non-text-fields. It doesn't check button in particular, but enumerating all of them seemed a bit excessive to me.

PR #611, opened 2020-10-13
`exists` and `absent` behave better when it comes to ambiguity

### LINKED ISSUE #488: `exists` and `absent` interactions should allow ambiguous matches
The `exists` and `absent` interactions currently use `unsafeSyncResolve` under the hood, just like actions do. However, this isn't ideal. Given the following markup:

``` html
<a href="/foo">Foo</a>
<a href="/bar">Foo</a>
```

We would expect an interactor like this to succeed:

``` typescript
Link("Foo").exists()
```

However it gets a bit murkier when we get to something like this:

``` html
<section title="Bar">
  <a href="/foo">Foo</a>
</section>
<section title="Bar">
  <a href="/bar">Foo</a>
</section>
```

``` typescript
Section("Bar").find(Link("Foo")).exists()
```

Should this pass or fail? It is imo not quite clear. My instinct is that it *should fail*, since `exists` refers to the link, and not the entire chain of interactors. But I think a case can be made for either option 🤔

jnicklas:
Previously, `exists` failed if there were multiple matches, now this is allowed.

Closes #488

### DIFF
--- a/.changeset/nasty-taxis-watch.md
+++ b/.changeset/nasty-taxis-watch.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/interactor": minor
+"bigtest": minor
+---
+
+Interactor methds exists() and absent() are better behaved when it comes to ambiguity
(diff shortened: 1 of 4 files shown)


cowboyd (approved):
Any chance we can add a few comments to the `resolveUnique()`, `resolveNonEmpty()`, and `resolveEmpty()` functions?

cowboyd (approved):

Issue #295, opened 2020-05-18
Refactor orchestrator internals from using Mailboxes for communication to using the Atom


cowboyd:
In the first versions of effection, there was no way to represent long running processes other than using `fork`. However, you need ways to communicate in and out of those long running processes, the simplest case being to be able to synchronize on when those services were ready (even if they are starting in parallel). To do this we supported passing a `Mailbox` into these processes, usually called a mailbox so that we could send and receive messages.

However, now we have resources that can be objects that hold operations connected to a long-running process, we can have operations to send, receive, synchronize or whatever. You can see that the BrowserManager [uses this pattern][1] for great good.

In order to accomplish this, and to unify how we treat services, the first step is to move all readieness state into the atom. We _think_ that we'd like to tread services as uniform, and so we want to take a step towards having a single type to represent services:

```ts
type ServiceStatus = 'unstarted' | 'starting' | 'available' | 'stopping ' | 'crashed';

interface ServiceState {
  id: string;
  name: string;
  status: ServiceStatus;
}

interface Service<S extends ServiceState> {
  (slice: Slice<S>): Operation<void>;  
}
 ```

This won't be a hard requirement yet, but hopefully this can be a proving ground. Each service will take a only slice to the minimal state it needs. However, if it needs to take more than one slice, that is ok for now. In the future, we can upgrade the slice method to create "virtual slices" that are not actually 100% mapped to the internal serialized state of the atom. It might be something like:

```ts
atom.slice({
  service: ['services', 'commandServer'],
  bundlerStatus: ['services','bundler','status']
});
```
The following services need to be added to the atom:

- [ ] command service
- [ ] manifest service
- [x] manifest generator
- [ ] manifest server
- [ ] connection server
- [ ] command processor
- [ ] browser manager (mailbox free, but uses promise, not state update)

[1]: https://github.com/thefrontside/bigtest/blob/d3abd7800a8e2e073f1ae4ae4ce99f9c237a3018/packages/server/src/browser-manager.ts#L20-L43

cowboyd:
I think the key here is that we need to have each service with its state in the atom https://github.com/thefrontside/bigtest/blob/v0/packages/server/src/orchestrator.ts#L105-L136

Then, we can synchronize on the atom, and not on a mailbox.

[...]

Issue #559, opened 2020-09-23
Migrate Atom to monocle-ts


cowboyd:
Ramda lenses only abide by the lens laws if you consider their own `equals` function to be the only definition of equality. However, if you throw `===` into the mix, they violate it pretty handily. Monocle does not which makes our life eaiser when it comes to publishing different versions of an Atom's state since we can just check if it is a reference to the same value.

Therefore, for performance and correctness reasons, we'd like to migrate to monocle-ts.

jnicklas:
This is done and atom has moved to effection

PR #612, opened 2020-10-13
create new Service<O> type and refactor app-server

dagda1:
This is the first part of #295 that introduces a `Service<O>` type to generalise the various services that the orchestrator starts and gets notified of status changes:

```ts
export type ServiceState<O> = {
  id: string;
  name: string;
  status: ServiceStatus;
  // TODO: we need to further constrain this to Slice(s)
  atom?: Atom<OrchestratorState>;
} & O;

export type Service<O> = {
  (options: Partial<Omit<ServiceState<O>, keyof O>> & O): Operation<void>;
};
```

I've refactored `app-service` to use the new `Service<O>` type and to transition state changes with the `ServiceStatus` discriminated union that all `Service` implmentations will use:

```ts
export type ServiceStatus =
  | { type: 'unstarted' } 
  | { type: 'started' }
  | { type: 'reachable' }
  | { type: 'unreachable' }
  | { type: 'crashed'; exitStatus: ExitStatus };
```

### DIFF
--- a/packages/server/src/effection/restartable.ts
+++ b/packages/server/src/effection/restartable.ts
@@ -2,10 +2,10 @@ import { spawn, Context, Operation } from 'effection'
 import { subscribe, ChainableSubscription } from '@effection/subscription';
 import { Slice } from "@bigtest/atom";
 
-export function *restartable<T>(slice: Slice<T, unknown>, operation: (t: T) => Operation<void>): Operation<void> {
+export function *restartable<S, A>(slice: Slice<S, A>, operation: (t: S) => Operation<void>): Operation<void> {
   let current: Context | null = null;
 
-  let subscription: ChainableSubscription<T, undefined> = yield subscribe(slice);
+  let subscription: ChainableSubscription<S, undefined> = yield subscribe(slice);
 
   for(let currentValue = slice.get();;currentValue = slice.get()) {
     if (current) {
--- a/packages/server/src/effection/restartable.ts
+++ b/packages/server/src/effection/restartable.ts
@@ -2,10 +2,10 @@ import { spawn, Context, Operation } from 'effection'
 import { subscribe, ChainableSubscription } from '@effection/subscription';
 import { Slice } from "@bigtest/atom";
 
-export function *restartable<S, A>(slice: Slice<S, A>, operation: (t: S) => Operation<void>): Operation<void> {
+export function *restartable<T>(slice: Slice<T, unknown>, operation: (t: T) => Operation<void>): Operation<void> {
   let current: Context | null = null;
 
-  let subscription: ChainableSubscription<S, undefined> = yield subscribe(slice);
+  let subscription: ChainableSubscription<T, undefined> = yield subscribe(slice);
 
   for(let currentValue = slice.get();;currentValue = slice.get()) {
     if (current) {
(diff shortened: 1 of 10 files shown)


jnicklas (approved):
This looks great 😍

packages/server/src/orchestrator/state.ts
```diff
@@ -37,19 +40,40 @@ export type TestRunAgentState = {
...
+  | { type: 'crashed'; exitStatus: ExitStatus };
+
+export type ServiceStatuses = ServiceStatus['type'];
 
-export type AppStatus = 'unstarted' | 'started' | 'reachable' | 'unreachable' | 'crashed'
+export type ServiceState<O> = {
+  id: string;
+  name: string;
+  status: ServiceStatus;
+  // TODO: we need to further constrain this to Slice(s)
```
cowboyd, line 66: I think this would be a good issue to tackle after the `monocle-ts` refactor of #559  Basically, we should make atom just a reference to a stream of values, and then everything we deal with is a slice off of that value. I've created an issue for this #616

packages/server/src/orchestrator/state.ts
```diff
@@ -37,19 +40,40 @@ export type TestRunAgentState = {
...
 
-export interface Manifest extends Test {
-  fileName: string;
-}
+export type ServiceStatus =
+  | { type: 'unstarted' } 
+  | { type: 'started' }
+  | { type: 'reachable' }
+  | { type: 'unreachable' }
+  | { type: 'crashed'; exitStatus: ExitStatus };
```
cowboyd, line 58: I'm ok merging this as-is, but I think we'll have to revisit what constitutes a service status, because these statuses are skewed towards the app services itself that is managing an external process (crashed) and that is also managing a web server (reachable, unreachable).

Take for example, the connection service (which is really an agent service) it has the webdriver processes that it is managing (which could be many), and then the individual agents that are launched from each webdriver (which could also be many) It feels like maybe those kinds of states should be pushed down into the `O` type?

packages/server/src/effection/restartable.ts
```diff
@@ -2,10 +2,10 @@ import { spawn, Context, Operation } from 'effection'
 import { subscribe, ChainableSubscription } from '@effection/subscription';
 import { Slice } from "@bigtest/atom";
 
-export function *restartable<T>(slice: Slice<T, unknown>, operation: (t: T) => Operation<void>): Operation<void> {
+export function *restartable<S, A>(slice: Slice<S, A>, operation: (t: S) => Operation<void>): Operation<void> {
   let current: Context | null = null;
 
-  let subscription: ChainableSubscription<T, undefined> = yield subscribe(slice);
+  let subscription: ChainableSubscription<S, undefined> = yield subscribe(slice);
```
cowboyd: I'm ok refactoring the types to be more expressive, but is this necessary for this change?

cowboyd (approved):
This looks great! I have some comments, but mostly to inform the work that this will be based on.

packages/server/src/orchestrator/state.ts
dagda1: @cowboyd ooooh that does sound nice.  We need the monocle-ts

packages/server/src/orchestrator/state.ts
dagda1: @cowboyd absolutely, could not agree more.  I think this will shake out as we do more.  This was just enough for this PR.

packages/server/src/effection/restartable.ts
dagda1: @cowboyd agreed and reverted.

PR #615, opened 2020-10-14
Radio button Interactor

jnicklas:
Does what it says on the tin.


cowboyd (changes requested):
On the call, the consensus was to remove the blur() interaction, and potentially the focus interaction too.

cowboyd (approved):
👍 🔘🔘🔘

PR #623, opened 2020-10-15
Don't thread actions type through filter

jnicklas:
This is not necessary and needlessly adds a generic type to a bunch of stuff that doesn't need it.


cowboyd (approved):
Diffs that only subtract are the bestest there is 😍

PR #624, opened 2020-10-15
Interactor children

jnicklas:
Allows interactors to have children, defined like this:

``` typescript
const MainNav = createInteractor('main nav')({
  selector: 'nav',
  children: {
    link: Link,
    itemLink: {
      find(interactor, item: number) {
        return interactor.find(Link(`Item ${item}`))
      }
    }
  }
});
```

And used like this:

``` typescript
MainNav().link()
MainNav().link('Foo')
MainNav().link({ href: 'http' })
MainNav().link('Foo', { href: 'http' })
MainNav().itemLink(3)
```

This enables interactors to have a relation to other interactors.

Questions:

1. Is this even a good idea at all? I personally wanted this for the `Select` interactor, so we can do stuff like `Select('Animal').option('Dog', { disabled: true }).exists()`. While this is totally possible today with `Select('Animal').find(Option('Dog', { disabled: true })).exists()`, I find that to be a bit clunky, and it would be neat if we had a cleaner API for it.
1. Is `children` an appropriate name? I could see having for example a `fieldset` interactor on a form element, which would actually fetch a parent element. What could we call it instead?
1. Is it necessary to take an interactor as an argument, or should we just assume that whatever is returned will always be passed to `find`?

### DIFF
--- a/packages/interactor/src/create-interactor.ts
+++ b/packages/interactor/src/create-interactor.ts
@@ -1,5 +1,5 @@
 import { bigtestGlobals } from '@bigtest/globals';
-import { InteractorSpecification, FilterParams, Filters, Actions, InteractorInstance, LocatorFn } from './specification';
+import { InteractorSpecification, FilterParams, Filters, Actions, Children, InteractorConstructor, InteractorInstance, LocatorFn } from './specification';
 import { Locator } from './locator';
 import { Filter } from './filter';
 import { Interactor } from './interactor';
@@ -9,8 +9,8 @@ import { converge } from './converge';
 const defaultLocator: LocatorFn<Element> = (element) => element.textContent || "";
 
 export function createInteractor<E extends Element>(interactorName: string) {
-  return function<F extends Filters<E> = {}, A extends Actions<E> = {}>(specification: InteractorSpecification<E, F, A>) {
-    let InteractorClass = class extends Interactor<E, F, A> {};
+  return function<F extends Filters<E> = {}, A extends Actions<E> = {}, C extends Children = {}>(specification: InteractorSpecification<E, F, A, C>): InteractorConstructor<E, F, A, C> {
+    let InteractorClass = class extends Interactor<E, F, A, C> {};
 
     for(let [actionName, action] of Object.entries(specification.actions || {})) {
       Object.defineProperty(InteractorClass.prototype, actionName, {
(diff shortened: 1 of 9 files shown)


packages/interactor/src/specification.ts
```diff
@@ -13,23 +13,37 @@ export type FilterObject<T, E extends Element> = {
   default?: T;
 }
 
+export type ChildObject<T extends Array<any>, E extends Element, F extends Filters<E>, A extends Actions<E>, C extends Children> = {
+  find: (interactor: Interactor<Element, {}, {}, {}>, ...args: T) => InteractorInstance<E, F, A, C>
+}
```
jnicklas: This has to be an object so we can disambiguate it from the interactor constructor, which is a function.

packages/interactor/src/create-interactor.ts
```diff
@@ -34,8 +34,25 @@ export function createInteractor<E extends Element>(interactorName: string) {
...
     }
 
-    function initInteractor(filters?: FilterParams<E, F>): InteractorInstance<E, F, A>;
-    function initInteractor(value: string, filters?: FilterParams<E, F>): InteractorInstance<E, F, A>;
+    for(let [childName, child] of Object.entries(specification.children || {})) {
+      Object.defineProperty(InteractorClass.prototype, childName, {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        value: function(...args: any[]) {
+          if(typeof(child) === 'object') {
+            return child.find(this, ...args);
```
jnicklas, line 42: We could change this to the following, see question (3):

``` typescript
return this.find(child.find(...args))
```

cowboyd:
Maybe call it `finders` itstead of `children`? finders seems to ring with filters, and also in the case where there are related HTML that isn't necessarily DOM children it would be more accurate. I'm specifically thinking about related Modal dialogs that are logically related to an element, but have a completely different DOM ancestry.

jnicklas:
We decided on a call to leave this alone for now as an idea, but not integrate it at this point in time.

cowboyd:
Ran into this again recently writing an interactor for a `Tooltip` component. The case was that there was a Tooltip floating somewhere in the DOM containing the tooltip text, usually in some floating layer. And then there was a `ProximityTooltip` that was always there inside the DOM for screen readers and strongly associated with the element it was tooltipping. It would have been nice to somehow relate these two interactors either to each other, or under a single tooltip:

```ts
await Tooltip("Enter Text").proximity.is({ visible: true });
await Tooltip("Enter Text").floating.is({ visible: false });

await trigger.hover();

await Tooltip("Enter Text").floating.is({ visible: true });
```

That's just a spit ball, although it could be done with using just a simple function now.

jnicklas:
I still think this is a decent idea, but this implementation is no longer relevant given the new builder syntax.

PR #627, opened 2020-10-15
Catch unexpected exceptions and offer to report them as bugs

### LINKED ISSUE #618: No naked crashes
Crashes will happen, but when they do, we should definitively report to the user that this is a crash and not expected in any way, and that they should report this as an error

cowboyd:
> closes #618 

Motivation
----------
It's a tough look whenever something terrible happens and the server or the CLI halts without doing the thing that you were asking it to do. For example, we were having an issue with our proxy server crashing on canceled requests with:

```
Error: unexpected end of file
    at Zlib.zlibOnError [as onerror] (zlib.js:180:17) {
  errno: -5,
  code: 'Z_BUF_ERROR'
}
```

It's even worse when that happens, and it silently makes a terrible impression on our users and we never even have the notification that it happened.

Approach
----------
This change catches all top-level exceptions that are _not_ effetion `MainError`s and prints out an explanation and generates a link to a new github issue with several diagnostic  values pre-filled in.

Screeshots
-----------

run bigtest command that crashes then open ticket

TODOS & Open Questions
--------------------------

- [x] I'm not sure the best way to test this. I could, of course unit test the `warnUnexpectedErrors()` middleware directly, but It's kind of a small test given something that is so over-arching in its effect. If there were some way to booby-trap the CLI so that by running the `bigtest throw` command, you could have it raise an arbitrary exception from the CLI. Ideally, we could use a plugin architecture inside the test case to exend the CLI with a `throw` command, but that seems like a lot of work. Still, having our thing that is there to give us a good look even when we look bad, makes us look even worse.... well that is really bad. So not having this really well tested feel dangerous
- [x] The way that this was implemented was to catch all non `MainError`s, but that felt not so great. It occurred to me that what might even be better would be to have the `Promise` returned from @effection/node` always resolve on both successful exits _and_ on `MainError` exits since these are still _expected_ outcomes. Only on an unexpected error would the promise actually reject. This would have the benefit of allowing custom error handling of unexpected exits, but baked in handling of all expected exits. In other words, we would have been able to write this as:
```ts
import { CLI } from './cli';
import { main } from '@effection/node';
import { unexpectedExit } from './unexpected-exit';

main(CLI(process.argv.slice(2)))
  .catch(unexpectedExit);
```

This just makes the unexpected exit code a lot easier to write since it is just a function, not an operation, and it doesn't need to know anything amout `MainError`

### DIFF
--- a/.changeset/no-naked-errors.md
+++ b/.changeset/no-naked-errors.md
@@ -1,0 +1,8 @@
+---
+"@bigtest/cli": patch
+"bigtest": patch
+---
+
+When an unexpected error happens in the CLI, catch it, let the user
+know it is our fault, and generate a link to a github issue containing
+diagnostic information
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -57,7 +57,8 @@
     "istanbul-lib-coverage": "^3.0.0",
     "istanbul-lib-report": "^3.0.0",
     "istanbul-reports": "^3.0.2",
-    "json5": "^2.1.3"
+    "json5": "^2.1.3",
+    "terminal-link": "^2.1.1"
   },
   "volta": {
     "node": "12.16.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -40,6 +40,7 @@
     "@typescript-eslint/parser": "^2.3.2",
     "eslint-plugin-prefer-let": "^1.0.1",
     "expect": "^24.9.0",
+    "jest-mock": "^26.6.0",
     "mocha": "^6.2.2",
     "ts-node": "*"
   },
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1,4 +1,6 @@
 import { CLI } from './cli';
 import { main } from '@effection/node';
 
-main(CLI(process.argv.slice(2)));
+import { warnUnexpectedExceptions } from './warn-unexpected-exceptions';
+
+main(warnUnexpectedExceptions(CLI)(process.argv.slice(2)));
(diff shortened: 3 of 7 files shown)


jnicklas:
Very cool!

Two small tweaks we could do:

1) We could use something like https://github.com/sindresorhus/terminal-link so we don't have to show a gigantic link in the terminal.

2) We could use a `<details>` tag to hide the stack trace in the issue template, this will make the resulting issue look much neater.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/sindresorhus/terminal-link">sindresorhus/terminal-link</a></strong></div><div>Create clickable links in the terminal. Contribute to sindresorhus/terminal-link development by creating an account on GitHub.</div></blockquote>

cowboyd:
@jnicklas done! Those are pretty cool techs. I did notice that including html comments inside the github link as I had it was breaking the `terminalLink` function.

jnicklas:
@cowboyd hmm, I tried this out, and the link didn't render for me, so I tried to figure out what's going on, and then realized that unfortunately terminal links are not supported in tmux, so tmux will just show this as plain, unlinked text :( They have an open PR for it, but it seems to be stalled. https://github.com/tmux/tmux/pull/2403

cowboyd:
What do you think w.r.t. testing?

jnicklas:
How about just unit testing this thing? That should be fine, right?

One issue that occurred to me is what happens in case of a port collision? We should handle those somewhat gracefully, imo, and not throw a screaming error in case it happens, because it really isn't an innternal failure.

cowboyd:
> One issue that occurred to me is what happens in case of a port collision? We should handle those somewhat gracefully, imo, and not throw a screaming error in case it happens, because it really isn't an innternal failure.

I agree. We're totally able to handle that, even if it means that we log an error and exit.

cowboyd:
Oops, forgot to commit a file

jnicklas:
How about a fix for `EADDREINUSE` should we put that in there?

cowboyd:
@jnicklas I created a ticket so that we wouldn't forget https://github.com/thefrontside/bigtest/issues/651 but there is not much chance that I'll get to it today. Can we merge this as-is and then add that fix later?

jnicklas (approved):
Let's go with this for now and we can fix the port issue later.

PR #629, opened 2020-10-16
Page interactor

### LINKED ISSUE #582: Check global page properties via Page/App interactor
Tracking this here after the discussion in discourse yesterday.

There are some properties of a page which we might want to test but which are not directly related to any element on the page, some examples:

- page title
- url (can be further broken down into its various components)
- text (regardless of location)

One possibility is to make the app interactor a proper interactor and implement these as filters:

``` typescript
await App.has({ title: "blah" });
```

A more appropriate name in this case might be `Page`, rather than `App`. Maybe that's where we could move `visit` as well, in that case.

``` typescript
await Page.has({ title: "blah" });
await Page.visit("/foo");
```

jnicklas:
Closes #582


packages/interactor/src/resolve.ts
```diff
@@ -7,7 +7,13 @@ const defaultSelector = 'div';
...
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function findMatches(parentElement: Element, interactor: Interactor<any, any, any>): Match<Element, any>[] {
-  let elements = Array.from(parentElement.querySelectorAll(interactor.specification.selector || defaultSelector));
+  let elements;
+  if(interactor.specification.selector === ':root') {
+    // this is a bit of a hack, because otherwise there isn't a good way of selecting the root element
+    elements = [parentElement.ownerDocument.querySelector(':root')];
+  } else {
+    elements = Array.from(parentElement.querySelectorAll(interactor.specification.selector || defaultSelector));
+  }
```
jnicklas, line 16: To clarify *why* I had to do this: for an interactor without a parent, the `parentElement` will be the HTML element, and there is no good way in CSS to select the element *itself* via a selector. We'd essentially want to do something like `parentElement.querySelector('self')`. The closest is the selector `:scope`, which is roughly meant to do this, but support for it is spotty. JSDOM doesn't support it and neither does IE.

The alternative is to change the type of `Element` to be `Element | Document` instead, and use the document as the root instead. This would be a more elegant solution, but it would make the already complicated types even more complicated and probably add a bunch of useless typing stuff and if statements, just for this one exception.

cowboyd (approved):
👍 we may want to extract our deprecation helper soon so that we can issue deprecations like these without causing an enormous about of console spew.

Issue #632, opened 2020-10-19
Create more robust effection websocket abstraction


cowboyd:
Right now, if a websocket ever closes with anything other than a [status code](https://tools.ietf.org/html/rfc6455#section-7.4.1) of `1000` (normal), or `1001` (peer shutdown), the express server [completely blows up][1] and there is no way to change this behavior.

To say that this strategy doesn't scale is perhaps a bit of an understatement. For example, we need to be able to handle temporary disconnects and reconnects without setting the world on fire.

Instead, I propose we create an effection-friendly websocket abstraction to be more like the new process API where the subscription is a stream of events and the closing value of the subscription is the [CloseEvent][2] raised when a websocket is closed on the remote end. That way, any consumer can handle the socket stream in its entirety however it sees fit, including throwing an exception depending on the exit code.

So the new API would be something like:

```ts
export type WebSocketIO = string | ArrayBuffer | ArrayBufferView;

export interface WebSocketResource<In extends WebSocketIO, Out extends WebSocketIO> extends Subscribable<Out, CloseEvent | ErrorEvent> {
  send(message: In): void;
  close(code?: number, reason?: string);
}
```

Because the return value is an `CloseEvent | ErrorEvent` low-level consumer will be forced to deal with the possibility that the socket closed with an error (this is really rare in our experience though. More often than not, you get a `1006` which means the connection closed without the close frame being sent when the user kills chrome for example.

We could then replicate the old behavior in the consumer by throwing an error if `CloseEvent.code` was anything other than 1000 or 1001. But the key difference is that the error would be in the _consumer_ of the socket subscription, not in the socket resource that lives in the express server. Different consumers of the same socket can treat errors the same way.

[1]: https://github.com/thefrontside/bigtest/blob/837a463029964145f06f73444a8cfb41acd78e2f/packages/effection-express/src/index.ts#L59-L61
[2]: https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent

PR #639, opened 2020-10-19
Graceful agent shutdown

### LINKED ISSUE #453: Disconnecting agent causes server to crash
Here is a log of connecting an agent (simulated iPad Pro), and then closing that agent with a hard kill (CMD-Q).

```
[connection] connected agent.1
[orchestrator] shutting down!
Error: websocket client closed connection unexpectedly: [1006]
    at /Users/cowboyd/Code/EBSCO/ui-courses/node_modules/@bigtest/effection-express/dist/index.js:53:31
    at Generator.next (<anonymous>)
    at /Users/cowboyd/Code/EBSCO/ui-courses/node_modules/effection/dist/effection.js:1:8094
    at e.r.advance (/Users/cowboyd/Code/EBSCO/ui-courses/node_modules/effection/dist/effection.js:1:7730)
    at e.r.trapExit (/Users/cowboyd/Code/EBSCO/ui-courses/node_modules/effection/dist/effection.js:1:8056)
    at t.n.finalize (/Users/cowboyd/Code/EBSCO/ui-courses/node_modules/effection/dist/effection.js:1:5567)
    at t.n.resume (/Users/cowboyd/Code/EBSCO/ui-courses/node_modules/effection/dist/effection.js:1:4921)
    at e.r.advance (/Users/cowboyd/Code/EBSCO/ui-courses/node_modules/effection/dist/effection.js:1:7776)
    at e.r.trapExit (/Users/cowboyd/Code/EBSCO/ui-courses/node_modules/effection/dist/effection.js:1:8056)
    at t.n.finalize (/Users/cowboyd/Code/EBSCO/ui-courses/node_modules/effection/dist/effection.js:1:5567)
```

cowboyd:
> fixes #453 

Motivation
-----------
As mentioned in #632, The effection-express hard fails on any socket close that isn't 1000 or 1001, but this makes it impossible to handle several valid shutdown scenarios (like when an agent closes because the browser tab was closed)

Approach
----------
This makes the effection-express websocket subscription actually return the close event rather than fail if it's not a "clean close" so that socket handlers can do what they please.

The other key change is that the websocket server _no longer_ waits until the socket is closed before shutting down the operation. This allows the handler to have its own handling of the close event (since before, the context was being shutdown before the child was actually seeing the close event) If you think about it, this makes sense. If you have a handler that wants to return _before_ the socket is closed by the other end, then it can now do so. It's now up to the handle itself to wait until the appropriate time to exit.

Eventually, it will probably be worth it to have a low-level socket abstraction in effection proper that encodes these ideas as well as makes it possible to work with binary data.

### DIFF
--- a/packages/server/src/connection-server.ts
+++ b/packages/server/src/connection-server.ts
@@ -35,12 +35,11 @@ export function* createConnectionServer(options: ConnectionServerOptions): Opera
         }
       });
 
-      let close: CloseEvent = yield subscribe(connection.events).forEach(function*(message: TestEvent) {
+      let { code, reason }: CloseEvent = yield subscribe(connection.events).forEach(function*(message: TestEvent) {
         console.debug('[connection] got message from agent', connection.agentId, message);
         options.delegate.send({ ...message, agentId: connection.agentId });
       });
 
-      let { code, reason } = close;
       console.debug(`[connection] disconnected ${connection.agentId} [${code}${reason ? `: ${reason}` : ''}]`);
 
       agent.remove();
--- a/packages/server/src/connection-server.ts
+++ b/packages/server/src/connection-server.ts
@@ -35,10 +35,13 @@ export function* createConnectionServer(options: ConnectionServerOptions): Opera
         }
       });
 
-      yield subscribe(connection.events).forEach(function*(message: TestEvent) {
+      let close: CloseEvent = yield subscribe(connection.events).forEach(function*(message: TestEvent) {
         console.debug('[connection] got message from agent', connection.agentId, message);
         options.delegate.send({ ...message, agentId: connection.agentId });
       });
+
+      let { code, reason } = close;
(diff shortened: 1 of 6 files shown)


packages/server/src/connection-server.ts
```diff
@@ -35,11 +35,14 @@ export function* createConnectionServer(options: ConnectionServerOptions): Opera
         }
       });
 
-      yield subscribe(connection.events).forEach(function*(message: TestEvent) {
+      let close: CloseEvent = yield subscribe(connection.events).forEach(function*(message: TestEvent) {
```
jnicklas: Why not just `let { code, reason }: CloseEvent = `?

jnicklas (approved):

packages/server/src/connection-server.ts
cowboyd: Excellent point!

PR #640, opened 2020-10-20
Add `Select` and `MultiSelect` interactors

### LINKED ISSUE #620: Select interactor
Might need to be multiple interfaces like `Select` and `MultiSelect`. However, we should consider how to message failures when the `multiple` attribute is different than what the interactor class is expecting.

jnicklas:
Closes #620

### DIFF
--- a/packages/interactor/src/definitions/multi-select.ts
+++ b/packages/interactor/src/definitions/multi-select.ts
@@ -22,11 +22,29 @@ const SelectOption = createInteractor<HTMLOptionElement>('option')({
         dispatchInput(select);
       }
     }),
+    select: perform((element) => {
+      let select = getSelect(element);
+
+      if(!element.selected) {
+        element.selected = true;
+        dispatchChange(select);
+        dispatchInput(select);
+      }
+    }),
+    deselect: perform((element) => {
+      let select = getSelect(element);
+
+      if(element.selected) {
+        element.selected = false;
+        dispatchChange(select);
+        dispatchInput(select);
+      }
+    }),
   },
 });
 
-export const Select = createInteractor<HTMLSelectElement>('select box')({
-  selector: 'select:not([multiple])',
+export const MultiSelect = createInteractor<HTMLSelectElement>('select box')({
+  selector: 'select[multiple]',
   locator: (element) => element.labels ? (Array.from(element.labels)[0]?.textContent || '') : '',
   filters: {
     title: (element) => element.title,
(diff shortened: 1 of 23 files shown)


cowboyd (commented):
Because we're in a situation where you have two selectors, I wonder if there is a need to have special error messaging around using a select for a nearly matching multi-select and vice versa.

I'm imagining a scenario where you see the select box in the UI, so you reach for a `Select` interactor, but much to your consternation, not only does it not match, it doesn't show up in the list of "nearly" matching elements that have the same selector, but different locators.

jnicklas:
I thought about this. The only way I can see of making this work at the moment is to make `multiple` a filter with a default value. This would give a better error message, at the expense of making it possible to override the value, which would be a bit odd. Do you think that's a better solution?

cowboyd:
hmmm... That might be doable. We could write a higher order interactor constructor that does not have a multiple filter, and then adds it to an underlying one?

Another option would be to provide a hook for the interactor to transform the default alternatives into an extended, filtered or mapped list of alternatives in the event of a miss.

packages/interactor/src/definitions/multi-select.ts
```diff
@@ -0,0 +1,77 @@
...
+});
+
+export const MultiSelect = createInteractor<HTMLSelectElement>('select box')({
+  selector: 'select[multiple]',
+  locator: (element) => element.labels ? (Array.from(element.labels)[0]?.textContent || '') : '',
+  filters: {
+    title: (element) => element.title,
+    id: (element) => element.id,
+    valid: (element) => element.validity.valid,
+    value: (element) => element.selectedOptions[0]?.label || '',
```
jnicklas: This is incorrect

cowboyd (changes requested):
Need to fix bug with the `MultiSelect.value`. It should be an array.

cowboyd (approved):
Resolved that multi-select is going to be very, very rare since why not checkbox? It's not used, so having this distinction is only going to be niche anyway.

jnicklas:
@cowboyd fixing the MultiSelect issue turned out harder than I thought, could you give this another look? I had to change our notion of equality for filters, so I tried to use lodash's `isEqual`, but unfortunately it doesn't seem to be possible to use this package without `esModuleInterop: true`, for some reason I don't really understand. I recall that we wanted to switch this to `true` anyway, I just changed it for the interactor package and it seems to work fine, but maybe this is not the way?

jnicklas:
Also cc @dagda1, since I think you had some opinions on `esModuleInterop`.

cowboyd:
@dagda1 added this to our TS config almost a month ago and it was released as version 1.1 

https://github.com/thefrontside/javascript/pull/16

jnicklas:
Great, we can remove the flag then once we upgrade all of the other packages.

PR #641, opened 2020-10-20
Allow steps to have duplicate descriptions

### LINKED ISSUE #622: Support duplicate step descriptions
When implementing #506, we arrived at the conclusion that it *should* be allowed for step descriptions to be duplicates within a given step. However, we currently do not support this fully. In our reporting we should ensure that this works properly.

jnicklas:
Include the step index in its path, this way we can report results even if step descriptions are duplicates.

Closes #622

### DIFF
--- a/.changeset/proud-plums-retire.md
+++ b/.changeset/proud-plums-retire.md
@@ -1,0 +1,7 @@
+---
+"@bigtest/agent": minor
+"@bigtest/server": minor
+"bigtest": minor
+---
+
+Handle duplicates in step descriptions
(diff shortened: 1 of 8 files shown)


jnicklas:
One downside of this is that the step index is sometimes visible. For example when using the lines formatter:

Screenshot 2020-10-20 at 14 37 41

I've thought about various ways of fixing this, but they all seem pretty hacky to me. I actually don't think this is so bad, so to me this feels acceptable 🤷‍♂️

cowboyd (approved):
When I see it in the screen shot, it's not so bad at all. As long as it's not shown in the tree view, then that works for me too.

PR #645, opened 2020-10-20
Add `focused` filter for asserting on focus state

cowboyd:
Motivation
-----------

When trying to implement the [todomvc test suite][1] using bigtest, the very first assertion is that the todo input field has focus, which made it immediately apparent that where focus is depending on various actions is quite important for usability and accessibility. Questions like "where is the focus when I open an aplication url?" or "where does the focus go when I navigate to a particular route?" are critical to be able to ask.

Approach
----------
This adds a `focused` filter to all of the focusable elements (link, button, checkbox, radio, textfield)

Open Questions
----------------

- [x] I named this `focused` to follow the convention of `disabled` and it reads better with an `is()` clause rather than a `has()` clause, but I could see an argument to name the filter `focus` since a document can only have but one focus and `has({ focus: true })` would work as well as `is({ focus: true })` if you read `focus` as **the** focus. in other words:
  - `is({ focus: true })`: "element is _the_ focus"
  - `has({ focus: true })`: "element has _the_ focus"

[1]: https://github.com/tastejs/todomvc/blob/master/tests/test.js

### DIFF
--- a/.changeset/focused-filter.md
+++ b/.changeset/focused-filter.md
@@ -1,0 +1,7 @@
+---
+"@bigtest/interactor": minor
+"bigtest": patch
+---
+
+Add `focused` filter to `Button`, `RadioButton`, `CheckBox`, `Link`,
+and `TextField` interactors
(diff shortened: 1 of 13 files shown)


jnicklas (approved):
Looks great! I personally prefer `focused` because IMO it reads better when expressed as a finder, but it doesn't matter really.

``` typescript
TextField('Name', { focused: true })
// vs
TextField('Name', { focus: true })
```

cowboyd:
Awesome, let's go with `focused` then since that was what my gut was telling me too.

PR #648, opened 2020-10-22
Show app errors on test run if application server exits prematurely

### LINKED ISSUE #321: Better errors when app command fails

### LINKED ISSUE #642: Should we really be checking app reachability so often?
I don't think 100ms is an appropriate interval to check app reachability. If we do this at all, after the app has started, which IMO is pretty questionable, we should at least decrease the interval quite significantly. Something like every 5s seems much more reasonable to me.

jnicklas:
If the app server exits for any reason, we should show the output of it so it's easy to debug what went wrong.

There are many ways of implementing this. We decided on adding the check at the beginning of the test run and allowing the orchestrator to boot fully even if the app server has exited. This way we open up for a more advanced handling of the app server in the future. For example we could attempt to restart it on every test run if it has exited.

We could even lazy-load it and not attempt to start it until the first test run is initiated, this would make our orchestrator start significantly faster. We could also do the the same for the bundler.

I also adjusted the possible values for the state of the app server: crashes has been renamed with the more neutral exited, since we don't actually know why the server exited, it could have just exited with `0`, for example if the user forgot to add `--watch` or something. The unreachable state was removed entirely, since it interacted badly with the `exited` state, effectively overwriting its value, and therefore the `exitStatus` information not being available when we want to print the error.

Closes #321
Closes #642

Screenshot 2020-10-22 at 09 52 25

### DIFF
--- a/.changeset/cyan-pandas-think.md
+++ b/.changeset/cyan-pandas-think.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/server": minor
+"bigtest": minor
+---
+
+Show app errors on test run if application server exits prematurely
--- a/packages/cli/test/cli.test.ts
+++ b/packages/cli/test/cli.test.ts
@@ -268,6 +268,23 @@ describe('@bigtest/cli', function() {
         expect(child.stdout?.output).toContain('⨯ FAILURE')
       });
     });
+
+    describe('running the suite with app errors', () => {
+      let child: TestProcess;
+      let status: ExitStatus;
+
+      beforeEach(async () => {
+        child = await run('ci', '--app-command', 'yarn node doesnotexist.js');
+        status = await child.join();
+      });
+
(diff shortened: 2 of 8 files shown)


cowboyd (approved):
Love it.

PR #649, opened 2020-10-22
Throw non-fetch errors when checking reachability

### LINKED ISSUE #643: App reachability swallows all errors
When [checking if app is reachable][1], we swall _every_ error which means if there is something that happens in which the fetch operation itself is invalid, we'll be guaranteed never to detect that exception condition, such as an invalid URL.

We need to change this to only catch `FetchEror` like the following code in @bigtest/webdriver: https://github.com/thefrontside/bigtest/blob/v0/packages/webdriver/src/until-url-available.ts#L21-L23

[1]: https://github.com/thefrontside/bigtest/blob/v0/packages/server/src/app-server.ts#L53-L60

jnicklas:
like for example an invalid URL

Closes #643

### DIFF
--- a/.changeset/yellow-actors-study.md
+++ b/.changeset/yellow-actors-study.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/server": patch
+"bigtest": patch
+---
+
+Throw non-fetch errors when checking reachability, like for example an invalid URL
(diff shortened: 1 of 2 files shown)


cowboyd (approved):

cowboyd:
@jnicklas looks good, but I wasn't able to re-run the CLI tests for some reason 🤷

Issue #552, opened 2020-09-21
Enhancing error reporting


minkimcello:
![Screen Shot 2020-09-21 at 12 03 12 PM](https://user-images.githubusercontent.com/29791650/93802770-e7186280-fc11-11ea-8bbe-09c77eea4546.png)

1. When we're running a large number of tests, we might see a variety of failing tests. Those errors might exceed the maximum number of lines allowed by your terminal. Even if it did not, it's a lot of scrolling up to get to the top of the error summary. So I don't know what the criteria should be but depending on the number of test files detected, maybe instead of listing all the individual errors, we can just list all the files (or even just the directories if the number of files is also too many) so that users know where to look. Out of however-many test files, if they know which files returned failing steps/assertions, the user could run those files in isolation to have a closer look.

2. Sometimes when I see a failing test, I want to quit the run immediately and tackle that failing test before proceeding.  So when a user stops Bigtest before it finishes running all of the tests, is it still possible to have it tell us which tests failed before the process was shut down?

minkimcello:
@pittst3r I created the issue here.

@jnicklas I'm not sure if this is related to the server or another component so I'll let you label it.

cowboyd:
Seems like two issues, but a `--fail-fast` option is a feature of almost every test runner (including GH actions)

jnicklas:
One thing I've always wanted in a test runner is for it to print the errors as soon as they happen, rather than waiting until the end. `--fail-fast` always kind of misses the mark for me, I don't care whether the suite fails quickly, I want it to *print* quickly.

As for the problem of exceeding the scroll limit: I quite like RSpec's solution of printing a summary of all of the failures at the end, so you can easily see which tests failed. I'm not sure how that'd fit our model though.

PR #650, opened 2020-10-22
Add maxwidth for interactor table columns

jnicklas:
See #552

### DIFF
--- a/.changeset/lemon-roses-type.md
+++ b/.changeset/lemon-roses-type.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/interactor": patch
+"bigtest": patch
+---
+
+Add maximul column with for interactor alternative suggestion tables
--- a/packages/interactor/src/format-table.ts
+++ b/packages/interactor/src/format-table.ts
@@ -3,13 +3,23 @@ export interface TableOptions {
   rows: string[][];
 }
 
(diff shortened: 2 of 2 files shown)


jnicklas:
Why not just use an already created ascii table library? I've looked at a few of them, and they just don't seem particularly good. All of them are overcomplicated, and I haven't seen a single one which allows you to set a maximum width, the one feature we'd actually like to have.

.changeset/lemon-roses-type.md
```diff
@@ -0,0 +1,6 @@
+---
+"@bigtest/interactor": patch
+"bigtest": patch
+---
+
+Add maximul column with for interactor alternative suggestion tables
```
cowboyd, line 6:
```suggestion
Add maximum column with for interactor alternative suggestion tables
```

cowboyd (commented):
How do we handle overflow if a value is longer than the column width?

jnicklas:
The value is just truncated with `…` if it overflows the column width.\

Screenshot 2020-10-22 at 17 33 08

cowboyd (approved):
w00t!

.changeset/lemon-roses-type.md
cowboyd: @jnicklas is this what you meant? ☝️

PR #655, opened 2020-10-23
make manifest-generator a service and remove Mailbox

dagda1:
Crossing off the manifest-generator from #295 and removing the `Mailbox` delegate in the process.

We might actually want to think about changing the `ServiceStatus` union

```ts
export type ServiceStatus =
  | { type: 'unstarted' }
  | { type: 'started' }
  | { type: 'reachable' }
  | { type: 'exited'; exitStatus: ExitStatus };
```

This is more `app-server` specific and seems less appropriate for others.


packages/interactor/test/definitions/link.test.ts
```diff
@@ -101,7 +101,7 @@ describe('@bigtest/interactor', () => {
     });
 
     describe('filter `focused`', () => {
-      it.only('filters `a` tags by if they are focused', async () => {
+      it('filters `a` tags by if they are focused', async () => {
```
jnicklas, line 104: How did these guys get in there?!

packages/server/src/assertions/assert.ts
```diff
@@ -0,0 +1,18 @@
...
+    Object.setPrototypeOf(this, new.target.prototype);
+  }
+}
+
+export function assert(condition: unknown, msg: string): asserts condition {
+  if (!condition) {
+    throw new ServiceAssertionError(msg);
+  }
+}
+
```
jnicklas: I don't see this actually used anywhere?

packages/server/src/orchestrator/atom.ts
```diff
@@ -7,6 +7,11 @@ interface OrchestratorAtomOptions {
 
 export const createOrchestratorAtom = (options?: OrchestratorAtomOptions) => {
   let atom = new Atom<OrchestratorState>({
+    manifestGenerator: {
+      id: '@bigtest/manifest-genertor',
```
jnicklas: I find it a bit confusing to name this as though it were a package, when it doesn't exist as a package.

packages/server/src/orchestrator/atom.ts
```diff
@@ -7,6 +7,11 @@ interface OrchestratorAtomOptions {
...
+      name: '[manifest generator]',
```
jnicklas: Why both an `id` and `name`?

dagda1: @jnicklas the `name` would be for logging and the id is a unique string for the service.

Neither is currently used so could definitely remove the id until a need arises.

Just spotted I spelt `generator` wrong in the `id` :)

cowboyd: @jnicklas It was I that recommended using an id like `@bigtest/foo` for the services as some sort of "plugin-ish" nomenclature so that we could get a taste of what it feels like to travel a bit on that path. What if we went with something like '@bigtest/server#manifest-generator` as the id? That would allow for multiple services per package, and right now, everything is owned by the `@bigtest/server` which is an actual package.

packages/server/src/manifest-generator.ts
```diff
@@ -39,37 +36,48 @@ module.exports = {
...
 }
 
-export function* createManifestGenerator(options: ManifestGeneratorOptions): Operation {
-  if(options.watch) {
-    let watcher = chokidar.watch(options.files, { ignoreInitial: true });
+export const manifestGenerator: Service<ManifestGeneratorOptions> = function *(options) {
+  let { atom, files, mode } = options;
 
-    yield ensure(() => watcher.close());
+  assert(atom, 'No atom in manifestGenerator options');
```
cowboyd: If an atom is required, why not make it required at the type level?

packages/server/src/manifest-generator.ts
```diff
@@ -39,37 +36,48 @@ module.exports = {
...
+      serviceStatus.update(() => ({ type: 'reachable' }))
     }
+
+    yield spawn(on(watcher, 'add').forEach(writeOperation));
+    yield spawn(on(watcher, 'unlink').forEach(writeOperation));
+    
+    yield subscribe(fileChanges).forEach(() => writeManifest(options));
   } else {
     yield writeManifest(options);
     console.debug("[manifest generator] manifest ready");
```
cowboyd: Is this now redundant since our sync-block will print this out?

dagda1: @cowboyd the eventual goal is for these `console.debug` statements just implicityly happen.  

What is the `sync-block` you mention here?

cowboyd: I mean the big pile 'o fork that waits for all services to be ready in `orchestrator.ts`

dagda1: oh, you mean the cutlery.

`console.debug` removed

packages/server/src/manifest-generator.ts
dagda1: atom is now required

packages/server/src/orchestrator/atom.ts
jnicklas: `@bigtest/server#manifest-generator` is definitely better, IMO.

I do kind of want to advocate for YAGNI though in that if we don't know what we are going to use these for, and we don't currently actually use them, then I don't think it's a good idea to include them at all.

jnicklas:
I would like to see a reconsideration of how the service status enum works in this PR already. I don't like using `reachable` and `unreachable` in the manifest generator. It feels like leaving the work in refactoring this half done. I don't like including a refactoring which takes us from "correct but not generalized" to "incorrect but generalized".

packages/server/src/orchestrator/state.ts
```diff
@@ -67,7 +69,7 @@ export type ServiceState<O> = {
 } & O;
 
 export type Service<O> = {
-  (options: Partial<Omit<ServiceState<O>, keyof O>> & O): Operation<void>;
+  (options: O & Required<Pick<ServiceState<O>, 'atom'>>): Operation<void>;
 };
```
jnicklas: This also feels to me like having something which is halfway done. It's adding bandaids on a type to force it into doing kind of what we want, rather than finding a good generalized solution. If it can't easily be generlized, then maybe it shouldn't be.

dagda1: This comment is a statement.

In order to resolve this, what needs to happen?  There is no indication.  This could be another PR that goes nowhere but I'll wait a few weeks to find that out.

Even better than that, how could we have found this out before arriving at this point after the fact?  

Should the types have been agreed upfront?  

It is a wasted effort on my part to reach this point after.

packages/server/src/orchestrator/atom.ts
cowboyd: I'm definitely 👍 to that. Although `id` and `name` are already part of the interface, so we'd have to remove it elsewhere, right? Also for the name, seems like we should remove the `[]`

packages/server/src/orchestrator/state.ts
cowboyd: I think @dagda1 is trying to not rush towards an abstraction as that's something that has stalled out PR's in the past. 

I think everyone is 100% not satisfied with service state carrying over artifacts of the app server and proxy server which are very much http service oriented things.

I'm happy to have that discussion now before work proceeds. I think we can model it very simply with

`unstarted`, `starting`, `running`, `stopping`, `stopped` as the discreet states:

```ts
type ServiceState<TOptions, TRunning, TStopped> = 
{
  status: 'unstarted';
} | 
{
  status: 'starting',
  options: TOptions;
} |
{ 
  status: 'running';
  options: TOptions;
  state: TRunning;
} |
{ 
  status: 'stopping';
  options: TOptions;
  state: TRunning;
} | 
{
  status: 'stopped';
  toptions: TOptions;
  state: TRunning;
  stop: TStopped;
}
```

The nice thing about this is that error and crash exits can be modeled with a custom `TStop` for example:

```ts
type AppServiceState = ServiceState<AppOptions, AppState, ExitStatus>
```

In this case, `ExitStatus` represents an application exit status, but it could be something different for a different service. Honestly though, it seems like in this case, the `AppState` is what should contain the concept of "reachability" and "exited", and so the app _service_ is ever-running, even though the _app_ might crash.

dagda1: I've absolutely no problem with this being queried or objected to, I don't really like the union TBH but was happy to just see how it played out in subsequent PRs under the guise of destroying the Mailboxes.

Getting it right the first time is difficult without fully understanding the services.

I think my point is that this is potentially a showstopper six days after raising the PR with no clear way to resolve it and then the PR just drifts due to our truly remote nature across timezones.

I've always felt git diffs are poor code reviews, I find it hard to get the big picture and only really notice small things.

I totally actually agree with the type being not up to scratch but I was going with the premise that it is good enough for now.

cowboyd: I can tell you're frustrated, but let's not call the PR adrift until it's actually adrift. It's true that it sat for most of last week while we were really laboring to close out remaining bugs for our beta milestone, but the moment that there was bandwidth this week, it got picked up and review started.

Still, it does suck to feel like the goal-posts have gotten moved to an undisclosed location. So how about this, what if we just didn't generalize over service states at all and say that the first step is to just get the state in the atom in the first place and use a one-off API for each service. I think that would be a clear win.

dagda1: definitely not frustrated with the points raised.  All good points

Just that it took a while for it to surface

I think in future, I would create a typescript playground of the types and put it in discord before progressing.

jnicklas: @dagda1 I'm sorry, reading this again you're totally right, my comment was not very actionable, and doesn't help in closing out this PR. I should have been clearer in my feedback.

I don't see this as stalled at all though, barring a few minor changes, I think this is pretty close. I do like this PR a lot, I'm very excited about this new service architecture, it's a very cool use of Effection, and I think it's going to be way nicer than what we have now.

As for getting timely feedback I don't work on the weekends, so the initial review I did was as timely as I could have given it, the feedback that I posted yesterday was after looking at this more thoroughly, and spotting some issues I hadn't seen before.

The heart of my comment here is that we started off with all of the service interfaces being pretty specific to the app server, which is fine, since it was the only service that we had. But once we move towards adding more services, IMO we should try finding an abstraction which works for all of them.

Reading my comment above it reads much more broadly than what I meant. I was specifically talking about the state of a service, where we've tried to create an abstraction trying to generalize what the state of a service looks like, but maybe this is not easily generalized, because the various services that we have have pretty different needs as to the state that they should maintain.

I've slightly modified the playground link you posted in Discord to illustrate what I mean:

https://codesandbox.io/s/typescript-playground-export-forked-zgx3n?file=/index.ts

In summary:

``` typescript
type AppStatus = { status: "pending" } | { status: "ready" } | { status: "exited"; exitStatus: ExitStatus };
type AppOptions = { url?: string; command?: string; };

interface ServiceStatus {
  status: string;
}

type ServiceState<O, S extends ServiceStatus> = {
  id: string;
  status: S;
  options: O;
};

type Service<O, S extends ServiceStatus> = {
  (status: Slice<S>, options: O & { atom: Slice<OrchestratorState> }): void;
};

type OrchestratorState = {
  appService: ServiceState<AppOptions, AppStatus>;
};

const appService: Service<AppOptions, AppStatus> = (status, options) => {
  status.set({ status: "starting" });
  console.log(options.atom.get().other);
};
```

The key difference here is that we don't try to create an abstraction over the status of various services, but allow them to specify their own status options. I think this makes this whole thing much easier to reason about, and makes it easy to see a path forward for this.

I think it could be sensible to add some constraint to the `status` in the future, maybe `pending` and `ready` or whatever we want to call them, should be defined for all services. Or maybe just `ready`? But it feels like we can probably hold off on this constraint until we know more about how we want to use services. I think just going with convention here is fine for now.

I also made some other minor modification which I think are nice, but these are just suggestions:

- Passing in the status slice directly to the Service, since modifying this will be a key part of the service.
- Removing the atom from the state (since that would be circular), and instead only adding it to the options passed to the service. The exact interface here is just a suggestion.
- Keeping the `options` separate, instead of in the same object makes it easier to extract them and pass them to the service.

dagda1: @jnicklas no need to apologise.  My reply could definitely have been worded better as I wrote with an air of frustration :).

At no point did I disagree with the points raised.   I'm in agreement with pretty much all of them

> Passing in the status slice directly to the Service, since modifying this will be a key part of the service.

@cowboyd and I did think about this but then thought about the case where you might need more than one slice.  I think I should have waited and crossed this bridge when I encountered it.

> Removing the atom from the state (since that would be circular), and instead only adding it to the options passed to the service. The exact interface here is just a suggestion.

This is actually the way it was but my type used `Omit` and other shenanigans to ensure that it was passed into the service but not in the state.

I think the type read terribly in retrospect so I need to make that clearer.

> Keeping the options separate, instead of in the same object makes it easier to extract them and pass them to the service.

I think this makes sense.

dagda1: > Or maybe just ready

I was thinking along these lines also.  We might just need:

```ts
type ServiceStatus = 'IDLE' | 'READY'; 
```

This is good enough for now. 

cc/ @cowboyd

packages/server/src/assertions/assert.ts
dagda1: Sorry, I've added it to `ManifestGeneratorOptions`.  It type narrows

```ts
  if(!options.files) {
    throw new Error('no files options in ManifestGeneratorOptions');
  }

  if(!options.destinationPath) {
    throw new Error('no files options in ManifestGeneratorOptions');
  }
```

to

```ts
  assert(options.files, 'no files options in ManifestGeneratorOptions');
  assert(options.destinationPath, 'no destinationApth in ManifestGeneratorOptions');
```

packages/interactor/test/definitions/link.test.ts
dagda1: fixed in a merge to `v0`

packages/server/src/orchestrator/atom.ts
dagda1: I have removed the `id` and `name` fields until we actually know what their use is.

packages/server/src/orchestrator/atom.ts
dagda1: I've removed the `id` until we have a clear use for it.

packages/server/src/orchestrator/state.ts
dagda1: @jnicklas @cowboyd I agree with the suggestion that each service having its own status is a better idea and have implemented this.

```ts
export type ServiceStatus = {
  type: string;
};

export type ServiceState<S extends ServiceStatus, O> = {
  options: O;
  status: S;
};

export type Service<S extends ServiceStatus, O> = {
  (status: Slice<S, OrchestratorState>, options: O & { atom: Atom<OrchestratorState> }): Operation<void>;
};
```

jnicklas (approved):
Looks great!

packages/server/src/orchestrator/state.ts
```diff
@@ -50,26 +49,19 @@ export type BundlerState =
...
+export type ServiceStatus = {
+  type: string;
+};
 
-export type Service<O> = {
-  (options: Partial<Omit<ServiceState<O>, keyof O>> & O): Operation<void>;
+export type ServiceState<S extends ServiceStatus, O> = {
+  options: O;
+  status: S;
 };
```
cowboyd, line 59: When in comes to the Service State I think that the guiding principle should be that the intrinsic properties of the state should be those that are managed by the orchestrator, and those portions of the state that vary are the ones about which only the particular service has knowledge. If we use that as a heuristic, then what does the orchestrator know?

It knows 1) when it has not even started the service operation as in it has not yet invoked it. It knows 2) it has invoked the service operation and 3a) when the service operation has completed or 3b) when the service operation has raised an error. This would seem to imply that the intrinsitc statuses are (for example):

`unstarted`, `started`, `completed`, `errored`

> Worth noting that `completed` is a useful piece of information since I've definitely been bitten by a service operation running to completion too quickly because of spawning a bunch of things and then not having an actual blocking operation.

Everything else can only be known internally as the service operation runs and so falls under the purview of the service itself, and so it's _that_ state that should form the basis for `S`. It does seem odd not to have a `ready` state in the above list, but readiness is something that can only be known by the service. We could make an exception by having the orchestrator pass a ready callback so that while technically it is managing that property, it is only via inversion of control.

Then, the service actually does not receive the service state, but only a slice of `S` and everything else would be managed by the orchestrator. Under that regimen, `S` would be completely unconstrained. 

I'd propose that the services are invoked with their options,  S is unconstrained, and that services are run in a uniform way by the orchestrator with a `manageService` operation (which will eventually restart the service when the options slice changes). Something like:

```ts
export interface ServiceState<S, O> {
  status = 'unstarted' | 'started' | 'completed' | 'errored';
  state: S;
  options: 0;
  error?: unknown;
}
```

That's the endgame that I think we're shooting for (with option of having a `ready()` callback). I think that we've got mostly that here, so I would just rename `ServiceStatus` to something like `props`, `attrs`, or `state` to indicate that it is completely owned by the service.

dagda1: @jnicklas @cowboyd I think this is slightly going back where I was 2 commits ago :).

I think we need to somehow reach an agreement with what the right thing is without writing code.  Not sure how we do that.

a small minor note, is that `Error` should be a generic type argument.  Then that covers the case of `ExitStatus` or `xhrError` or plain error or whatever.

```ts
export interface ServiceState<S, O, E = Error> {
  status = 'unstarted' | 'started' | 'completed' | 'errored';
  state: S;
  options: 0;
  error?: E;
}
```

PR #656, opened 2020-10-24
Spike to parse cucumber files into bigtest object

dagda1:
I originally did this on a fork and I very nearly deleted it before remembering about this spike.

Just leaving it here until it is ready to be picked up.


cowboyd:
Let's go ahead and keep this branch around, but close the PR so that we have a clean PR slate. It's likely we'll pick this up in the future, but no plans to merge this any time soon.

PR #657, opened 2020-10-27
Update getting started instructions

jnicklas:
It is no longer necessary to install sub-packages.

[Rendered](https://github.com/thefrontside/bigtest/blob/getting-started/README.md)


cowboyd (changes requested):
Let's change it from `App` interactor to `Page` interactor while we're at it

cowboyd (approved):

PR #659, opened 2020-10-27
Start adding docs via typedoc

jnicklas:
This is a first stab at adding docs via TypeDoc.

Unfortunately this uses a frozen in time, outdated fork of TypeDoc, found here https://github.com/TypeStrong/typedoc/pull/1184, because the output generated just isn't very nice, but is pretty great with this fork 🤷‍♂️

It seems like some other projects are also doing this. Not sure what this says about TypeDoc.

It also takes some finagling to get docs for interactors to display in a comprehensible way.

Screenshot 2020-10-27 at 14 41 38
Screenshot 2020-10-27 at 14 41 46


jorgelainfiesta:
How can I preview the docs in a browser, like what command should I run? I tried browsing TypeDoc's website but I couldn't figure it out.

Gerrit0:
Recommendation: switch to typedoc@beta, which is version 0.20.0-beta.#, and is under active development. I need to update the linked PR to point people towards that...

taras:
Thank you @Gerrit0!

jnicklas:
Thanks @Gerrit0, I upgraded to the beta and it works perfectly.

@jorgelainfiesta I don't have any good way of putting these docs online right now. You can run `yarn typedoc` in `packages/bigtest`, and it will output the generated docs into `packages/bigtest/docs`.

jnicklas:
We'll probably want to add a deploy hook for these!

jnicklas:
This is by no means done, but I think it's good enough now where we can actually merge this and start working on integrating these docs into the website.

Docs can be generated for the `@bigtest/suite`, `@bigtest/interactor` and `bigtest` packages, by running `yarn typedoc` in their respective directories.

packages/bigtest/API.md
```diff
@@ -0,0 +1,3 @@
+# BigTest
+
+This is cool stuff {@link createInteractor}, yeah!
```
jnicklas, line 3: This file is just a placeholder, we'll want to expand this to link to the most relevant parts of the documentation.

cowboyd (approved):

PR #662, opened 2020-11-03
Initial Documentation setup

jorgelainfiesta:
## Approach

**So far I**:
- Ran @docusaurus/init@next init template
- Moved Min's docs to the folder structure used by docusaurus
- Adjusted the sidebar with Min's current docs
- Removed the blog

**Pending**:
- [x] Custom styles
- [x] Deployment along-side API and homepage.


website/docs/docs/getting-started/1-installation.md
```diff
@@ -0,0 +1,19 @@
...
+title: Installing BigTest
+slug: /
+---
+
+The `bigtest` package includes everything you need to get started sogo ahead and install BigTest to your project:
+```
+$ yarn add -D bigtest
+```
+
+Once that's finished installing, you can run the following command:
```
jorgelainfiesta: Where should I run this? is it the within the project that I want to test or is it arbitrary?

website/docs/docs/getting-started/2-your-first-test.md
```diff
@@ -0,0 +1,41 @@
+---
+id: your-first-test
+title: Writing your first test
+---
+
+To demonstrate how you can write tests, let's install the TodoMVC app from BigTest:
+```
+$ yarn add -D @bigtest/todomvc
```
jorgelainfiesta: When we install this, is it to the project initiated by BigTest?

website/docs/docs/getting-started/1-installation.md
```diff
@@ -0,0 +1,19 @@
...
+```
+$ yarn bigtest init
+```
+
+This command will help generate the `bigtest.json` file for you.
```
jorgelainfiesta: Is generating the bigtest.json the only thing that happens with this command?

website/docs/docs/getting-started/2-your-first-test.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+Assuming you went with the default settings for `bigtest.json`, the app command should be configured as `yarn start`:
+```json
+{
```
jorgelainfiesta: It this json the default content from `bigtest.json`?

website/docs/docs/getting-started/2-your-first-test.md
```diff
@@ -0,0 +1,41 @@
...
+  "app": {
+    "command": "yarn start",
+    "url": "http://localhost:3000"
+  }
+}
+```
+Go into your `package.json` file and modify the start script as `yarn bigtest-todomvc 3000`:
```
jorgelainfiesta: I think I'm lost here, but perhaps because I don't know to whom this package.json belongs to. Might be clarified from the question above of where this is happening.

website/docs/docs/getting-started/2-your-first-test.md
jorgelainfiesta: @cowboyd is there a way we can extract `todomvc` from the monorepo?  I think it'd be better if we renamed it `accessible-todomvc` for example, instead of linking it to bigtest directly. In my opinion, the current setup smells fishy because you're using a bigtest app to test bigtest. Makes me wonder if it'll run on any other app or not.

Plus, we hint a lot of stuff on the `accesible-todomvc` name: that we care a11y, that your app must follow good practices to use bigtest

website/docs/docs/getting-started/2-your-first-test.md
```diff
@@ -0,0 +1,41 @@
...
+Finally, let's go into the `test/` directory and create `todomvc.test.ts`:
+```js
+import { Heading, Page, test } from `bigtest`;
+
+export default test('bigtest todomvc')
+  .step(Page.visit('/'))
+  .assertion(Heading('todos').exists());
+```
+
+That's it! Now you can run `yarn bigtest ci` and you should see one passing step and one passing assertion.
```
jorgelainfiesta: I'm a bit confused about the relationship between bigtest cli because we had to modify somebody's package json to make a `start`.

jorgelainfiesta (commented):
Nice work so far! I left a few comments on the Getting Started section.

website/docs/docs/getting-started/1-installation.md
minkimcello: Both commands can be run from anywhere in the project. I assume it does a thing where it searches up the directory until it hits package.json because you can run `yarn bigtest init` from `./test/` or even `./node_modules` and it'll create/modify `bigtest.json` at the root of the project.

I'll need to confirm how it works in a monorepo.

website/docs/docs/getting-started/2-your-first-test.md
minkimcello: I don't think bigtest is initiating a project. I think the assumption is people would be adding bigtest _to_ their pre-existing project.

So that being said, we're adding the todomvc app as a dev dependency (presumably to their pre-existing app) just so they can run bigtest against it. It's for demonstration purposes only and people should remove/delete it once they're done following along and understand how it works.

All they would need to do is remove the dependency of todomvc and modify the start script in package json.

website/docs/docs/getting-started/1-installation.md
minkimcello: Yeah. And the command will modify the bigtest.json file if it already exists. And if the bigtest.json file already exists, running bigtest init will make suggestions based on what your previous configurations are.

For example, when you run `bigtest init` for the first time (and there's no bigtest.json file), the default value for the app port will be 3000. But if you already have bigtest.json configured to run the app in 4000, running bigtest init will then go "Where would you like to run your application? 4000?"

website/docs/docs/getting-started/2-your-first-test.md
minkimcello: No, I omitted some of the properties of the full default config file. If you run `bigtest init` and hit enter all the way through (so it's only default suggested values), this is what it would look like:
```json
{
  "port": 24000,
  "launch": [
    "chrome.headless"
  ],
  "app": {
    "command": "yarn start",
    "env": {
      "PORT": 3000
    },
    "url": "http://localhost:3000"
  },
  "testFiles": [
    "test/**/*.test.{ts,js}"
  ]
}
```

website/docs/docs/getting-started/2-your-first-test.md
minkimcello: Okay so to give a better idea of what I was trying to do, I made some diagrams:

Say I have `Min's Website` and I want to try out bigtest. Normally I would install bigtest and configure it to my website. The red arrow indicates configuration.
Screen Shot 2020-11-09 at 9 24 25 AM

The idea is we want to guide them step by step as much as possible but if we were to make a guide that works for any app/website, we have to be more vague. We can't help them write test for a headings at the top of the website because some websites for some reason might not have a heading. So that's why we're slapping on the todomvc app on `Min's Website` so people can practice writing test directly from their project.

Another strategy would be to just have readers clone the todomvc app and write tests for it from the todomvc repo directly. But I think people want to dive into it as soon as possible. Meaning when there's a tool and the guide tells you to `git clone ourTools-tutorial.git`, I end up skipping it and just reading the documentation. But if the tutorial lets you put it on your project as soon as possible, it's more encouraging. We can discuss that theory further.

> The summary of the last two paragraphs: put it in your actual repo asap = "okay let's do this!". make people clone an example repo for practice = "... eh maybe i'll do it some other time. too much of a hassle."

So then when people add the todomvc app to their repo, it looks like this:
Screen Shot 2020-11-09 at 9 25 25 AM

Instead of configuring bigtest to Min's Website, we're just configuring it to todomvc app. And once people go through the guides, we can instruct them to remove todomvc app and fix the start script in their package.json to react-scripts start or ng serve or whatever it is they're using.

Screen Shot 2020-11-09 at 9 26 23 AM

website/docs/docs/getting-started/2-your-first-test.md
minkimcello: Between bigtest cli and?

So there were two ways to run todomvc app instead of the reader's app: modify the start script of package.json or modify the app.command in bigtest.json. I chose to go with modifying package.json because we all know how that works.

Once they're done with the tutorial, it'll be easier for someone to realize that the start script needs to be re-adjusted for their own app versus going into bigtest.json and realize that the app.command in bigtest.json is connected to the start script in package.json.

If the guide tells them to change the app.command script to `yarn bigtest-todomvc 3000`, they might think it's a bigtest-related command that needs to stay in their configuration.

website/docs/docs/getting-started/2-your-first-test.md
jorgelainfiesta: It seems a bit counterintuitive to install an entire app in your app to try out a framework, but the rationale you provide is very strong. Perhaps we can just make clear from the beginning that that's the approach that we'll take, and even add a small note on the fact that they can also clone the project if it makes them feel more comfortable. 

I can try to write that down so you can focus on new content, I'm asking to understand what you want to say before I bring it to the editor and such. With this context that you provided now I can write out some more details on the guide. Thanks!

website/docs/docs/getting-started/2-your-first-test.md
jorgelainfiesta: Ok great info, thanks! clarifies a lot

cowboyd (approved):
Let's roll 🙌🏻

minkimcello (approved):

PR #663, opened 2020-11-03
simplify the passing of options and status into a service

dagda1:
I was experiencing a lot of friction between passing config into the 2 services we have so far and this PR is to simplify this before refactoring any more code.

For example, the `manifestGenerator` currently shares the same properties when `createOrchestratorAtom` is called and when the service is created in `orchestrator`:

```ts
    manifestGenerator: {
      status: { type: 'pending' },
      options: {
        mode: 'idle'
      }
    },
```

and in orchestrator:

```ts
  yield fork(manifestGenerator(manifestServiceStatus, {
    destinationPath: manifestSrcPath,
    atom: options.atom,
    mode: options.project.watchTestFiles ? 'watch' : 'build',
    files: options.project.testFiles
  }));
```

This is further complicated with `strictNullChecks` in tsconfig.json.   It is also annoying to have to pass the `atom` in when creating the service.

After the refactoring in this PR, the above looks like this:

```ts
  let appServerState = options.atom.slice('appService');

  yield fork(appServer(appServerState));

  let manifestGeneratorState = options.atom.slice('manifestGenerator');

  yield fork(manifestGenerator(manifestGeneratorState));
```

The whole slice is passed into the service which should make the `orchestrator` much clearer as we create more services.

The adding of the configuration has been pushed into `createOrchestratorAtom`:

```ts
export const createOrchestratorAtom = (project: OrchestratorAtomOptions) => {
  let manifestSrcDir = path.resolve(project.cacheDir, 'manifest/src');
  let manifestSrcPath = path.resolve(manifestSrcDir, 'manifest.js');

  let atom = new Atom<OrchestratorState>({
    manifestGenerator: {
      status: { type: 'pending' },
      
      options: {
        destinationPath: manifestSrcPath,
        mode: project.watchTestFiles ? 'watch' : 'build',
        files: project.testFiles
      }
    },
    appService: {
      status: { type: 'pending' },
      options: project.app,
    },
```

### DIFF
--- a/packages/server/src/app-server.ts
+++ b/packages/server/src/app-server.ts
@@ -2,19 +2,21 @@ import { timeout, spawn } from 'effection';
 import { fetch } from '@effection/fetch';
 import { exec, Process } from '@effection/node';
 import * as process from 'process';
-import { OrchestratorState, AppOptions, Service, AppServiceStatus } from './orchestrator/state';
+import { OrchestratorState, AppOptions, Service, AppServiceStatus, ServiceState } from './orchestrator/state';
 import { Slice } from '@bigtest/atom';
 import { restartable } from './effection/restartable'
 import { assert } from 'assert-ts';
 
-export const appServer: Service<AppServiceStatus, AppOptions> = (slice, options) => {
-  let appOptions = options.atom.slice('appService', 'options');
+export const appServer: Service<AppServiceStatus, AppOptions> = (serviceSlice) => {
+  let appOptions = serviceSlice.slice('options');
 
-  return restartable(appOptions, startApp(slice));
+  return restartable(appOptions, startApp(serviceSlice));
 }
 
-const startApp = (appStatus: Slice<AppServiceStatus, OrchestratorState>) => function* (options: AppOptions) {
+const startApp = (serviceSlice: Slice<ServiceState<AppServiceStatus, AppOptions>, OrchestratorState>) => function* (options: AppOptions) {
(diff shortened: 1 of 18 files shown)


packages/server/src/app-server.ts
```diff
@@ -2,19 +2,21 @@ import { timeout, spawn } from 'effection';
...
 
-  return restartable(appOptions, startApp(slice));
+  return restartable(appOptions, startApp(serviceSlice));
 }
 
-const startApp = (appStatus: Slice<AppServiceStatus, OrchestratorState>) => function* (options: AppOptions) {
+const startApp = (serviceSlice: Slice<ServiceState<AppServiceStatus, AppOptions>, OrchestratorState>) => function* (options: AppOptions) {
   assert(!!options.url, 'no app url given');
+
+  let AppServiceStatus = serviceSlice.slice('status')
```
jnicklas: Should this be `appServiceStatus` with a lower case first letter?

jnicklas (approved):
This is great! Very nice work @dagda1!

packages/server/src/app-server.ts
dagda1: @jnicklas absolutely.  Good spot

dagda1:
> This is great! Very nice work @dagda1!

Thanks.

I was appeasing too many non-null values in 2 different places.  This starts to put a stop to that

cowboyd (approved):

PR #665, opened 2020-11-04
Replace command processor with runner

jnicklas:
The command processor is ... weird. How it communicates with the rest of the system, as well as its purpose are pretty byzantine. So let's get rid of it.

This replaces the command processor with the `Runner`, which has a similar purpose, but with a much more well defined API. Because the runner needs to be able to "background" test runs, that is not run them in the context of where they were initiated, it bundles a context on which to run the tests.

The communication between the command processor and the agents currently uses mailboxes, but this replaces it with something completely new, but which we've discussed many times before: duplex channels. A duplex channel is a pair of ends which can send and receive messages to each other. One end is connected to the agents, the other end makes it possible to communicate with them through subscriptions and sending messages.

I'm submitting this as a spike, to see if this is a direction we want to head in. If we do want to do this, there is some cleanup that needs to be done on this first.


packages/server/src/runner.ts
```diff
@@ -0,0 +1,170 @@
...
+    console.debug('[command processor] running test', testRunId);
+    let stepTimeout = 60_000;
+    let testRunSlice = this.options.atom.slice('testRuns', testRunId);
+
+    let bundlerSlice = this.options.atom.slice('bundler');
+
+    let bundler: BundlerState = yield bundlerSlice.once((state) => state.type === 'GREEN' || state.type === 'ERRORED');
+
+    if(bundler.type === 'GREEN') {
+      let events = yield Mailbox.from(this.options.agents.match({ testRunId }));
```
jnicklas, line 72: @cowboyd I would love to hear your thoughts on this pattern.

The reasoning behind this is roughly to use mailboxes as sort of a supercharged subscription. Unlike `Subscription`, a mailbox can selectively pull certain messages, while leaving the remaining messages in the inbox. However, there is a problem we've seen with mailboxes where creating a global mailbox makes it very unclear who pulls from it and when, and how any remaining messages get cleaned up. Here I've tried using channels (or in this case a DuplexChannel) and mailboxes *together*. We're creating a mailbox specifically for this test run, and once it is done, we no longer need it and can throw it away.

cowboyd: I think this is a good pattern in general, and I was thinking something similar on my [comment about a queue primitive][1]. I think this pattern will be necessary to avoid race conditions, but also to avoid having shared clobber-prone spaces. Eventually, I'd like to see mailbox evolve into a general buffer class that can be used as a basis for lifo and fifo queues, or any other number of stateful structures.

[1]: https://github.com/thefrontside/effection/issues/212#issuecomment-726495079

cowboyd (commented):
We'd talked about making the ConnectionServer a ready resource that then contains a refrence to the local end of the duplex channel. That way the remote end remains private.

packages/effection/src/duplex-channel.ts
```diff
@@ -0,0 +1,29 @@
+import { subscribe, ChainableSubscribable } from '@effection/subscription';
+import { Channel } from '@effection/channel';
+
+export interface Sendable<T> {
+  send(message: T): void;
+}
+
+export type DuplexChannel<S, R, RReturn = undefined> = ChainableSubscribable<R, RReturn> & Sendable<S>;
+
+export type DuplexPair<Tx, Rx> = [DuplexChannel<Tx, Rx>, DuplexChannel<Rx, Tx>];
```
cowboyd, line 10: It's common in JS to make things like this destructurable via both name and index. E.g.

```ts
interface DuplexPair<Tx,Rx> extends IterableIterator<[Tx,Rx]> {
  local: DuplexChannel<Tx,Rx>;
  remote: DuplexChannel<Rx,Tx>;
}
```

That way both of the following work.

```ts
let { local, remote } = DuplexChannel<Send,Receive>();
let [tx,rx] = DuplexChannel<Send,Receive>();
```
Not necessary here, but something to consider when we move it over to effection proper

jnicklas:
Updated this to make the DuplexChannel internal to the connection server.

The tests seem to be failing on Windows, I'm not sure if this is a legit failure, or what is going on here.

cowboyd:
I think this might be an issue with our timeouts being not long enough on windows. Also looks like we’re getting bit by the chrome driver bug #669

jnicklas:
@cowboyd updated to use `createDuplexChannel` instead.

cowboyd (approved):
🎉

PR #666, opened 2020-11-04
Modify post function to include content-type to satisfy geckodriver

minkimcello:
## Motivation
Trying to launch Firefox would result in Geckodriver returning an `Invalid content-type` error.

## Approach
- Added `headers` property to set content-type as `application/json`.

### DIFF
--- a/.changeset/pink-parents-help.md
+++ b/.changeset/pink-parents-help.md
@@ -1,0 +1,5 @@
+---
+"bigtest": patch
+---
+
+Modify post function in webdriver for firefox
--- a/packages/webdriver/src/web-driver.ts
+++ b/packages/webdriver/src/web-driver.ts
@@ -19,9 +19,12 @@ export class WebDriver implements Driver<WDSession> {
   }
 
   *navigateTo(url: string): Operation<void> {
-    yield request(`${this.serverURL}/session/${this.session.sessionId}/url`, {
+    yield post(`${this.serverURL}/session/${this.session.sessionId}/url`, {
       method: 'post',
-      body: JSON.stringify({ url })
+      body: JSON.stringify({ url }),
+      headers: {
+        "Content-Type": 'application/json'
+      }
     });
   }
 }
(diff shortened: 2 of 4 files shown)


cowboyd:
@minkimcello Let's make it a patch for both `@bigtest/webdriver` and `bigtest` That way, it'll come in with a patch upgrade

packages/webdriver/src/web-driver.ts
```diff
@@ -42,15 +45,18 @@ export function* connect(driver: WebDriver, options: Options): Operation<void> {
...
+    }),
+    headers: {
+      "Content-Type": 'application/json'
+    }
   });
 }
 
-function* request(url: string, init: RequestInit): Operation<WDResponse> {
+function* post(url: string, init: RequestInit): Operation<WDResponse> {
   let response: Response = yield fetch(url, init);
```
cowboyd: Since we're always posting JSON, let's have the `post` method always set the content type header, method, and serialize the body:

```ts
let response: Response = yield fetch(url, {
  ...init, 
  method: 'post',
  headers: {
    "Content-Type": "application/json"
  }
});

**Review by cowboyd — CHANGES_REQUESTED** [review]
Don't forget to add "bigtest" to the changeset 👍🏼

File: packages/webdriver/src/web-driver.ts (hunk 1 above)
Comment (minkimcello) [reply packages/webdriver/src/web-driver.ts]: By serialize the body, is this what you meant? I wasn't sure how to type the body.
I assumed since we know this function is used in two places in the same file, we can safely remove the spread of `init` and make the second argument to `post` just the body instead.
```js
function* post(url: string, body: string|object): Operation<WDResponse> {
  let response: Response = yield fetch(url, {
    body: JSON.stringify(body),
    method: 'post',
    headers: {
      "Content-Type": "application/json"
    }
  });
  // ...
}
```

.changeset/pink-parents-help.md
```diff
@@ -0,0 +1,5 @@
+---
+"bigtest": patch
+---
+
+Modify post function in webdriver for firefox
```
cowboyd, line 5: This is the how, not what/why. Let's add a refererence to the github issue that this fixes. Something like "fixes [url-to-issue] by adding content type to all webdriver requests"

minkimcello: I haven't created an issue for it. Should I create the issue? Or just describe the what/why?

cowboyd: Nah, this description is much better.

cowboyd (approved):

Issue #584, opened 2020-10-02
Can we run everything on the same port?


jnicklas:
One thing that I've been thinking about: BigTest currently uses 4 or 5 different ports:

1. command server
2. connection
3. manifest server
4. proxy server (also serves the agent)
5. app server [optional]

I am sure we do not really need to run the first three on different ports, and I'm *pretty* sure we can even map the manifest server to the same port. The app server needs to be running on a different port obviously.

The way this would work is basically through routing like this:

- `__bigtest/api` => command server/GraphQL API
- `__bigtest/connection` => WS connection for agent
- `__bigtest/manifest` => serve generated manifest, this is a bit tricky since we need to prefix the paths in the generated bundle
- `__bigtest/agent` => serve static agent from agent package
- `*` => any other path goes to the proxy server

This way BigTest itself would only occupy one single port with possibly an additional port for the app server.

jnicklas:
cc @cowboyd

cowboyd:
It seems like a sound idea. I have two thoughts that occur:

1. I think we're going to want to make this type of mechanism extensible so that plugins can run their own servers without any special consideration from the orchestrator.
2. It still feels like the (a) proxy, (b) static agent html, and (c) manifest belong at a separate origin. For example, does it make sense for a running test case to have access to the api or the connection server on a same origin basis? What does that mean for node agents, and iOS agent? For example, should the iOS agent ever need to see or load manifests from the web agent?

However, there's no reason we couldn't have a general mechanism that allowed plugins to allocate an entire origin, and then allocate urls within that origin. If everybody was configuring themselves over this central registry of origins and urls, then moving services around, splitting them or consolidating them can at least be straightforward. Bigtest can handle port allocation and configuration and your extension code never needs to worry about things like url discovery.... I like this idea of having something like a rails router, except including the origin too!

jnicklas:
That's a good point, plugins will need to be able to load custom mappings somehow. As for routing, we currently handle this through express, and it doesn't feel like we need to reinvent the wheel here.

I just hit another reason to do this: if an error occurs in the test files, its details won't be reported to the uncaught error event, it will just show up as "script error", since errors from other origins cannot be "seen" in a global error handler for security reasons. As far as I can see, the only way of solving this is to move the  manifest onto the same origin.

cowboyd:
[...]

PR #667, opened 2020-11-04
Change default suggested app port to 3000

minkimcello:
## Motivation
Sort of related to #584.

When running `yarn bigtest init`, the default suggested port for the application is [calculated by adding 1](https://github.com/thefrontside/bigtest/blob/v0/packages/cli/src/init.ts#L53) to the server port for which will [often be 24000](https://github.com/thefrontside/bigtest/blob/v0/packages/cli/src/init.ts#L25-L35). This means the application will be set up to run on 24001 but that is reserved for the proxy server.

## Approach
If `options.app.env.port` already exists, `yarn bigtest init` will use that variable as the port to suggest for running the application, if not, I've modified the default suggestion to `3000`.

I don't know if we should prevent users from setting `24001`, `24002`, `24003`, and `24005` from within `yarn bigtest init` but we need to stray away from `PORT + 1` for now in order to release beta.

### DIFF
--- a/.changeset/late-starfishes-return.md
+++ b/.changeset/late-starfishes-return.md
@@ -1,0 +1,5 @@
+---
+"bigtest": patch
+---
+
+Change default suggested app port to 3000
--- a/.changeset/sharp-rocks-compete.md
+++ b/.changeset/sharp-rocks-compete.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/cli": patch
+---
+
+Change the default suggested app port to 3000 in bigtest init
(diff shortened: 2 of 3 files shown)


minkimcello:
@cowboyd I added patch changeset for both `packages/cli` and `bigtest` here too.

jnicklas:
Ahh, this is indeed a bit unfortunate right now, but I don't think this is the right solution either. 3000 is a port which is *very* commonly occupied by a dev server, so I don't think it's a good alternative. I do think that PORT+1 is a good choice, we should just make sure that the default doesn't collide with any other ports we're using. So just switching the default to 28000 or whatever would be better, IMO.

minkimcello (commented):
@jnicklas Changed default suggested port to 28000 and reverted the default port of application to port+1.

jnicklas (approved):

jnicklas:
@minkimcello Took the liberty of merging this.

FYI you could have combined both changesets into one, but it doesn't really matter.

PR #668, opened 2020-11-05
Pc multi step context threading

dagda1:
I had to resort to type overloads to fulfil the requirement of a `...steps` rest argument. 

I still got recursive compiler errors with typescript 4.1.

All take the same form and just build on the previous.

```ts
export type Nothing = void | undefined | null;

export type ActionContext = Context | void;

export type ResolveContext<C extends Context, R extends ActionContext> = R extends Nothing ? C : C & R;

export type Stepable<C extends Context> = {
  <R extends ActionContext>(
    step: StepDefinition<C, R>
  ): TestBuilderImplementation<ResolveContext<C, R>>;
  <R1 extends ActionContext, R2 extends ActionContext>(
    step1: StepDefinition<C, R1>,
    step2: StepDefinition<ResolveContext<C, R1>, R2>
  ): TestBuilderImplementation<ResolveContext<ResolveContext<C, R2>, R1>>;
// etc. etc.
```

I currently have gone 6 deep but if you want more, then you need more overloads.

I've pulled the overloads out into a separate [types file](./packages/suite/src/types.ts) to hide the ugliness.

### DIFF
--- a/packages/suite/src/dsl.ts
+++ b/packages/suite/src/dsl.ts
@@ -1,4 +1,6 @@
 import { TestImplementation, Context, Step, Assertion } from './interfaces';
+import { Stepable, ActionContext, StepDefinition, Action, ResolveContext, AssertionList, Check, TestBuilderImplementation } from './types';
+
 
 export function test<C extends Context>(description: string): TestBuilder<C> {
   return new TestBuilder<C>({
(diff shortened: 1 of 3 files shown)


cowboyd (approved):
You're a hero @dagda1! I don't know where we go from here, but not being able to inject context from a step object is the worst.

dagda1:
> You're a hero @dagda1! I don't know where we go from here, but not being able to inject context from a step object is the worst.

Why do these overloads make me feel more like a villain :).

PR #670, opened 2020-11-07
make ProxyServer a service

dagda1:
Conversion of the proxy server into a service.

Like the other services, the configuration now takes place in the `createOrchestratorAtom` and not in the orchestrator.

```ts
    proxyService: {
      status: {type: 'unstarted'},
      options: {
        port: project.proxy.port,
        prefix: agentServerConfig.options.prefix,
        appDir: agentServerConfig.appDir(),
        harnessUrl: agentServerConfig.harnessUrl(),
        appOptions: project.app
      }
    },
```

The `proxyServer` then just takes the `Slice` in the `orchestrator`

```ts
yield fork(proxyServer(options.atom.slice('proxyService')));
```

### DIFF
--- a/packages/server/src/orchestrator/atom.ts
+++ b/packages/server/src/orchestrator/atom.ts
@@ -11,7 +11,7 @@ export type OrchestratorAtomOptions = Pick<ProjectOptions, 'app' |'watchTestFile
 export const createOrchestratorAtom = (project: OrchestratorAtomOptions) => {
   let manifestSrcDir = path.resolve(project.cacheDir, 'manifest/src');
   let manifestSrcPath = path.resolve(manifestSrcDir, 'manifest.js');
-  let agentServerConfig = new AgentServerConfig({ port: project.proxy.port, prefix: project.proxy.prefix, });
+  let agentServerConfig = new AgentServerConfig(project.proxy);
 
   let atom = new Atom<OrchestratorState>({
     manifestGenerator: {
@@ -44,7 +44,7 @@ export const createOrchestratorAtom = (project: OrchestratorAtomOptions) => {
         prefix: agentServerConfig.options.prefix,
         appDir: agentServerConfig.appDir(),
         harnessUrl: agentServerConfig.harnessUrl(),
-        // TODO: this is duplication but currently we can only pass 1 slice into a service
+        // TODO: this is duplication because currently we can only pass 1 slice into a service
         // Is this problematic?
         appOptions: project.app
       }
--- a/packages/server/src/orchestrator/atom.ts
+++ b/packages/server/src/orchestrator/atom.ts
@@ -2,14 +2,16 @@ import { Atom } from "@bigtest/atom";
 import { OrchestratorState } from "./state";
 import { ProjectOptions } from '@bigtest/project/dist';
 import path = require('path');
+import { AgentServerConfig } from '@bigtest/agent';
 
 // TODO: eventually we can remove this type and just use ProjectOptions.
 // But until then we can just pick the bits we need.
-export type OrchestratorAtomOptions = Pick<ProjectOptions, 'app' |'watchTestFiles' | 'cacheDir' | 'testFiles'>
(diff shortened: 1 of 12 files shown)


packages/server/src/orchestrator/atom.ts
```diff
@@ -36,7 +38,16 @@ export const createOrchestratorAtom = (project: OrchestratorAtomOptions) => {
...
-      proxyStatus: 'unstarted'
+      status: {type: 'unstarted'},
+      options: {
+        port: project.proxy.port,
+        prefix: agentServerConfig.options.prefix,
+        appDir: agentServerConfig.appDir(),
+        harnessUrl: agentServerConfig.harnessUrl(),
+        // TODO: this is duplication because currently we can only pass 1 slice into a service
+        // Is this problematic?
+        appOptions: project.app
```
cowboyd, line 49: I'm not super concerned about finding away around this duplication for now as I think there are a number of strategies we can follow and I'm not sure which is the best one.

However, on the set side, maybe we can hedge by instead of having the test helper set both locations, but instead just have an operation in the orchestrator that syncs one to the other. Roughly:

```ts
atom.slice('appOptions').forEach(state => {
  atom.slice('proxyService','options','appOptions').set(state);
});
```

That way, if we ever do get to a point where we're setting the app options from a running orchestrator it will work in both the deployed version and the test case.

cowboyd (commented):
This is looking great to me. What do you think about aliasing the states?

packages/server/src/orchestrator/atom.ts
jnicklas: Maybe we could have one unified options property in the state, which would store all of the options in the same format that we have in the `json` file?

jnicklas: (not saying we should do this in this PR, but as a future direction to head in)

dagda1: @cowboyd @jnicklas I was heading in the direction of keeping all configuration out of orchestrator.  By the time it reaches there, it should be just a matter of starting the services in the correct sequence.  

I quite liked the idea of each service having its own config and a status.

A service or really anything to do with the state should only care about the information it needs in order for it to do what it has to do.

The `proxyService` blew that out of the water, unfortunately.

> However, on the set side, maybe we can hedge by instead of having the test helper set both locations

Is keeping it in sync only important for the tests?  If so then I would be hesitant to do anything in the production code, just to make the tests pass.

cowboyd: fwiw, I think the proxyService and `appService` will end up being merged since they're really part of the same mechanics of testing webapps.

cowboyd: Which is to say I expect they'll get "de-duped" later on down the road...

dagda1: @cowboyd it looks like the `url` is the only property it needs from `appOptions`.

Am I right in saying that the point of 

```ts
return restartable(appOptions, startProxyServer(serviceSlice));` 
```

is to restart if any of the props of the `appOptions` slice change?

If so then maybe we could put something in the orchestrator to sync this like you mentioned earlier.

dagda1: or maybe as @jnicklas stated, we could have all the config in one slice and restart when anything changes.

dagda1:
> This is looking great to me. What do you think about aliasing the states?

i'm not sure what this would involve.  would it be possible to give an example please?

cowboyd:
> > This is looking great to me. What do you think about aliasing the states?
> 
> i'm not sure what this would involve. would it be possible to give an example please?

It was the first sample of copying on value to the other

packages/server/src/orchestrator/atom.ts
dagda1: are we happy just to live with the duplication for now?

cowboyd (approved):
Sounds like we're going to punt on de-duping the `set()` of options since we only need to do it in the test, and we expect that we will merge these two services into one at some point anyhow.

PR #671, opened 2020-11-11
✨Windows Support

### LINKED ISSUE #432: Run Tests on Windows

cowboyd:
> resolves #432

Motivation
-----------

Bigtest development workflows should work on windows as well. This will not only guarantee that bigtest can run on windows, but that BigTest itself can be developed on Windows whether a developer is using cmd.exe, powershell.exe, or the linux subsystem.

Approach
----------

Most of the problems with running the tests on windows have to do with the build scripts not being able to run. For example, `chmod`, `rm -rf`, etc are unix-only commands. Also, the trick of setting a one-off environment variable does not work for running a command (FOO=bar my-command).

In order to make these robustly cross-platform, build scripts have been moved, where necessary into javascript.

### DIFF
--- a/packages/agent/test/agent.test.ts
+++ b/packages/agent/test/agent.test.ts
@@ -118,7 +118,7 @@ describe("@bigtest/agent", function() {
           if(error && stack && logEvents) {
             expect(error.name).toEqual('Error');
             expect(error.message).toEqual('boom!');
-            expect(stack[0].source && stack[0].source.fileName).toContain('/test/fixtures/manifest.js');
+            expect(stack[0].source && stack[0].source.fileName).toContain('fixtures/manifest.js');
             expect(logEvents).toEqual(expect.arrayContaining([
               expect.objectContaining({ type: "message", message: { level: 'log', text: 'this is a good step' } }),
               expect.objectContaining({ type: "message", message: { level: 'log', text: 'some log message here' } }),
--- a/packages/agent/test/agent.test.ts
+++ b/packages/agent/test/agent.test.ts
@@ -32,7 +32,12 @@ describe("@bigtest/agent", function() {
     expect(fixtureManifest).toBeDefined();
   });
 
-  this.timeout(process.env.CI ? 60000 : 10000);
+  if (process.platform === 'win32') {
+    this.timeout(process.env.CI ? 120000 : 30000);
+  } else {
+    this.timeout(process.env.CI ? 60000 : 10000);
+  }
+
 
   describe('config', () => {
     it('has an agent url where it will server the agent application', () => {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -50,7 +50,7 @@
     "@bigtest/performance": "^0.5.0",
     "@bigtest/project": "^0.12.0",
     "@bigtest/server": "^0.19.0",
-    "@effection/node": "^0.9.0",
+    "@effection/node": "^0.9.1",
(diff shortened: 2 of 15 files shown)


packages/agent/test/agent.test.ts
```diff
@@ -32,7 +32,12 @@ describe("@bigtest/agent", function() {
     expect(fixtureManifest).toBeDefined();
   });
 
-  this.timeout(process.env.CI ? 60000 : 10000);
+  if (process.platform === 'win32') {
+    this.timeout(process.env.CI ? 120000 : 30000);
```
jbolda, line 36: 😂

packages/cli/package.json
```diff
@@ -16,7 +16,7 @@
     "test": "mocha -r ts-node/register \"test/{,!(fixtures)/**}/*.test.ts\"",
     "mocha": "mocha -r ts-node/register",
     "bigtest": "ts-node src/index.ts",
-    "prepack": "tsc --outdir dist --module commonjs && chmod +x dist/bin/bigtest.js"
+    "prepack": "tsc --outdir dist --module commonjs && node chmod.js 755 dist/bin/bigtest.js"
```
jnicklas: Why do we even need to `chmod` here? Shouldn't yarn/npm take care of this for us?

packages/server/generate-schema.js
```diff
@@ -0,0 +1,52 @@
...
+  console.log(`$ ${command}`);
+  let child = yield exec(command, options);
+
+  yield spawn(child.stdout.subscribe().forEach(function*(datum) {
+    process.stdout.write(datum);
+  }));
+
+  yield spawn(child.stderr.subscribe().forEach(function*(datum) {
+    process.stderr.write(datum);
+  }));
```
jnicklas, line 34: Can't we just use `stdio: inherit`?

packages/cli/package.json
dagda1: You will need `chmod` for shebangs etc.

you can  use the `--chmod` switch of `git update-index` though to commit

e.g.

```bash
git update-index --chmod=755 --add dist/bin/bigtest.js
```

cowboyd: Never new that was a even a thing!

packages/server/generate-schema.js
cowboyd: We actually don't export the `stdio` option in your `exec()` api https://github.com/thefrontside/effection/blob/v0/packages/node/src/exec/api.ts#L25-L47

We could add that to `exec()`, but now that I think of it, that was always a pretty confusing API. Maybe we should add `sh()` to `@effection/node`?

jbolda: If we "inherit", then we can't act on any of the output, essentially a pure side effect. Using "inherit" is what nearly every use of child_process does which is partly why we are likely one of the few that has run into the deep issues on windows that we have had. Maybe another function makes sense, but I wonder if the use case might be outside the scope of this package?

jbolda (approved):
Looking forward to make use of this!

packages/server/generate-schema.js
cowboyd: Yeah, it's a dang useful function, but forcing the the user to decide where the bytes go feels like the right primitive.

dagda1 (approved):
we can come back to this at any time.

PR #672, opened 2020-11-13
Export Cypress commands from Interactor

minkimcello:
## Motivation
To allow Cypress users to easily integrate interactors into their tests.

## Approach ⭐
- Put the file inside `interactor/src/integration/` directory in case we need to add similar functionality for other testing tools.
- The functions that are added are `do` and `expect` for `steps` and `assertions` respectively and they can take either a single interactor or an array of interactors.
- The commands are automatically evaluated so users do not need to import anything separately or modify their cypress configurations.
- I added a simple test to run against `@bigtest/todomvc` to confirm that `do` and `expect` can take either single interactors or an array of interactors. The cypress folder is nested under `packages/interactor/test/integrations/cypress/`.
  - ~~I've also added a new test script `test:cypress` which I've also slapped onto the end of the test script so that it runs in our github actions. I considered using the cypress action but it doesn't provide any value for us and given the matrix strategy of our test workflow, the tidiest approach was just to add the script to the package.json.~~
  - ⭐ I added the cypress test as a separate job within the test workflow so that it'll all be nested together with the other tests in the github actions GUI (as opposed to having an entirely separate workflow)

## TODOs
- [x] ~Add instructions?~ Not necessary
- [x] ~Check cypress logging contents.~ Not urgent.
- [x] ~Having difficulty running tests. Says `Cypress is not defined` even if Cypress is added as a dependency.~ Fixed :+1:
- [x] Setting `bigtestGlobals.runnerState = "steps"` won't trigger warning when running assertion.
  - Supposed to be like this. Confirmed it fails when you run a step with `bigtestGlobals.runnerState = "assertion"`.
- [x] Some typescript config problems with running cypress and the relevant issue on cypress are still open [#8477](https://github.com/cypress-io/cypress/issues/8477)
  - ~~"sourceMap" needs to be set to false but tsconfig configs inside `packages/interactor/test/integrations/cypress/` doesn't make a difference.~~ All good.
- [x] ~~Custom command won't be recognized by typescript if there are imports/requires in `packages/interactor/test/integrations/index.d.ts` [#1065](https://github.com/cypress-io/cypress/issues/1065#issuecomment-351769720) so for now I have the arguments typed as `any`.~~ All good.
- [x] ~~Not sure if it's appropriate to include `index.d.ts` in `interactor/src/integrations/`. All of the other type files are auto-generated but `src/integrations/cypress` doesn't have any named exports.~~ All good.

## Screenshot
Screen Shot 2020-11-25 at 11 13 17 AM


packages/interactor/src/integration/cypress.ts
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { bigtestGlobals } from '@bigtest/globals';
+
+declare const Cypress: any;
```
dagda1: if you add to the tsconfig.json
```ts
 "types": ["cypress"]
```

Then you can delete these `any` declarations and get full cypress support

minkimcello: Oh thank you. I added cypress as a dev dependency to get `"types": ["cypress"]` to work and I was able to remove the `any` declarations.

packages/interactor/src/integration/cypress.ts
```diff
@@ -0,0 +1,38 @@
...
+  }).then(() => {
+    Cypress.log({
+      displayName: runnerState,
+      message: interaction.description
+    });
+  })
+}
+
+Cypress.Commands.add('do', (
+  interaction: Interaction<void>
```
minkimcello: @dagda1 Does this one look okay? The command can take either a single interactor or an array of interactors so I feel like I'm not typing this accurately.

packages/interactor/src/index.ts
```diff
@@ -14,3 +14,5 @@ export { CheckBox } from './definitions/check-box';
 export { RadioButton } from './definitions/radio-button';
 export { Select } from './definitions/select';
 export { MultiSelect } from './definitions/multi-select';
+
+export * as CypressCommands from './integration/cypress';
```
cowboyd: This module doesn't export anything, but instead is evaluated for its side-effects. The canonical method for doing this is like so:

```ts
import './integration/cypress'
```

That way it's clear to anyone reading the code that this think is going to tinker with global state.

cowboyd:
So how do we test this?

minkimcello:
@cowboyd I could run our cypress commands against todomvc to confirm that the commands work.

cowboyd:
> @cowboyd I could run our cypress commands against todomvc to confirm that the commands work.

That sounds good. Hopefully it should be pretty easy to set up on CI. I wonder if we should have a separate "cypress" check so that we don't have to d/l and install cypress as part of every test suite rather than putting it directly into the interactor unit test suite.

minkimcello:
@cowboyd I have cypress added as a devdependency for typing purposes so wouldn't each test suite install it regardless?

cowboyd:
@minkimcello It looks like it. I guess we just need to be mindful of install time

packages/interactor/test/integrations/cypress/integration/interactor.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { Heading } from '../../../../src';
+
+describe('Cypress with Interactors', () => {
+  it('', () => {
+    cy.visit('/')
+    cy.do(Heading('todos').exists())
```
cowboyd: Technically here we want to check a "step" in a `do` and an assertion in an `expect` and make sure they work when used correctly, but that they fail if you use a step outside of a `do`. 

In this example, I'd use `.expect()`

minkimcello: Oh so the test looks like that right now because setting `bigtestGlobals.document = "step"` for `cy.do()` does not return an error when you pass in an assertion so I'm trying to figure out why.

cowboyd: It's ok to pass an assertion to `do`, what's not ok is to pass a step to `expect`

packages/interactor/tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "extends": "@frontside/tsconfig",
   "compilerOptions": {
-    "baseUrl": "/",
     "esModuleInterop": true,
-    "types": ["cypress"]
+    "types": ["cypress"],
+    "sourceMap": false
```
cowboyd: 🤔

minkimcello: I think I'm required to set sourceMap to false in order to run cypress (as seen in this cypress issue [#8477](https://github.com/cypress-io/cypress/issues/8477) so I tried to nest it inside `packages/interactor/test/integrations/cypress/tsconfig.js` but the tsconfig in that directory was not registering. (So I added `"sourceMap": false` just to get cypress running. I'll revert it once I figure out the config.)

cowboyd:
I’m ok if we run this as a separate GitHub workflow that only happens on Ubuntu

cowboyd (commented):
This is looking ready to ship IMHO. However, for the workflow, if we're only planning on running the cypress integration on ubuntu, then let's not have a matrix of one.

.github/workflows/test.yml
```diff
@@ -65,3 +65,31 @@ jobs:
         run: yarn
       - name: Run tests
         run: yarn workspace @bigtest/${{ matrix.package }} test
+
+  cypress:
+    runs-on: ubuntu-latest
+    name: interactor cypress (ubuntu-latest)
```
cowboyd:
```suggestion
    name: interactor cypress integration
```

minkimcello: done

.changeset/chilled-ears-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigtest/interactor": minor
+---
+
+Add cypress implementation for interactors
```
cowboyd:
```suggestion
Add cypress integration for interactors
```

cowboyd (approved):

PR #673, opened 2020-11-14
add cross-env and a test:debug to @bigtest/server

dagda1:
small PR to make life easier for forgetful people like myself.

I can never remember how to set environment variables in windows.

I've added [cross-env](https://github.com/kentcdodds/cross-env) and a simple yarn script to `@bigtest/server` so I don't have to remember.

```json
"test:debug": "cross-env LOG_LEVEL=debug yarn test",
```


cowboyd (approved):
Looks good to me, but at the same time I have do a vague recollection of @jbolda  trying out `cross-env` before and there was something wonky about it?

jbolda:
@cowboyd as far as I know, I don't think we shouldn't have any issues with it. It does _also_ use `cross-spawn` and spawns a process. I might have been avoiding it simply to avoid having more processes to worry about.

cowboyd (approved):

dagda1:
@jbolda @cowboyd i will merge this but please feel free to remove it, if it is problematic

PR #675, opened 2020-11-16
Replace switch with match spike

dagda1:
> switches don't bring riches

I was finding it harder and harder to look at the switches in `manifest-builder` and the `bundler` and especially as I am responsible.  

They are necessary to type narrow but I had a look around and there does seem to be a common pattern of using a matcher.

I had a look at [ts-adt](https://github.com/pfgray/ts-adt) but it only worked with a `_type` tag so I had a go at writing my own.

The upshot is that it replaces, this code:

```ts
      let messages = on(rollup, 'event')
        .map(([event]) => event as RollupWatcherEvent)
        .filter(event => ['START', 'END', 'ERROR'].includes(event.code))
        .map(event => {
          switch (event.code) {
            case 'START':
              return { type: 'START' } as const;
            case 'END':
              return { type: 'UPDATE' } as const;
            case 'ERROR':
              return { type: 'ERROR', error: event.error } as const;
            default:
              throw new Error(`unexpect event ${event.code}`);
          }
        });
```

With this

```ts
      let messages = on(rollup, 'event')
        .map(([event]) => event as RollupWatcherEvent)
        .filter(event => ['START', 'END', 'ERROR'].includes(event.code))
        .map(event => {
          return match<RollupWatcherEvent>('code')({
            START: () => ({ type: 'START' }) as const,
            END: () => ({ type: 'UPDATE' } as const),
            ERROR: ({ error }) => ({ type: 'ERROR', error } as const),
          })(event)
        });

```


cowboyd:
Would be cool if we could treat non-functions as constant functions. Thus the matcher:

```ts
{
  START: () => ({ type: 'START' }) as const,
  END: () => ({ type: 'UPDATE' } as const),
  ERROR: ({ error }) => ({ type: 'ERROR', error } as const)
}
```
Could be written as:

```ts
{
  START: { type: 'START'} as const,
  END: () => { type: 'UPDATE' } as const,
  ERROR: ({ error }) => ({ type: 'ERROR', error } as const)
}
```
```

**Comment by dagda1** [comment]
> Would be cool if we could treat non-functions as constant functions. Thus the matcher:
> 
> ```ts
> {
>   START: () => ({ type: 'START' }) as const,
>   END: () => ({ type: 'UPDATE' } as const),
>   ERROR: ({ error }) => ({ type: 'ERROR', error } as const)
> }
> ```
> 
> Could be written as:
> 
> ```ts
> {
>   START: { type: 'START'} as const,
>   END: () => { type: 'UPDATE' } as const,
>   ERROR: ({ error }) => ({ type: 'ERROR', error } as const)
> }
> ```

Does that not mean we have to decide whether it is an object or a function when all `cases` should be the same

**Comment by cowboyd** [comment]
Yeah, nvm. Because the result could itself be a function and that would introduce an ambiguity

**Comment by jnicklas** [comment]
I like this a lot. I really miss exhaustive match statements in languages that don't have them. It might be worth considering if an array-based API could be more flexible in the end, something like this:

``` typescript
return match('code')<RollupWatcherEvent, BundlerMessage>({
  ["START", () => ({ type: 'START' } as const)],
  ["END", () => ({ type: 'UPDATE' } as const)],
  ["ERROR", ({ error }) => ({ type: 'ERROR', error } as const)],
})(event)
```

This would make it possible to have future extensions like this:

``` typescript
return match('code')<RollupWatcherEvent, BundlerMessage>({
  ["START", () => someCondition, () => ({ type: 'START', thing: 'foo' } as const)],
  ["START", () => ({ type: 'START' } as const)], // fallback for START
  ["END", () => ({ type: 'UPDATE' } as const)],
  [({ error }) => ({ type: 'ERROR', error } as const)], // default case
})(event)
```

dagda1:
> I like this a lot. I really miss exhaustive match statements in languages that don't have them. It might be worth considering if an array-based API could be more flexible in the end, something like this:

Thanks!!

I can make this is exhaustive or not, it actually was at the start but I changed it.

I was going to add an `otherwise(() =>({ a:  'whatever' })` , for the default case expression.

I could do this with function overloading though, I don't think I need the array though or am I missing something?

dagda1:
@jnicklas I was also thinking of a `match` and `matchStrict` where you had to supply all members of the union.

jnicklas:
So one downside of this that I realized is that it interacts in an unfortunate way with Effection. Unfortunately, Effection and anonymous functions are not friends.

So this:

``` typescript
switch(something) {
  case 'WHATEVER':
    return yield someOperation(); 
  // ...
}
```

Is pretty much impossible to translate into a function based construct. This is an unfortunate side-effect of using continuations in a functional language which does not have TCP-preserving closures, see [this excellent blog post about the same problem in Rust](https://boats.gitlab.io/blog/post/the-problem-of-effects/) for context.

We could make `match` itself an operation, or we could have an effectful version of match, but neither of these options seem particularly great to me.

dagda1:
> Is pretty much impossible to translate into a function based construct.

This definitely occurred to me afterwards also.  The only way to make this work would be to have some sort of runner that promisified or made everything a generator for all the anonymous case functions.

Possibly my favourite bit of typescript or what makes typescript feasible is the ability to deal with discriminated unions or algebraic data types, e.g.

```ts
type Auth  = 
| {
     type: 'Unauthenticated'
  }
| {
    type: 'Authenticated',
    token: string
}
```

This beats class-based state representation because if the current state is `Unauthenticated` then I cannot access the token.

Matching on the discrminated union is a poor man's pattern matching but the best we got right now.

I really want a `match` similar to `once` in the `Atom` with an API like the `match` in this PR but obviously works with generators.

If `once` works with one condition then a `match` can work with multiple conditions....I think :).

I still think this `match` function is useful for non-async matching but I am totally cool if this PR gets refused.  It was a spike and did not take that long.

PR #678, opened 2020-11-17
Bump all @frontside/typescript to v1.1.0

minkimcello:
## Motvation
To bump all of `@frontside/typescript` from `v1.0.1` to `v1.1.0`.

## Approach
Not adding changesets because I don't think this constitutes a patch bump.

### DIFF
--- a/.changeset/loud-ladybugs-burn.md
+++ b/.changeset/loud-ladybugs-burn.md
@@ -1,0 +1,20 @@
+---
+"@bigtest/agent": patch
+"@bigtest/atom": patch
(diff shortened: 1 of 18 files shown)


jnicklas (approved):
Nice!

cowboyd:
@minkimcello We should patch bump every package that it effected, including `bigtest` because this will change the javascript that is output because it contains a TypeScript upgrade. So code packaged into a `.tgz` will be materially different.

cowboyd (approved):

Issue #683, opened 2020-11-19
Epic: Release Interactors Beta, Platform Alpha


cowboyd:
We want to release what we have at the beginning of December with a bang, and what that means is that we want everything that we _do_ have to be "swipe to unlock". As it currently stands, we feel that the interactors as a standalone library fits that bill. You can take it and install it in Jest, Cypress, Mocha, and yes, in our own runner, and experience immediate, palpable benefit. 

On the other hand, the runner, test syntax, and dev experience, etc... still feels a bit rough around the edges and some time is needed for it to settle. Part of this is that a library is simpler to drop in than something that has a CLI and lots of moving parts. The other is that we still have a lot that feels experimental in our runner.

As a result, we are going to push the interactors as the main feature of the release, and then bill everything else as Platform that not only contextualizes interactors with the important motivating factors about why we are doing all this work, but also provides an on ramp for early adopters that want to get started with writing the tests in a new way.

> We initially thought about calling it "runner", but that doesn't really do the system justice. From here on out, when we talk about BigTest Platform, we're talking about the CLI, Orchestrator, driver, agent, etc...

Along side the code, we want to release an entire site dedicated to bigtest that will hang directly off of the main `frontside.com` website: `https://frontside.com/bigtest` that will contain everything that users will need to know about bigtest. This will including marketing material such as rationale, and value proposition as well as detailed technical guides for the various pieces which today include interactors and platform. 90% of our effort will be focused on the interactor story since we feel that this is what is most actionable and ready as of today. But we'll spend 10% of our effort on the large picture and a _very_ minimal guide for starting the 

@jorgelainfiesta, @cherewaty and I created draft sitemap for the release site that will contain all of the relevant pages/sections. So once this sitemap is in place, each page can be picked off and executed, or executed in pieces. The person spearheading each piece is marked next to that piece. That doesn't necessarily mean that that person will be doing all of the work on that section, but it does mean that they'll be the main point of contact for it.

- [x] create https://frontside.com/bigtest placeholders for the following sitemap @jorgelainfiesta 
- [x] /bigtest/ @jorgelainfiesta 
  - [x] value proposition: interactors
  - [x] call to action: interactors (link to interactors guide)
  - [x] value proposition: platform (not as emphasized as interactors, mostly recycle from current marketing site)
  - [x] call to action: platform (link to platform guide)
- [x] /bigtest/about @cowboyd @jnicklas #710 
- [x] /bigtest/interactor/: @jenweber
[...]

PR #686, opened 2020-11-22
Separates documentation for Interactors and Platform

jorgelainfiesta:
## Motivation

As stated on #683, we will release Interactors as beta and the Platform as alpha. The user can use Interactors with their existent test suite.

We expect Interactors to get more traction than the Platform at this point. This means that users reading about Interactors may not be interested in the documentation for the Platform, so it makes sense to keep them in separate places. 

Furthermore, even after the Platform is stable, Interactors will still have a broader adoption so separate documentation would still be convenient for users later on. At some point we may merge everything into one documentation, but in a short or medium term I don't see any benefit to that.

## Approach

To keep everyone unblocked on writing their respective content, this PR introduces the pages' structure on Docusaurus.

### Interactors (@jenweber @minkimcello)

I moved the files you were working on inside `/docs/interactors`, which creates the issue of having `/docs/interactors/interactors`. But now that we have re-arranged the contents on the overall website, maybe we can rethink how we structure that section.

Also, I created a redirect on `/docs` to keep `/docs/interactors` as the default if anybody enters that URL. It may also be handy now that the top nav won't have a direct link to your work.

### Platform (@cowboyd @jnicklas)

You can add the pages about architecture and such here. If you have any questions about Docusaurus, I'd be more than happy to help. 

### "Marketing" pages

The lading pages for value props and a less technical intro to the tools are `src/pages/index.js`, `src/pages/interactors.js`, `src/pages/platform.js`.

I still don't know how I'll fill that in. Maybe it's worth investing in porting more formal styles of our website to React instead of relying on hacks to integrate webflow. 

_PS: I'm making a PR to another PR because I thought it was too aggressive to move everything in a branch, where others have been working, without prior notice. Netlify won't create a preview for this PR because it only does that for PRs against `v0`_


jenweber (approved):
Hi! I won’t have time until later tonight to review, but feel free to move forward anyway without me. Thanks for the heads up!

minkimcello (approved):

Issue #679, opened 2020-11-17
typescript errors about missing tsconfig.json are printed to the console in non-typescript projects


cowboyd:
When integrating the bigtest runner into the stripes-components project, the bundler threw this error out to the console:

```
rpt2: config error TS18003: No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["**/*"]' and 'exclude' paths were '[]'.
```

The project is in vanilla javascript and not typescript, so that's probably related. However, even though it prints this error, the tests to run properly.

Here is the full project.https://github.com/cowboyd/stripes-components/tree/integrate-bigtest

jnicklas:
There is no error ;)

cowboyd:
@jnicklas 😹 ok, thanks for that.... comment updated.

dagda1:
it is an error coming from the `rollup-plugin-typescript2` which is why it is prefixed with `rpt2`.

It is because there is no tsconfig.json and nowhere in the rollup config do we specify `include` or `exclude` fields that tsc uses to know where to look for or avoid typescript files.

I think we need to update our defaults to:

```ts
export declare function defaultTSConfig(): {
compilerOptions: {
    skipLibCheck: boolean;
    target: string;
    lib: string[];
    include: [
        "nowhere/**/*.ts",
      ],
    };
};
```

cowboyd:
@dagda1 That seems to change it to:

```
rpt2: config error TS18003: No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["nowhere/**/*.ts"]' and 'exclude' paths were '[]'.
```

dagda1:
@cowboyd oh right so the error is that it cannot find any typescript files.

I wanted to avoid us deciding if it is a typescript or javascript project

dagda1:
stdout or stderr from [here](https://github.com/ezolenko/rollup-plugin-typescript2/blob/42173460541b0c444326bf14f2c8c27269c4cb11/src/parse-tsconfig.ts#L51).

dagda1:
a couple of things we could do is:

```ts
const isTypescript = fs.existsSync(process.cwd(), 'tsconfig.json);
```

Not sure if we can guarantee that but might be fine

The error message is just a console.something and not an actual error so we could suppress it somehow.

dagda1:
if it is typescript, we either disable the plugin or do not add it to the plugins array

cowboyd:
[...]

PR #687, opened 2020-11-25
don't include the typescript plugin if there is no tsconfig option

dagda1:
This solution all hangs on the assumption that if there is no option for `tsconfig` then we can assume this is not a typescript project.

This is to silence the

> rpt2: config error TS18003: No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["**/*"]' and 'exclude' paths were '[]'.

issue #679

slightly speculative at this stage.   

```ts
    plugins: [
      resolve({
        mainFields: ["browser", "module", "main"],
        extensions: ['.js', '.ts'],
      }),
      commonjs(),
      bundle.tsconfig && typescript({
        tsconfig: bundle.tsconfig,
        tsconfigDefaults: defaultTSConfig(),
        tsconfigOverride: {
          compilerOptions: {
            module: "ESNext",
          }
        }
      }),
      babel({
        babelHelpers: 'runtime',
        extensions: ['.js', '.ts'],
        presets: ['@babel/preset-env', '@babel/preset-typescript'],
        plugins: ['@babel/plugin-transform-runtime']
      }),
      injectProcessEnv({
        NODE_ENV: 'production'
      }),
    ].filter(Boolean)
```

### DIFF
--- a/.changeset/honest-jeans-ring.md
+++ b/.changeset/honest-jeans-ring.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/bundler": minor
+---
+
+don't include the typescript plugin if there is no tsconfig option
--- a/packages/bundler/src/bundler.ts
+++ b/packages/bundler/src/bundler.ts
@@ -33,7 +33,7 @@ function prepareInputOptions(bundle: BundleOptions, channel: Channel<BundlerMess
         extensions: ['.js', '.ts'],
       }),
       commonjs(),
-      typescript({
+      bundle.tsconfig && typescript({
         tsconfig: bundle.tsconfig,
         tsconfigDefaults: defaultTSConfig(),
         tsconfigOverride: {
@@ -51,7 +51,7 @@ function prepareInputOptions(bundle: BundleOptions, channel: Channel<BundlerMess
       injectProcessEnv({
         NODE_ENV: 'production'
       }),
-    ]
+    ].filter(Boolean)
   }
 }
 
--- a/packages/bundler/src/bundler.ts
+++ b/packages/bundler/src/bundler.ts
@@ -45,7 +45,7 @@ function prepareInputOptions(bundle: BundleOptions, channel: Channel<BundlerMess
       babel({
         babelHelpers: 'runtime',
         extensions: ['.js', '.ts'],
-        presets: ['@babel/preset-env', '@babel/preset-typescript'],
+        presets: ['@babel/preset-env'],
         plugins: ['@babel/plugin-transform-runtime']
       }),
       injectProcessEnv({
--- a/packages/bundler/test/bundler.test.ts
+++ b/packages/bundler/test/bundler.test.ts
@@ -4,9 +4,12 @@ import { promises as fs, existsSync } from 'fs';
 import expect from 'expect';
 import rmrf from 'rimraf';
 import { subscribe } from '@effection/subscription';
+import path from 'path';
 
 import { spawn } from './world';
 import { Bundler } from '../src/index';
+
+const tsconfig = path.join(process.cwd(), 'tsconfig.json');
 
 describe("Bundler", function() {
   this.timeout(5000);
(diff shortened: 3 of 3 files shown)


cowboyd:
@dagda1 are the CLI tests failing because of a dependency issue?

cowboyd:
@dagda1 this seems like a legit failure https://github.com/thefrontside/bigtest/pull/687/checks?check_run_id=1475212211#step:7:212 maybe something is off in the order in which plugins are getting required because of filesystem directory listing order differences between OSes?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/bigtest/pull/687">don't include the typescript plugin if there is no tsconfig option by dagda1 · Pull Request #687 · thefrontside/bigtest</a></strong></div><div>This solution all hangs on the assumption that if there is no option for tsconfig then we can assume this is not a typescript project.
This is to silence the

rpt2: config error TS18003: No inputs ...</div></blockquote>

dagda1:
> @dagda1 this seems like a legit failure 

it most definitely is, it is because `bundle.tsconfig` is undefined when the cli tests run.

I was expecting this to be truthy if it was a typescript project but maybe this is not something that can be relied upon?

```ts
      bundle.tsconfig && typescript({
```

jnicklas:
It's a bit of a shame that with this change, typescript users will *have* to use a custom tsconfig file. Is there no other way around this issue?

dagda1:
> It's a bit of a shame that with this change, typescript users will _have_ to use a custom tsconfig file. Is there no other way around this issue?

Oh, maybe I've explained this badly.

This is to do with #679 and silencing the errors if there are no typescript files.

I've not removed any default tsconfig.json

In this PR the code checks that the configuration passed to the bundler has a `tsconfig` option and only if it does have such an option does it add the `rollup-plugin-typescript2` plugin to the rollup plugins array.

```ts
      bundle.tsconfig && typescript({
```

It sounds like this was a bad assumption.

Is there any way that we can determine whether we should be building for typescript other than scanning for `.ts?` files?

dagda1:
@cowboyd @jnicklas as this output is a warning and not an error, we can silence it by only outputting errors in the plugin of the `rollup-plugin-typescript2`  by setting `verbosity` to `0`;

e.g. 

```ts
typescript({
  tsconfig: bundle.tsconfig,
  tsconfigDefaults: defaultTSConfig(),
  verbosity: 0,
  tsconfigOverride: {
    compilerOptions: {
      module: "ESNext",
    }
  }
}),
```

verbosity has the following settings:

```ts
	- 0 -- Error
	- 1 -- Warning
	- 2 -- Info
	- 3 -- Debug
```

The default is `1`. 

If we change it to `0` the warning is no longer displayed.

 That does mean no warnings which we would want in development unless we can configure it somehow.

cowboyd:
@dagda1 @jnicklas I think we should keep warnings as they're generally indicative of something not being quite right and I hate the idea that it quietly do bad things without giving some indication to the user.

For me, requiring a test tsconfig.json for ts projects isn't terrible, since you can always have it `{"extend": "../tsconfig"}`. It's just a little less magic.

dagda1:
@jnicklas @cowboyd 

I am not up to speed with what files get generated yet, unfortunately.

How does the `tsconfig` setting below get set and is it something that can be relied upon to determine if the current project is typescript or not?

```ts
interface BundleOptions {
  entry: string;
  outFile: string;
  globalName?: string;
  tsconfig?: string;
  watch?: boolean;
};
```

jnicklas:
@dagda1 the `tsconfig` is set by the `init` script normally. The status quo is that typescript will work just fine, using the default tsconfig if this option is not set. So someone could set up the project to use typescript, and *not* set the `tsconfig` option.

https://github.com/thefrontside/bigtest/blob/c8daf4d888ed22f95c8e3fedd7485faaf9ffa409/packages/cli/src/init.ts#L71-L78

Issue #680, opened 2020-11-17
JSX not supported in .js test bundles


cowboyd:
I'm attempting to write a React component harness for BigTest in order to test the `@stripes/components` design system.  My idea is that there will be a `Mount()`  step that takes some JSX and then sends a message to the app that mounts that component:

```ts
import { test } from 'bigtest';
import { KeyValue as Interactor } from '@stripes/testing';
import { Mount } from './helpers';

export default test("KeyValue")
  .step(Mount((React, KeyValue) =>
    (<KeyValue
       data-test-foo="bar"
       label="Label"
       value="Value"
       subValue="subValue"
     />), "KeyValue"))
  .assertion(KeyValueInteractor("Label").exists());
```

This compiles into a function which, when passed to the render app takes the value of `React` and the `KeyValue` component and then returns the corresponding react component.

However, our current bundler doesn't support a custom babel config, so this is a syntax error. We need a way to extend the babel config in our bundler. If we can already do this, then we should document it.

See this repo for details:

https://github.com/cowboyd/stripes-components/tree/integrate-bigtest

taras:
@cowboyd try `.jsx`

cowboyd:
@taras sadly, that doesn't seem to work:

```
[manifest builder] build error:
Unexpected token (Note that you need plugins to import files that are not JavaScript)
```

dagda1:
@cowboyd does this `Mount` function exist yet or is there a branch that is causing the error with the code in your previous comment?

dagda1:
@cowboyd looking at the source code for `@rollup/plugin-babel`, it does search for a `babelrc` by default although if it finds a `babel.config.js`, that takes precedence.

It is a long time since I used babel.  If possible, could you help me get a failing example?

cowboyd:
@dagda1 I pushed the mount helper to this branch (I'd forgotten to check it in)

dagda1:
@cowboyd many thanks.  That is a big help

dagda1:
@cowboyd I made progress on this.

The `@rollup/plugin-babel` does pick up a `babel.config.js` or a `babelrc` automatically and merges it with the config we currently have in the rollup config.

We currently have the rollup `commonjs` plugin in the wrong order in the `plugins` rollup config.  I've created #689 to correct this.

PR #689, opened 2020-11-26
commonjs plugin needs to come after babel plugin

dagda1:
While looking at #680 I discovered that the `commonjs` needs to follow the babel plugin.

This PR moves it down:

```ts
      babel({
        babelHelpers: 'runtime',
        extensions: ['.js', '.ts'],
        presets: ['@babel/preset-env', '@babel/preset-typescript'],
        plugins: ['@babel/plugin-transform-runtime']
      }),
      commonjs(),
```

### DIFF
--- a/.changeset/short-stingrays-deny.md
+++ b/.changeset/short-stingrays-deny.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/bundler": patch
+---
+
+move commonjs down the rollup plugin order
--- a/packages/bundler/package.json
+++ b/packages/bundler/package.json
@@ -20,7 +20,6 @@
     "@babel/core": "^7.10.4",
     "@babel/plugin-transform-runtime": "^7.10.4",
     "@babel/preset-env": "^7.10.4",
-    "@babel/preset-typescript": "^7.10.4",
     "@babel/runtime": "^7.10.4",
     "@bigtest/effection": "^0.5.4",
     "@bigtest/project": "^0.12.0",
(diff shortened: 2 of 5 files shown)


cowboyd:
@dagda1 at least it seems that the ubuntu and windows CLI tests are failing with a consistent error message 🤷🏻‍♂️

dagda1:
@cowboyd I fired this and a couple of other PRs through on Friday without due diligence.  Apologies for that. I'm going to pick them off one by one and take a bit more care.  Not going to rush

I do think this is the correct thing here though in this PR and I've done some more things in the last commit so please have a good look.

- Removed `@babel/preset-typescript`, we should have removed this when we added `rollup-plugin-typescript2`.
- Added a fuller extension list for babel and the `@rollup/plugin-node-resolve` plugins.

I'm going to knock off the PRs I created with a bit more care and one at a time.  I think they all have a good reason but I just fired them off and I should know better at my age :)

taras (approved):
Great catch

cowboyd (approved):
@dagda1 were you able to test this on the runner in stripes-components?

dagda1:
@cowboyd I've had a quick hack by adding the preview package and I had to add a resolution to the main package.json in `stripes-components` 

`"@bigtest/bundler": "0.11.0-74609361",`

I also added this `babel.conifg.js` to stripes also

```json
module.exports = {
  presets: [
    [
      "@babel/preset-react"
    ]
  ]
}
```

The rollup babel plugin merges any babel file it finds with our config in the bundler

The jsx error certainly goes away once I do the above but we would need a proper `Mount` function to test it fully but I am fairly sure it will work now.  

I'm almost tempted to add `"@babel/preset-react"` to our babel-configuration.

Give me a shout tomorrow or anytime if you want to go through it.

We don't need `'@babel/preset-typescript'` anymore now we have the rollup typescript plugin so I'm glad we can remove this.

I have some thoughts about our build configuration in the bundler.

PR #694, opened 2020-11-28
Do not wrap action functions in a convergence

### LINKED ISSUE #691: Interactor convergence in action composition
I was working on documenting the interactors, and I noticed something that we might want to change before the official release. Wanting to get feedback on this. It's a breaking change which is why I think it would be good to do this before the release.

The question is how interactors behave when actions are composed of other actions. Let's look at how Interactor defines an action from the specification on an interactor:

https://github.com/thefrontside/bigtest/blob/c2c4bd119961de7cd1d5295b153b34bcc612467a/packages/interactor/src/create-interactor.ts#L16-L34

Removing all of the unimportant bits, this boils down to:

``` typescript
  function(...args) { 
     return interaction(`click on link`, async () => { 
       return await converge(() => { 
         return action(this, ...args); 
       }); 
     }); 
   }
```

Note that the action is called with the *interactor*, not an actual element. To get an element we need to wrap the action function with the `perform` helper function like this:

``` typescript
const LinkInteractor = createInteractor<HTMLLinkElement>('link')({
  actions: {
    click: perform((element) => { element.click(); })
  },
});
```

The perform really only delegates to [perform](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/interactor.ts#L58-L64) on the interactor. If we inline all of this, for an action which uses `perform`, we'd end up with something like this:

``` typescript
  function(...args) { 
     return interaction(`click on link`, async () => { 
       return await converge(() => { 
         return interaction(`${this.description} performs`, () => {
           return converge(() => {
             let element = this.unsafeSyncResolveUnique();
             element.click();
           });
         });
       }); 
     }); 
   }
```

Hmmm 🤔

Let's look at an interactor which composes multiple actions:

``` typescript
const DatePicker = createInteractor('date picker')({
  actions: {
    selectYeadAndMonth: async (interactor, year: number, month: number) {
      await DatePickerYear.selectYear(year);
      await DatePickerMonth.selectMonth(year);
    }
  }
});
```

This will roughly boil down to:

``` typescript
  function(...args) { 
     return interaction(`select year and month on date picker`, async (year, month) => { 
       return await converge(async () => { 
         await DatePickerYear.selectYear(year);
         await DatePickerMonth.selectMonth(year);
       }); 
     }); 
   }
```

Which we can further expand to:

``` typescript
  function(...args) { 
     return interaction(`select year and month on date picker`, async (year, month) => { 
       return await converge(async () => { 
         await interaction(`select year`, async (year, month) => { 
           return await converge(async () => { ... }
         });
         await interaction(`select month`, async (year, month) => { 
           return await converge(async () => { ... }
         });
       }); 
     }); 
   }
```

Hmmm 🤔

It's kind of hard to reason about this all, but it becomes quite clear that the first example converges twice, which is obviously a bit strange. This is currently not a problem since both convergences have the same timeout, but once we move to timeouts per action (we'll definitely want this at some point), it becomes very problematic. Also this unnecessary convergence is confusing, and makes it even harder to reason about what's actually going on.

The second example is the real problem though. In my opinion this is simply broken. Wrapping the async function in a convergence will cause it to run the inner convergences in the delegated interactors for each iteration of the convergence loops. We'll end up with `n` convergences, where `n` is the number of times the outer convergence loop runs.

### A possible solution

We could change the definition of `perform` to something similar to this:

``` typescript
export function perform(fn: (element, ...args) => void) {
  return converge(() => {
    let element = interactor.unsafeSyncResolveUnique();
    fn(element, ...args);
  });
}
```

Does that mean we should rename it to something else?

And we would *not* wrap the action in a convergence when defining the action, like this:

``` typescript
  function(...args) { 
     return interaction(`click on link`, async () => { 
       return await action(this, ...args); 
     }); 
   }
```

jnicklas:
Closes #691

This removes the automatically inserted convergence for action functions, which means that actions do not converge automatically anymore. Instead the user will need to ensure they converge by awaiting a converging function inside the action. This could be either an action/assertion on another interactor, or it could be one of the built in converging functions on the interactor itself. Most likely, the user will want to call `perform`. To remove some boilerplate for this, we have the `perform` function, which most actions already use, so this will not be a breaking change for most interactors.

This also means that "normal" actions which use `perform` to not nest calls of converge. Additionally, this change explicitly awaits the converge call in the `perform` helper function, removing the useless wrapping in an interaction as well.

I've been wracking my brain trying to come up with a way of testing this, but any test for this change would really be testing on the absence of something, e.g. testing that the async function is *not* exectuted multiple times, which seems like a bit of a silly test to me.

### DIFF
--- a/.changeset/pink-flowers-grab.md
+++ b/.changeset/pink-flowers-grab.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/interactor": minor
+"bigtest": minor
+---
+
+Interactor actions are not automatically wrapped in a convergence
--- a/packages/interactor/src/create-interactor.ts
+++ b/packages/interactor/src/create-interactor.ts
@@ -4,7 +4,6 @@ import { Locator } from './locator';
 import { Filter } from './filter';
 import { Interactor } from './interactor';
 import { interaction } from './interaction';
-import { converge } from './converge';
 
 const defaultLocator: LocatorFn<Element> = (element) => element.textContent || "";
 
@@ -23,9 +22,7 @@ export function createInteractor<E extends Element>(interactorName: string) {
             if(bigtestGlobals.runnerState === 'assertion') {
               throw new Error(`tried to ${actionDescription} on ${this.description} in an assertion, actions should only be performed in steps`);
             }
-            return await converge(() => {
-              return action(this, ...args);
-            });
+            return action(this, ...args);
           });
         },
         configurable: true,
--- a/packages/interactor/src/perform.ts
+++ b/packages/interactor/src/perform.ts
@@ -1,10 +1,10 @@
 import { Interactor } from "./interactor";
 import { Filters, Actions } from "./specification";
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function perform<E extends Element, F extends Filters<E>, A extends Actions<E>, T extends any[]>(fn: (element: E, ...args: T) => void) {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return (interactor: Interactor<E, F, A>, ...args: T) => interactor.perform(element => {
-    fn(element, ...args);
-  });
+export function perform<E extends Element, F extends Filters<E>, A extends Actions<E>, T extends unknown[]>(fn: (element: E, ...args: T) => void) {
+  return async (interactor: Interactor<E, F, A>, ...args: T) => {
+    return await interactor.perform(element => {
+      fn(element, ...args);
+    });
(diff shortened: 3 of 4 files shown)


cowboyd (approved):
Should we explicitly export convergence now that it's conceivable you might need it?

pittst3r (approved):

jnicklas:
@cowboyd maybe? It feels like you should probably be able to get by with just using `perform`, but there may be rare cases where manually calling `converge` is better.

PR #697, opened 2020-12-01
Introduce Sample app to replace TodoMVC

### LINKED ISSUE #703: Add the sample app to the monorepo
Min has a sample app that people can copy to their own machines and run locally like `npx bigtest-sample-app`, which has an experience similar to create react app. We can add this sample app to the monorepo without adding to the CI install burden by omitting it from the list of packages in the workspace.

minkimcello:
### BigTest Sample App
- Moved all of the app-related files into the `app/` directory
  - Package is named `bigtest-sample-app` and is marked `private`
  - The temporary `package.json` file is labelled `app-pkg.json` and it gets renamed back to `package.json` during `npx`
- Added new and super minimal README with instructions on which test commands are available
### BigTest Sample (dx)
- Dev start script has been modified to adapt to the new directory layout and works for both `npm` and `yarn`
- The `setup.js` script has been renamed `create-sample-app.js` to be more descriptive for future devs
- When running `npx bigtest-sample`, `migrate()` will copy over ALL files from `/_npx/cache/bigtest/sample/app/* with the exception of `app-pkg.json` which gets copied over as `package.json`
  - Currently both `package-lock.json` and `yarn.lock` are being published but I'll get rid of `yarn.lock` when we finally migrate the entire monorepo to `npm`
- Added dx README

## ~Latest~ Old Updates
- Created a custom start script for DX. When consuming sample app via `npx bigtest-sample`, the DX start script will be replaced with a normal production start script.
  - This is so that the other packages can `yarn --cwd ../sample start` instead of having to `yarn --cwd ../sample install && start`.
    - This shouldn't be a blocker for this PR since we're going to rewrite most of those scripts using effection.
- ~~Sample app directory is at the root of `packages/` as we discussed. It'll be left out of the packages workspaces but we decided putting it into integrations wasn't accurate.~~ Sample app is now at the root of the monorepo. Didn't belong in packages and didn't belong in integrations.
- Undid the test refactors so `cli`, `server`, and `interactors` still relies on todomvc. The tests will be refactored and todomvc removed in separate PRs.
  - The demo video on bigtest platform shows todomvc so that may need to be updated if we think it's necessary.
- Ran `npx bigtest-sample@${preview_sha}` to confirm the app is being installed correctly and all three test runners are passing.

## Motivation

This app will serve two purposes:
  1. For people to install and play around with it to see how Interactors work with Jest, Cypress, and the BigTest platform.
  2. To replace `@bigtest/todomvc`.
    - :warning: But this will be further down the road as we'll need to update the other packages' tests before we can remove `todomvc` from the monorepo.

Screen Shot 2020-12-03 at 6 10 27 PM

Resolves #703

## Approach
- I'm calling it `bigtest-sample` because it's meant to just be a sample; like the sample you'd get at Costco. Example implies maybe a more complex demonstration of its full potential.
- Simple React app that toggles between two buttons.
- Added `Cypress` and `BigTest` platform with very minimal tests for both and `Jest` as well.
- In Discord we discussed not using `react-scripts` but for the convenience of getting this up and running I started the app with it. We can refactor to webpack later if need be.
  - But I was thinking we can build and push the `dist/` directory so that when other packages need to test against it, we can just link the packages locally so that the sample app does not need to be installed and built in CI.
- The package itself has not been added to the monorepo config so it won't install or build itself in CI. It's just sitting in this repository for the sake of tidiness and not having yet another repository in the frontside org.

## TODO
- [x] Working on integrating `npx bigtest-sample` similar to `create-react-app`.
- [x] Pretty up the console logs using `chalk`.
- [x] ~~Put output from yarn/npm install into a small animated textbox~~
  - :warning: the way in which NPM implements their output is different from Yarn and so instead of trying to do something fancy, I changed the loading text from `this might take a few minutes` to `this WILL take a few minutes`.
- [x] ~~Create custom prepublish script so that it can populate the secondary package.json on its own and thus developers won't need to update the versions themselves~~
- [x] Create README
- [x] Remove react scripts and use parcel instead
- [x] ~~Replace todomvc with sample app in interactor/cypress test~~
- [x] ~~Replace todomvc with sample app for cli~~
  - :warning: There was a deprecated flag and therefore giving a false positive in one of the tests. I commented it out for now.
- [x] Replace todomvc with sample app for server/orchestrator test
- [x] ~~Last step: deprecate todomvc app to trigger deprecate functionality in preview workflow and then remove the directory~~ :warning: do this in a separate pr. removing todomvc will cause all sorts of eslint errors.


minkimcello:
@cowboyd @jorgelainfiesta @jenweber 

Can I get some help on what we should display when a user tries to install `bigtest-sample` for the second time? Saying "hey, we see this isn't your first time, if you need help, reach out to us" seems a little too informal.

What do you think of this?
```
$ npx bigtest-sample
npx: installed 25 in 3.775s

Error: directory 'bigtest-sample' already exists

Please rename the existing `bigtest-sample` directory and rerun the command.
```

cowboyd:
@minkimcello we could also have it say something like "this directory already exists. Do you want to overwrite and start from scratch?'

minkimcello:
@cowboyd I considered that but wasn't sure how i felt about rimraf-ing users' local files (even with their input) so I thought it would be a safer approach to just ask them to rename or delete it themselves.

minkimcello:
@cowboyd Should I have the bigtest dependency of the sample app as `*` so that developers won't need to bump it themselves?

cowboyd (commented):
Would it make sense to put this in an integrations package with `@bigtest/cypress` and `@bigtest/jest` since it would depend on them? #759

sample/bin/setup.js
```diff
@@ -0,0 +1,119 @@
...
+  yield populate(messages.creating_dir);
+  try {
+    yield install(messages.downloading_repo);
+    yield migrate(messages.organizing_files);
+    yield install(messages.installing_dep);
+    console.log(messages.success);
+  } catch(e) {
+    yield clean(e);
+  };
+};
```
minkimcello: @cowboyd I brought out `populate()` from the try/catch and gave it its own `MainError()` for when the directory already exists and converted the old `abort()` function into `clean()`. But I wasn't sure whether or not `clean()` should be async and yielded.

sample/bin/setup.js
```diff
@@ -0,0 +1,95 @@
...
+
+function* download(messages) {
+  yield spin(messages.before, install({ cwd: TARGET_DIR }));
+  console.log(formatSuccess(messages.after));
+};
+
+function* run() {
+  let rollback = true;
+  yield populate(messages.creating_dir);
```
cowboyd: Note that you won't rollback if `populate()` fails or you halt here because you won't have entered the `try{}` is that intentional?

cowboyd (commented):
@minkimcello This is looking really good. How does it feel?

sample/bin/setup.js
minkimcello: > This is looking really good. How does it feel?

Very excited to finally learn how effection works. 👍 I don't mind doing a rewrite of this in Typescript but should we scope that to a separate PR or do that as the next step?

> Note that you won't rollback if populate() fails or you halt here because you won't have entered the try{} is that intentional?

populate() will throw an error if `sample-app` directory already exists. so i think forcing a rollback on that function might unintentionally delete someone's previous progress/experiments with bigtest. i could either keep it as is or throw in `rollback = false` inside the if block so that the entire function can move into the try/catch:
```
if (fs.existsSync(TARGET_DIR)) {
  rollback = false;
  throw new MainError({ 
    message: `${formatErr('directory \'bigtest-sample\' already exists')}\n${messages.abort}`
  });
}
```

sample/bin/install.js
```diff
@@ -0,0 +1,20 @@
...
+
+function* install({ cwd, stdio }) {  
+  let command = yarn ? 'yarn' : 'npm';
+  let install = yield exec(`${command} install`, { cwd });
+  if(stdio === 'inherit'){
+    yield spawn(subscribe(install.stdout).forEach((data) => {
+      process.stdout.write(data);
+      return Promise.resolve();
+    }));
+  };
```
cowboyd: We should also pipe `stderr` if we're inheriting.
```suggestion
  if(stdio === 'inherit') {
    yield spawn(subscribe(install.stdout).forEach((data) => {
      process.stdout.write(data);
      return Promise.resolve();
    }));
    yield spawn(subscribe(install.stderr).forEach((data) => {
      process.stderr.write(data);
      return Promise.resolve();
    }));
  };
```

minkimcello: done 👍

sample/bin/setup.js
```diff
@@ -0,0 +1,73 @@
...
+        };
+  
+        fs.writeFileSync(`${TARGET_DIR}/package.json`, JSON.stringify(pkgjson, null, 2));
+      } else if(file !== 'bin') {
+        fs.renameSync(`${SOURCE_DIR}/${file}`, `${TARGET_DIR}/${file}`);
+      };
+    }));
+  });
+  console.log(formatSuccess(messages.after));
+};
```
minkimcello: @cowboyd with the help of `path.dirname(__dirname)` as `SOURCE_DIR` the script now grabs the necessary files from the cache location of the npx install; removing the need for the middle install step.

However, with effection in our dependencies, I've been seeing this install error. It doesn't affect the result of the install script but still:
Screen Shot 2021-01-14 at 9 45 27 PM

sample/bin/constants.js
```diff
@@ -0,0 +1,33 @@
+const chalk = require('chalk');
+const path = require('path');
+
+const startScript = 'parcel src/index.html';
+const PACKAGE_NAME = 'bigtest-sample';
+const TARGET_DIR = `${process.cwd()}/${PACKAGE_NAME}`;
+const SOURCE_DIR = path.dirname(__dirname);
```
cowboyd: Let's inline these as local variables. They aren't overridable, and so having them in a `constants` file only serves to remove them from the context in which they're used, making it hard for the reader to understand.

having the messages in a separate file can make sense though

sample/bin/setup.js
cowboyd: *sigh* Ok, looks like a mini-yak that needs shaving. Let me have a look

cowboyd: Ok, looks like the problem is that we didn't test our install script on npm, just yarn for `ctrlc-windows`. I think the same error is occuring, but yarn just sits on it, while npm is noiser about it. I think I have a fix...

sample/bin/start.js
```diff
@@ -0,0 +1,36 @@
...
+    port = process.argv.indexOf('-p');
+    args.push(...process.argv.slice(port, port + 2))
+  };
+
+  let { stdout } = yield daemon(`${command} ${args.join(' ')}`);
+
+  yield spawn(subscribe(stdout).forEach((data) => {
+    process.stdout.write(data);
+    return Promise.resolve();
+  }));
```
cowboyd: We'll want stderr here too.

sample/bin/setup.js
cowboyd: https://github.com/thefrontside/ctrlc-windows/pull/19

cowboyd (commented):
Alright, we're really coming along here, and the code is looking good.

One thing that remains is to decouple the package that creates the sample app in the target dir from the app that is itself created in the target dir:

* `bigtest-sample`: this is a cli utility for creating the sample app. It is not itself a webapp, nor does it have a dependency on bigtest. Rather, it's only dependencies are the things that it needs to perfom the install like effection/node, chalk, etc...
* `bigtest-sample-app`: this is a webapp which happens to be contained in the `bigtest-sample` npm tarball and which `bigtest-sample` copies into the target directory. This has bigtest, cypress, jest, etc... in its dev dependencies, but it does not have a dependency on chalk because chalk has nothing to do with the sample.

minkimcello:
> Alright, we're really coming along here, and the code is looking good.
> 
> One thing that remains is to decouple the package that creates the sample app in the target dir from the app that is itself created in the target dir:
> 
>     * `bigtest-sample`: this is a cli utility for creating the sample app. It is not itself a webapp, nor does it have a dependency on bigtest. Rather, it's only dependencies are the things that it needs to perfom the install like effection/node, chalk, etc...
> 
>     * `bigtest-sample-app`: this is a webapp which happens to be contained in the `bigtest-sample` npm tarball and which `bigtest-sample` copies into the target directory. This has bigtest, cypress, jest, etc... in its dev dependencies, but it does not have a dependency on chalk because chalk has nothing to do with the sample.

I changed the name value of the package.json the setup script creates to `bigtest-sample-app` so the package in the monorepo remains `bigtest-sample` and the project that is created via `npx bigtest-sample` will be `bigtest-sample-app`, is that what you meant?

Should I have jack publish `bigtest-sample-app` so that the name belongs to us?

sample/bin/constants.js
```diff
@@ -0,0 +1,23 @@
+const chalk = require('chalk');
```
cowboyd: let's rename this file "messages.js" now since it's descriptive of what it is

cowboyd (commented):
Alright, here we go, I bet this is the last round

the final countdown from arrested development

cowboyd (approved):
nice work @minkimcello!
🌟 🙌🏻 🌟

PR #698, opened 2020-12-02
Bump chromedriver/geckodriver versions

jnicklas:
And down constrain them upwards, most likely we want these to be freely upgradeable.

### DIFF
--- a/.changeset/little-bulldogs-prove.md
+++ b/.changeset/little-bulldogs-prove.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/webdriver": patch
+"bigtest": patch
+---
+
+Upgrade chromedriver and geckodriver versions and loosen their constraints
--- a/packages/webdriver/package.json
+++ b/packages/webdriver/package.json
@@ -33,9 +33,9 @@
     "@effection/fetch": "^0.1.2",
     "@effection/node": "^0.9.1",
     "abort-controller": "^3.0.0",
-    "chromedriver": "~86.0.0",
+    "chromedriver": ">=87.0.0",
     "effection": "^0.7.0",
-    "geckodriver": "^1.19.1",
+    "geckodriver": ">=1.21.0",
     "node-fetch": "^2.6.0"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -4207,10 +4207,10 @@ adjust-sourcemap-loader@2.0.0:
     object-path "0.11.4"
     regex-parser "2.2.10"
 
-adm-zip@0.4.11:
-  version "0.4.11"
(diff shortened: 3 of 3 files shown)


cowboyd (approved):

cowboyd:
🤣 looks like the chrome version is 87 on ubuntu, but still 86 on windows

jnicklas:
> 🤣 looks like the chrome version is 87 on ubuntu, but still 86 on windows

Ugh.... So what do we do here?

cowboyd:
> Ugh.... So what do we do here?

It sounds like this is a bug in the virtual environments https://github.com/actions/virtual-environments/issues/1275 since if I read correctly, they should release with the latest version of chrome weekly. The fact that 87 was not picked up on windows would indicate that something didn't go quite right wit their image builder.

We can either wait for that fix, or maybe we can run an installer on the windows environment, or use a custom image? @jbolda do you have any experience using custom windows images on github actions?

jbolda:
I haven't tried to use a custom image at all, but I imagine that isn't a great long term solution? Particularly when considering that Github Actions seems to be the low bar in terms of setting up CI/CD, and we don't want to turn people away with a complicated setup for bigtest.

I assume since we are switching to `>=` that bigtest doesn't _really care_ about the exact version, just that it works and the user has whatever version that they need for their feature set (which most of the time probably doesn't matter either). In userland, they will install and get a version in their lockfile. On CI though, I wonder if we (internally and for users) push towards [using the system chromedriver](https://github.com/giggio/node-chromedriver#skipping-chromedriver-download) which should also be compatible with the system chrome.

cowboyd:
> I haven't tried to use a custom image at all, but I imagine that isn't a great long term solution? Particularly when considering that Github Actions seems to be the low bar in terms of setting up CI/CD, and we don't want to turn people away with a complicated setup for bigtest.
> 
> I assume since we are switching to `>=` that bigtest doesn't _really care_ about the exact version, just that it works and the user has whatever version that they need for their feature set (which most of the time probably doesn't matter either). In userland, they will install and get a version in their lockfile. On CI though, I wonder if we (internally and for users) push towards [using the system chromedriver](https://github.com/giggio/node-chromedriver#skipping-chromedriver-download) which should also be compatible with the system chrome.

@jbolda This isn't so much for people using bigtest in their own CI, but only for our CI. I think that the solution is that we will end up having people use _remote_ webdrivers that are running in an image. So, for example, we would say recommend that you use one of the public webdriver images in your CI and use it as a service. However, this isn't an option for us, because we actually need to unit test the _local_ webdrivers in CI which in the longer term will only really be used in development.

cowboyd:
@jbolda I think the best option is your suggestion to use the system chromedriver if we can

jbolda:
May have the added benefit of speeding up test runs too.

cowboyd:
Well, looks like in the intervening time, they updated the windows image. Until next time we see you chromedriver update bug!

🥫🏃🏻‍♂️ 

(although we should still use the system chromedriver at some point)

PR #699, opened 2020-12-02
Allow starting agents from remote webdriver servers

cowboyd:
Motivation
-----------

Currently, it is assumed that BigTest will manage your webdriver server in-process, both starting it and stopping it on demand. In fact, for things like geckodriver which only allows a single browser to be run per-server, there is a server process per web agent.

This causes a bunch of issues because it means that the bigtest user has to manage their own browser binaries since their server must ultimately start and stop browser processes. This change allows bigtest to create agents directly off of an existing webdriver server. Among other things, this opens the door to running browsers completely off-container in things like BrowserStack or Saas labs, but it could also be running off of a server started in the container before the test run.

Approach
----------
To do this, the driver options have been expanded to account for both "remote" and "local" options. The type of driver is used to resolve the url of the driver server before ultimately using that url to connect and disconnect new web drivers.

### DIFF
--- a/packages/webdriver/src/web-driver.ts
+++ b/packages/webdriver/src/web-driver.ts
@@ -4,6 +4,7 @@ import { fetch } from '@effection/fetch';
 import { Driver } from '@bigtest/driver';
 
 export class WebDriver implements Driver<WDSession> {
+
   session: WDSession = { sessionId: '' };
 
   constructor(public serverURL: string) { }
@@ -23,6 +24,11 @@ export class WebDriver implements Driver<WDSession> {
       url,
     });
   }
+
+  //we need this to synchronize in testing because
+  //async finally blocks are not yet supported. It
+  //should not be used anywhere.
+  active = false;
 }
 
 /**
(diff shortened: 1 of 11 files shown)


packages/webdriver/src/web-driver.ts
```diff
@@ -34,7 +40,7 @@ export function* connect(driver: WebDriver, options: Options): Operation<void> {
 
   let capabilities = new Atom(Capabilities);
 
-  if (options.headless) {
+  if (options.headless != null && options.headless) {
```
dagda1: is that an intentional `!=` instead of `!==`?

would `!!options.headless` work instead?

I would have the [eqeqeq](https://eslint.org/docs/rules/eqeqeq) rule set in eslint for strict equals

packages/webdriver/src/web-driver.ts
```diff
@@ -44,37 +50,61 @@ export function* connect(driver: WebDriver, options: Options): Operation<void> {
...
+  browserName: 'chrome' | 'firefox' | 'safari';
+};
+
+export type RemoteOptions = {
+  type: 'remote';
+  url: string;
+  headless?: boolean;
+};
+
+export type Options = LocalOptions | RemoteOptions;
```
dagda1, line 106: Good use of a discriminated union or adt!

dagda1 (approved):
This is really great.

packages/webdriver/src/web-driver.ts
cowboyd: I had done this originally so that specifying the boolean option wasn't _required_ so only do the check if it is not undefined, but I guess if it's undefined it will have the same effect as being false.

PR #700, opened 2020-12-03
Fix race condition in chromedriver

jnicklas:
Making an attempt to fix this.

### DIFF
--- a/.changeset/hungry-monkeys-confess.md
+++ b/.changeset/hungry-monkeys-confess.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/webdriver": patch
+"bigtest": patch
+---
+
+Fix race condition in chromedriver start
--- a/packages/webdriver/src/local.ts
+++ b/packages/webdriver/src/local.ts
@@ -35,7 +35,7 @@ export function * Local(options: Options): Operation<WebDriver> {
     yield;
   });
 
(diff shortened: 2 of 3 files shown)


cowboyd:
@jnicklas maybe disentangle the chromedriver upgrade which is failing on windows, so we can get this one through?

jnicklas:
Locally this seems to improve things for me, but I think we really need to release this and upgrade bigtest on cvex to see whether it actually works.

cowboyd (commented):
What if we just removed the exterior timeout entirely? We only have two times we use this, in our mocha tets, and in our CLI, both of which are subject to external timeouts. 

Internally, we would still have an iteration timeout that had a linear backoff or something.

If we leave the timeout in here, we're just moving the booby trap from one corner of the room into another.

cowboyd:
This also would make it compatible with a higher order operation whose only puprose is to timebox something. Operations with implicit timeouts are inherently un-composable.

jnicklas:
@cowboyd I think that's a good idea. I've removed the global timeout, but kept the per-request timeout. Since that timeout is caught, it shouldn't bubble up and cause a failure, it will only cause the request to abort and be retried, which is what we want to do to make the startup as fast as possible.

cowboyd (approved):

PR #709, opened 2020-12-04
chore: remove lodash resolution and sync lockfile

jbolda:
lodash updated and we specify the updated version in other packages, but didn't remove the resolution in the root. Hopefully the new version publishing issues are resolved on their end now.


cowboyd (approved):
If the check is green, the code is clean

PR #711, opened 2020-12-04
[interactor] focus action on all elements

jbolda:
This extracts out the focus action and imports it into each interactor (and adds it to a few that were missing it).

### DIFF
--- a/packages/interactor/src/focused.ts
+++ b/packages/interactor/src/focused.ts
@@ -7,10 +7,32 @@ export function focused(element: Element): boolean {
   return element.ownerDocument.activeElement === element;
 }
 
+/**
+ * Helper function for focus action, performs `element.focus()` on the selected element.
+ * 
+ * Can be used with object property value shorthand in your interactor actions.
+ * 
+ * ```js
+ * actions: {
+ *   focus,
+ * }
+ * ```
+ */
 export const focus = perform(<E extends HTMLElement>(element: E) => {
     element.focus();
 });
 
+/**
+ * Helper function for blur action, performs `element.blur()` on the selected element.
+ * 
+ * Can be used with object property value shorthand in your interactor actions.
+ * 
+ * ```js
+ * actions: {
+ *   focus,
+ * }
+ * ```
+ */
 export const blur = perform(<E extends HTMLElement>(element: E) => {
     element.blur();
 });
--- a/packages/interactor/src/focused.ts
+++ b/packages/interactor/src/focused.ts
@@ -1,5 +1,4 @@
 import { perform } from './perform';
-import { ActionFn } from './specification';
 
 /**
  * Helper function for focused filters, returns whether the given element is focused.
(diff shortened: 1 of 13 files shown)


packages/interactor/src/focused.ts
```diff
@@ -1,6 +1,19 @@
...
+import { ActionFn } from './specification';
+
 /**
  * Helper function for focused filters, returns whether the given element is focused.
  */
 export function focused(element: Element): boolean {
   return element.ownerDocument.activeElement === element;
 }
+
+export const focus: ActionFn<any> =
```
jbolda: This (and the blur) have an `any` type that I am struggling to narrow. Would appreciate any ideas to address that.

jnicklas: I think `HTMLElement` is the appropriate type here instead of `any`. That's what you should use, but I'm not sure if there will be any problems due to subtyping/variance on generics, so TypeScript might reject it.

packages/interactor/src/focused.ts
```diff
@@ -1,6 +1,19 @@
...
+  perform(<T extends keyof HTMLElementTagNameMap>(element: HTMLElementTagNameMap[T]) => {
```
jnicklas: Why this complicated generic type? Can we leave this to be inferred?

packages/interactor/src/focused.ts
jbolda: `<E extends HTMLElement>` was the ticket here. Thanks!

cowboyd:
Interesting that this changed to interactor, triggered a preview of every package /cc @minkimcello

packages/interactor/src/focused.ts
```diff
@@ -1,6 +1,16 @@
...
   return element.ownerDocument.activeElement === element;
 }
+
+export const focus = perform(<E extends HTMLElement>(element: E) => {
+    element.focus();
+});
+
+export const blur = perform(<E extends HTMLElement>(element: E) => {
+    element.blur();
+});
```
jnicklas: Since these become public API, maybe we could add some documentation for them?

jnicklas (approved):
Approved, but would be good to add some docs as well before merging :)

minkimcello:
> Interesting that this changed to interactor, triggered a preview of every package /cc @minkimcello

Hmm I'm not sure why it did that either. I'll keep an eye out to see if it does that for every PR.

minkimcello:
@cowboyd this is a wild guess but I think it might have published all preview packages because we merged a changeset PR recently and this branch might not have rebased yet.

packages/interactor/src/focused.ts
jbolda: 👍

jbolda:
Added a few doc notes (and did a rebase off of `v0` fyi @minkimcello ).

cowboyd (approved):

PR #712, opened 2020-12-06
[Atom] replace ramda lens with monocle-ts

dagda1:
This PR as the name suggests replaces the ramda Lens with the more complete monocle-ts Lens.

I have replaced the `Atom<S>` and the `Slice<S, A>` classes with a `createAtom<S>` that returns an `Atom<S>`.

A `Slice<A>` can be created from the `Atom<S>` with only 1 type argument now:

```ts
let atom = createAtom(state);

let status = subject.slice()('agents', "agent-2", "status");
```

`slice` is now a higher-order function.

The following example illustrates how an `Atom` behaves similarly with and without actual state.

```ts
let slice = atom.slice()("agents", "agent:1234", "platform", "vendor"); // does not exist
slice.set("Windows"); // does not error
console.log(slice.get()); // undefined

let atom2 = createAtom<TestRunState>(example);
let slice2 = atom2.slice()("agents", "agent-1", "platform", "vendor");
slice2.set("Windows");

console.log(slice2.get()); // windows

let slice3 = atom2.slice()("agents", "agent-2", "status"); // 'pending'

slice3.update((prev) => {
  console.log(prev); // pending

  return "running";
});

console.log(slice3.get()); // running
```


.changeset/six-flowers-smash.md
```diff
@@ -0,0 +1,15 @@
+---
+"@bigtest/agent": minor
+"@bigtest/atom": minor
+"bigtest": minor
```
cowboyd: Let's make "@bigtest/atom" a minor, and "bigtest" a patch. Everything else will get a patch because it depends on atom and it got a minor bump.

packages/atom/src/atom.ts
```diff
@@ -1,59 +1,170 @@
...
-    this.initial = this.state = initial;
-  }
+import { Operation } from 'effection';
+import { Atom, Sliceable } from './sliceable';
+import { assert } from  'assert-ts';
 
-  setMaxListeners(value: number) {
-    this.states.setMaxListeners(value);
-  }
+export function createAtom<S>(init?: S): Atom<S> {
```
cowboyd: Any reason that we're implicitly making `init` nullable? In the event that you need a nullable initial value, why not force the type parameter to be something like `OrchestratorState | undefined`?

packages/atom/src/atom.ts
```diff
@@ -1,59 +1,170 @@
...
+    }
+
+    state = pipe(
+      current,
+      lens.set(value)
+    );
+
+    if(state) {
+      states.send(state);
+    }
```
cowboyd: A null state is still noteworthy, no? Seems like the only case where we want to omit sending a new value is when the state is the same as `current`?

packages/atom/src/sliceable.ts
```diff
@@ -1,34 +1,115 @@
...
+  set(value: S): void;
+  update(fn: (state: S) => S): void;
+  slice(): Sliceable<S>;
+  once(predicate: (state: S) => boolean): Operation<S>;
+  remove(): void;
+  over(fn: (value: S) => S): void;
+}
+export type Atom<S> = Omit<Slice<S>, 'remove' | 'over'> & {
+  reset(initializer?: (initial: S, curr: S) => S): void;
+  setMaxListeners(value: number): void;
```
cowboyd, line 17: I feel like we have a lot of duplication just to support these two methods on Atom. I wonder what it would take to get by without these and then we could just have create Atom return a slice directly.

cowboyd (commented):
Wow, a lot to process here. I'd love it if you could walk me through this at some point... but this is looking phenomenal 💪🏻 

The curried api is a bit strange. I'm wondering if it might be feasible down the road to make slice a getter?

```ts
atom.slice()('appService', 'status')
//vs
atom.slice.appService.status
//vs
atom.slice(s => s.appService.status)
```

dagda1:
> Wow, a lot to process here. I'd love it if you could walk me through this at some point... but this is looking phenomenal 💪🏻
> 
> The curried api is a bit strange. I'm wondering if it might be feasible down the road to make slice a getter?

I think as a first step, I could just make `slice` a function like it used to be.  I was keeping it in line with monocle which has all its methods curried.

So instead of 

`.slice()('info', 'employment', 'phone', 'number')`

We just have

`.slice('info', 'employment', 'phone', 'number')`

Would make the PR smaller also.

packages/atom/src/sliceable.ts
jnicklas: I'm not a huge fan of `reset` anyway, it's a symptom of problems in bigtest that we're trying to hack around. As for `setMaxListeners`, we could just set that number very high and hope for the best 🙈

jnicklas:
I'd personally prefer this without the curried API. As far as I can tell the API without currying would behave identically? If so I think we should definitely not have that additional method call.

dagda1:
> I'd personally prefer this without the curried API. As far as I can tell the API without currying would behave identically? If so I think we should definitely not have that additional method call.

@jnicklas @cowboyd   I agree.  I was keeping it in line with monocle but we don't have a good reason to do it so we probably should not.

packages/atom/src/sliceable.ts
dagda1: @jnicklas @cowboyd I agree about `setMaxListeners`. I only ever remember seeing an error about this when I had a memory leak with an EventEmitter and too many were getting created. 

I think the code can really be simplified more but not for this PR, I wanted to keep it as close to the original API as possible but I think the code surface area could be quite small.

I think I can get rid of the horrible overloads with typescript 4.1.2 also if we upgrade.

dagda1:
> I'd love it if you could walk me through this at some point..

@cowboyd absolutely, let us do that

jnicklas (approved):

packages/atom/src/sliceable.ts
dagda1: @cowboyd @jnicklas I sort of feel there is unnecessary duplication in this current PR, I think the `createAtom` function could just create a `Slice` using the `sliceMaker` HOF and have 1 set of `get`, `set` etc .

It would be exactly the same API but get rid of the duplication that is nagging at me.

Maybe that is one for a future PR or maybe....just go with this?

cowboyd (approved):
@dagda1 Let's go for it!

Follow on work to do right after this:

1. fold in the `setMaxListeners` to the `createAtom` API which should really never be needed, but which we provide as an escape hatch.
2. make `reset` a hidden API so that it can be used in testing, but is not part of the public API of what is returned.
3. de-dupe the atom/slice api and just have createAtom() return a top-level slice with the sliceMaker
4. extract to `@effection/atom`

Do you want me to create tickets for these?

dagda1:
@cowboyd nice one!  i'll not sleep properly until i've removed that duplication

dagda1:
@cowboyd i don't think we need separate tickets for these.

PR #715, opened 2020-12-08
feat: typedoc to docusaurus

jbolda:
This is an almost working implementation of typedoc (to markdown) to docusaurus. This was made more difficult by the fact that we are in a monorepo and that doesn't seem well tested. It seemed the best path forward was to use the plugin multiple times, but there is (what I would consider) a bug that will not allow it (see the `patch-package`). The `typedoc-markdown-plugin` also does not work on the `0.20.x` version due to breaking changes. There have been recent commits that seems to imply that it will work eventually though, and `0.20.x` is still in beta.

@jnicklas it seemed that the reason for `0.20.x` was mostly styling. Is that a fair statement? I think we might be fine on `0.19.2`.

My remaining TODO is dealing with a bunch of formatting / parsing issues from mdx. Example:
```shell
SyntaxError: C:\Users\Jacob\Documents\dev\frontside\bigtest\website\docs\docs\api\suite\types\steplist.md: Expected corresponding JSX closing tag for <C> (10:38)

   8 | <hr></hr>
   9 | <p>{`id: "steplist"
> 10 | title: "Type alias: StepList`}<C>{`"`}</p>
     |                                       ^
  11 | <h2 {...{"id":"sidebar_label-steplist"}}>{`sidebar_label: "StepList"`}</h2>
  12 | <hr></hr>
  13 | <p>{`id: "steplist"
```


jbolda:
So it seems that there is an issue with the `docusaurus-plugin-typedoc` that prevents it from running the plugin multiple times. Since we are mostly focused on get interactor out, I am running on that and dumping it verbatim into the `docs/api` folder. We may want to attend to how the functions come out (and public vs private APIs through typedoc hopefully?), but we can work on that in further PRs.

The formatting / parsing issue noted in the OP is actually the frontmatter being produced multiple times due to my edits allowing it to run the plugin multiple times. I opened an issue in that repo to see if we can get to the bottom of it.

jorgelainfiesta:
I'm not sure I understand completely what is going on. In my opinion, we only need the Interactors API for now. Is that something that we can achieve at the moment? Anything else can be addressed in the future.

jbolda:
@jorgelainfiesta that was essentially our assessment as well. You can see how the API shows up here: https://deploy-preview-715--bigtest.netlify.app/docs/interactors/api

<blockquote><div><strong><a href="https://bigtestjs.io/docs/interactors/api">@bigtest/interactor | BigTest</a></strong></div><div>Interactors</div></blockquote>

jorgelainfiesta:
Got it, cool! Can we hand pick which pages we generate? there's quite a few that seem to be more noise 

I think the most valuable part is the `Variables` category, because it contains the interactors. However, even here it's confusing because they all have like two entries (i.e. `Button` and `ButtonInteractor`)

website/docs/.gitignore
```diff
@@ -7,6 +7,8 @@
 # Generated files
 .docusaurus
 .cache-loader
+docs/**/api
+sidebars
```
jorgelainfiesta: why are `sidebars` ignored?

website/docs/sidebars.js
```diff
@@ -4,7 +4,12 @@ module.exports = {
     'interactors/built-in-dom',
     'interactors/locators-filters-actions',
     'interactors/write-your-own',
-    'interactors/integrations'
+    'interactors/integrations',
+    {
+      "type": "category",
+      "label": "API",
+      "items": require('./sidebars/typedoc-interactor-sidebar')
```
jorgelainfiesta: I'm curious to know how this is defined, could be an option to make things look more organized (I'm up for writing a manual sidebar for the Interactors API for instance)

website/docs/docusaurus.config.js
```diff
@@ -96,5 +96,78 @@ module.exports = {
   ],
   stylesheets: [
     'https://use.typekit.net/gyc5wys.css'
+  ],
+  plugins: [
+    // [
+    //   'docusaurus-plugin-typedoc',
```
jorgelainfiesta: Are we leaving these comments as reminders somehow?

website/docs/.gitignore
jbolda: The `docusaurus-plugin-typedoc` will create the sidebar files for you based on the output from typedoc. I needed a folder for the build to dump all of the sidebars into that wasn't part of the main docs and won't be committed.

website/docs/sidebars.js
jbolda: Mentioned in the previous comment, it is built by the plugins. I have to imagine there is some customizability of this through what typedoc outputs?

website/docs/docusaurus.config.js
jbolda: Oh, yes, I should have been more clear about that. The current version does not support multiple instances of this plugin. Eventually I think we expect to use typedoc for these other ones, but we need to upstream support first. I had set them up so I left them in there as a reminder.

jbolda:
I was wondering about that as well. I imagine that there are configuration options within typedoc to let us modify public vs private APIs and how this is output. I didn't dig into it too much though as I thought it made sense as a separate PR. Get it working first and then we work on tweaking and modifying it.

jbolda:
Looks like we can use `@internal` to help clean up some of this (but subsequent PR?).
```ts
/** @internal */
export const secret = 1122;
```
https://typedoc.org/guides/doccomments/#internal

<blockquote><div>TypeDoc Documentation Generator</div><div><strong><a href="https://typedoc.org/guides/doccomments/">Doc Comments</a></strong></div><div>A documentation generator for TypeScript projects.
</div></blockquote>

jorgelainfiesta:
Ok sure, let's clean it up in a later PR. I'd also like to separate the guides from the API, but that may depend on what we end up leaving in the output.

website/docs/docusaurus.config.js
jorgelainfiesta: Ok got it!

jorgelainfiesta (approved):

PR #717, opened 2020-12-08
Remove last of the delegates

jnicklas:
The delegate pattern is no more. Now all service state is centralized in the atom. This is just a first step towards a few additional improvements I'd like to make around startup synchronization.

I've also tried to make the various service states a bit more consistent. Generally we use `starting` -> `started`, for server like things, but `pending` -> `ready` for static things.

### DIFF
--- a/.changeset/new-queens-check.md
+++ b/.changeset/new-queens-check.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/server": patch
+---
+
+Remove delegate mailboxes in orchestrator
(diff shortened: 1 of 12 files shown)


jnicklas:
Once #712 lands, I have an idea of how to improve the service status situation.

dagda1:
@jnicklas the last domino to fall!  This is a righteous PR :).

dagda1 (approved):
Great to see these gone.

PR #720, opened 2020-12-08
chore: gitignore vscode

jbolda:
This allows someone to set specific settings for BigTest within vscode such as formatting and running eslint.

Example `.vscode/settings.json`:
```settings.json
{
    "[typescript]": {
        "editor.formatOnSave": false,
        "editor.defaultFormatter": "dbaeumer.vscode-eslint",
        "editor.codeActionsOnSave": {
            "source.fixAll.eslint": true
        }
    }
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ node_modules
 yarn-error.log
 /packages/*/dist
 /packages/*/docs
+
+# personal editor settings
+.vscode


jnicklas:
You can set this in `~/.gitignore` and it will apply to all projects, not just this one :)

jbolda:
True, that is fair. There are some projects that actually commit the `.vscode`, but I haven't seen that much recently, at least in the projects I have been involved with. We can just rely on that here though; closing.

PR #723, opened 2020-12-09
Change to baseUrl /bigtest

jorgelainfiesta:
## Motivation 

We want to deploy bigtest's website to frontside.com/bigtest

## Approach

On frontside.com we'll setup a proxy on `/bigtest` that will read from `bigtestjs.io/bigtest`. Docusaurus provides a very handy baseURL setting that basically takes care of everything, and even serves it on development under `/bigtest`. 

In order to deploy netlify to a subdirectory I did some dirty commands. Please review if this is ok, or you know a better way of doing it. If we don't deploy it to a subdirectory, the proxy from frontside.com/bigtest would look ok but preview deployments would look broken.

We will need to set up a redirect from `bigtestjs.io/` to `frontside.com/bigtest/` when we're in production. But I think that might be better at the dns level. 

Additionally, this PR moves docusaurus to the root of `website` because everything will be integrated into docusaurus (docs, landing pages and API).


.netlify/state.json
```diff
@@ -0,0 +1,3 @@
+{
+	"siteId": "7477dfa0-78a2-4749-92a1-9c1eef30ad89"
```
jorgelainfiesta, line 2: I started using `netlify dev` and linked this repo to the netlify platform and that caused this file to be added. Do we keep it or ignore it @jbolda >

cowboyd:
If we set up bigtestjs.io to redirect to frontside.com/bigtest will be be in an infinite loop when we try and load content from bigtestjs.io/bigtest from frontside.com/bigtest ?

jorgelainfiesta:
I just mean the index, if somebody accesses `bigtestjs.io/` they should be redirected to `frontside.com/bigtest`.
Another alternative is to proxy directly from a netlify deployment url and use bigtest.io only for redirects.

cowboyd:
If we have 3 sites involved, is it going to be a performance hit to first request frontside.com/bigtest then bigtestjs.io/bigtest and then netlify?

jorgelainfiesta:
I don't know. But Netlify's guides does suggest hosting the site that you will proxy on netlify's own subdomain. This PR is not related to that though.

jbolda:
It sounds to me like we would like this site hosted at something like `bigtest.netlify.app` only. At the DNS level, we can redirect `bigtestjs.io` to `frontside.com/bigtest`. Then on `frontside.com` we can proxy to `bigtest.netlify.app`.

Configuration wise, `bigtestjs.io` would not be involved in netlify at all. The `netlify.toml` for this repo would not have any redirects / proxies, and the `netlify.toml` in `frontside.com` would proxy to `bigtest.netlify.app`.

cowboyd (approved):

PR #726, opened 2020-12-11
proposal: shift cypress out of main workspace and test separately

jbolda:
This is a test and proposal to see if we can get cypress out of our workspace dep tree. The installs become real slow and fail with it currently included. I also have a suspicion it will keep our lockfile a little more stable as well.

pros: cypress can cache separately in CI from other deps
cons: complicates the `prepack` step slightly and the publish step a bit more

It seems the benefits outweigh the negatives at face value.


packages/interactor/src/integrations/cypress.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-namespace */
-/// <reference types="cypress" />
+/// <reference types="../../integrations/cypress-interactors/node_modules/cypress" />
```
jbolda, line 2: I looked at the resulting package that was published, and it still has this set and obviously doesn't resolve. We may need to look into a way to handle this or can we remove it?

Do you happen to know @minkimcello?

minkimcello: I still haven't had a chance to pull the branch to see what's happening. I'll do that soon. I like to generate a preview package and consume it in a separate repository to confirm that everything is working as intended (just to be extra sure).

But from the short talk today, it seems Charles doesn't have any complaints against installing Cypress. Still, having it in the sample app AND in one of the workspace packages means we'll be installing Cypress twice. But from what I can see, there doesn't seem to be a significant difference in time between the runs in this PR and the other PRs.

minkimcello: my tsconfig knowledge is very limited but i don't see any errors showing with either of the reference types (i can use either or and restarting the typescript server each time and i don't see any red lines anywhere) 🤷

minkimcello:
so i pulled the branch to see what was being proposed.

1. as i mentioned in a reply to your review above, taking cypress out from the workspace dependency doesn't seem to help with the CI test run times. even for generating the preview packages, i see other PRs taking about 3-4 minutes to run as it did for this PR. so i'm not entirely sure how much time we're saving here.

2. for DX, i'm not sure if i'm a fan of bringing the interactor-specific test out to the root. if we had several other packages that needs to run cypress, maybe it would make sense to bring it out then. but for the sake of keeping things tidy and organized, i'd rather keep it all in the interactors package. but that's just my own personal preference. 🤷

jbolda:
From what I can tell, the cache is more stable and the installs take about half of the time. Note that this CI run was run without the cache, so depending on what CI runs you are comparing with, the comparable times might be with a cache? To put another way, I am seeing a run without the cache take the same time as a run with the cache without this change. I think some of that is a need to rebuild the lockfile, but comparing to [this run](https://github.com/thefrontside/bigtest/runs/1579402833?check_suite_focus=true), it hits the cache and takes longer (see lockfile issue partly). This might be [a better comparison](https://github.com/jbolda/bigtest/runs/1579844396?check_suite_focus=true#step:3:7) as it is a run without a cache hit. It seems it makes a much bigger difference on windows (maybe how cypress does the cache on windows?), as the [only linux run](https://github.com/thefrontside/bigtest/runs/1565222771?check_suite_focus=true#step:4:7) that I could find without a cache hit seems similar.

The DX does suffer a bit, but I think it is mostly alleviated by the setup scripts. This also partly serves as a model for future integrations as more dependencies will increase CI times, but also cause issues with clobbering deps / transitive deps (as seen recently in our tsconfig dep).

> But from the short talk today, it seems Charles doesn't have any complaints against installing Cypress. Still, having it in the sample app AND in one of the workspace packages means we'll be installing Cypress twice. But from what I can see, there doesn't seem to be a significant difference in time between the runs in this PR and the other PRs.

My understanding is that it isn't an issue because Cypress is still outside of the main workspace in your sample app setup. This seeks to achieve the same thing. Maybe I am misunderstanding the direction on the sample app? (Or did the direction change and I missed it?)

jbolda:
Seems we have decided to fully separate things out into different packages. Closing in favor of those implementations.

PR #728, opened 2020-12-12
readd @babel/preset-typescript to bundler

dagda1:
This PR is to basically back out the changes I made in #689 in which I removed the `@babel/preset-typescript`.

This caused problems in projects where they have monorepo packages like `@org/interactors` which has no `tsconfig.json` and therefore no compiled output.

Files are directly imported as `.ts` files.

When rollup encounters a reference like 

`import { Logo, Table } from '@org/interactors'`;

The `rollup-plugin-typescript2` does not find a match because by default, it will look in the current package for `**/*.ts+(|x)` and not outside so it was not transpiling the import.

The file was getting passed straight to babel which was barfing because the file contained type annotations.

I am of the opinion that you either use `@babel/preset-typescript` or `rollup-plugin-typescript2` and both seems overkill.  You could argue type checking is the job of CI and the code editor and not the job of bigtest but I've readded the babel preset.

### DIFF
--- a/.changeset/itchy-ladybugs-compete.md
+++ b/.changeset/itchy-ladybugs-compete.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/server": minor
+---
+
+readd @babel/preset-typescript to @bigtest/bundler
--- a/packages/bundler/package.json
+++ b/packages/bundler/package.json
@@ -20,6 +20,7 @@
     "@babel/core": "^7.12.9",
     "@babel/plugin-transform-runtime": "^7.12.1",
     "@babel/preset-env": "^7.12.7",
+    "@babel/preset-typescript": "^7.12.7",
     "@babel/runtime": "^7.12.5",
     "@bigtest/effection": "^0.6.0",
     "@bigtest/project": "^0.13.0",
--- a/packages/bundler/src/bundler.ts
+++ b/packages/bundler/src/bundler.ts
@@ -31,8 +31,9 @@ function prepareInputOptions(bundle: BundleOptions, channel: Channel<BundlerMess
     plugins: [
       resolve({
         mainFields: ["browser", "module", "main"],
-        extensions: [...RESOLVE_DEFAULTS.extensions, '.jsx'],
+        extensions: [...RESOLVE_DEFAULTS.extensions, '.ts', '.tsx', '.jsx'],
       }),
+      commonjs(),
       typescript({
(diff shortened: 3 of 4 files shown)


cowboyd:
> I am of the opinion that you either use @babel/preset-typescript or rollup-plugin-typescript2 and both seems overkill. You could argue type checking is the job of CI and the code editor and not the job of bigtest but I've readded the babel preset.

Do you think that this is a bad idea? Is there a long-term solution that involves us having one or the other but not both?

dagda1:
> > I am of the opinion that you either use @babel/preset-typescript or rollup-plugin-typescript2 and both seems overkill. You could argue type checking is the job of CI and the code editor and not the job of bigtest but I've readded the babel preset.
> 
> Do you think that this is a bad idea? Is there a long-term solution that involves us having one or the other but not both?

@cowboyd I'm not sure if it will cause problems but they are both providing typescript integration in different ways with the latter really just stripping out the types and giving an easier path.

Having both is a bit like having rollup and webpack doing the same build.

It might be OK to have both and I definitely did not consider the situation that caused the bug with typescript files directly imported without being compiled first.  Typescript project references might be a better idea than that and something I was going to suggest for bigtest.  

The babel preset might be the safer option long term but it all depends on what are requirements are for typing.

cowboyd (approved):

cowboyd:
Is it something where if we provided a build and a proper entry point (.js) that it would also fix the problem?

dagda1:
> Is it something where if we provided a build and a proper entry point (.js) that it would also fix the problem?

The problem with the `rollup-plugin-typescript2` is that it has the default rollup plugin `include` setting of `**/*.ts?x` so it does not look outside of the `cwd`.

Rollup plugins work by calling a `filter` function on any file in the pipeline and they use the `include` setting as part of the filter function.  If `fitler` returns false, it is passed onto the next plugin which is babel in our case.

So with the bug in question that caused this problem, files that are directly imported from another package are not transpiled because the `filter` function will return false unless we change the `include` setting to deal paths outside of the `cwd`

We could change the `include` to be more permissive.

People have weird and wonderful typescript configurations and I include myself in that and I wonder if life might be easier with the babel preset.

We could though change the `include` of the typescript2 plugin and see how we get on though.

dagda1:
or if you can change the project to have compiled output, it goes away also.  you could use typescript project references and i think we could try this for bigtest also.

PR #731, opened 2020-12-14
Better interactors hero

jorgelainfiesta:
## Motivation

The Interactors landing page was too abstract. I want to showcase Interactors already on the landinge page. 

## Approach 

I added a tabs component with examples of Interactors. I made up the examples so they look kind of ok, but I'm not sure if they're all good. 

Screenshot 2020-12-14 at 17 28 23


website/src/pages/interactors.js
```diff
@@ -111,8 +114,50 @@ function Interactors() {
...
+                {label: 'Jest', value: 'jest'},
+                {label: 'Cypress', value: 'cypress'},
+                {label: 'BigTest (alpha)', value: 'bigtest'}
+              ]}>
+              <TabItem value="jest">
+                <CodeBlock className='language-js'>
+                  {`it('subscribes to newsletter', async () => {
+  await Input('email').fillIn('jorge@frontside.com');
+  await Button('Subscribe').click();
+  await Heading('Thanks!').isVisible();
```
cowboyd:
```suggestion
  await Heading('Thanks!').exists();
```

website/src/pages/interactors.js
```diff
@@ -111,8 +114,50 @@ function Interactors() {
...
+              </TabItem>
+              <TabItem value="cypress">
+                <CodeBlock className="language-js">
+                  {`it('subscribes to newsletter', () => {
+  cy.do([
+    Input('email').fillIn('jorge@frontside.com'),
+    Button('Subscribe').click()
+  ]);
+  cy.expect([
+    Heading('Thanks!').isVisible();
```
cowboyd:
```suggestion
    Heading('Thanks!').exists();
```

website/src/pages/interactors.js
```diff
@@ -111,8 +114,50 @@ function Interactors() {
...
+              </TabItem>
+              <TabItem value="bigtest">
+                <CodeBlock className="language-js">
+                  {`test('subscribes to newsletter')
+  .step([
+    Input('email').fillIn('jorge@frontside.com'),
+    Button('Subscribe').click()
+  ])
+  .assertion(
+    Heading('Thanks!').isVisible()
```
cowboyd:
```suggestion
    Heading('Thanks!').exists()
```

cowboyd (commented):
I've suggested changed the assertions to one of the ways that they would actually appear. There are two options really:

```js
Heading("Thanks!").exists()
```

or 

```js
Heading("Thanks!").is({ visible: true })
```

Both of which would accomplish the same thing, but I opted for `exists()` since it's simpler.

cowboyd:
@jorgelainfiesta Also, worth adding an empty line between the actions and assertions in jest?

cowboyd (approved):
🎉

PR #734, opened 2020-12-15
Updates about copy

jorgelainfiesta:
## Motivation

Charles distilled the project's values and objectives, and now we're taking care of structuring the copy in a more snappy way.

## Approach 

I edited the about page with Olson as customary for the Frontside blog. I think he did a great job! 

[link to page preview](https://deploy-preview-734--bigtest.netlify.app/bigtest/about)

### DIFF
--- a/website/src/pages/about.mdx
+++ b/website/src/pages/about.mdx
@@ -1,124 +1,40 @@
 # BigTest
 
-The goal of the BigTest Project is to miniaturize the testing feedback
-loop while at the same time expanding the sizes of the systems it is
-possible to test.
+The goal of the BigTest Project is to minimize the testing feedback loop while at the same time expanding the size of the system it can test.
 
-# Fast
+## Fast and Smart
 
-The length of the testing feedback loop is equivalent to the time it
-takes from when you start a test until it either succeeds, or it fails
-and you are accurately able to understand why it failed. Why do we
-care about this metric? The math is simple. A project that can
-validate 100 changes in a 24 hour period has the potential of 10 times
-the throughput as a project that can only validate 10. This means that
-in order to achieve this, tests must be able to run quickly. But while
-test speed is a critical ingredient to a tight feedback loop, it is
-not the only thing. In the event of failure, the speed of diagnosis
-must be commensurately fast. Imagine a test that takes 5 milliseconds
-to run, but fails with an error message so inscrutable that it takes a
-further 90 minutes to figure out why. Unfortunately, despite a
-lightning fast test run, the effective length of the feedback loop is
-90 minutes. Thus in order to make testing feedback loops as small as
-possible, BigTest components must both run quickly, and provide
-failures that are precise and helpful. Apart from this, the only other
-substantial factor in the value of a test is its size.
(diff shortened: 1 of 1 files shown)


website/src/pages/about.mdx
```diff
@@ -1,124 +1,40 @@
...
-wait times to observable states, interactors naturally use the minimum
-amount of synchronization time while performing actions and
-assertions. Finally, because interactors are strongly associated with
-both the type and properties of the UI elements they actuate
-(e.g. Button, Checkbox, Input, ...), they can provide the next level
-in error messaging when something goes wrong, or an assertion fails.
+The length of the testing feedback loop is not determined by just how fast the test can report it succeeds or fails. Of course a test suite that can validate 800 assertions per minute has twice the throughput as a project that can only validate 400. But the critically important aspect of the speed of the testing feedback loop is how quickly and accurately you are able to understand why a test failed. A test that takes only 5 milliseconds to run but fails with an error message so inscrutable that it takes a further 90 minutes to figure out why is in fact excruciatingly slow. In order to make testing feedback loops as short as possible, BigTest components run quickly and provide failures that are precise and helpful.
```
cowboyd:
```suggestion
The length of the testing feedback loop is not determined by just how fast the test can report success or failure. Of course a test suite that can validate 800 assertions per minute has twice the throughput as a project that can only validate 400. But the critically important aspect of the speed of the testing feedback loop is how quickly and accurately you are able to understand why a test failed. A test that takes only 5 milliseconds to run but fails with an error message so inscrutable that it takes a further 90 minutes to figure out why is in fact excruciatingly slow. In order to make testing feedback loops as short as possible, BigTest components run quickly and provide failures that are precise and helpful.
```

cowboyd (commented):
These are really awesome!

The only change that didn't quite enhance the vibe we're going for was this statement:

> A test that is both fast and precise is not enough; it has to measure something of value as well. That’s not to say a React component rendering its props correctly isn’t critical; it’s just that a truly useful test would also test the component used in a workflow that correctly transfers money from one user account to another. 

I just don't think it really captures the juxtaposition of a test that covers very little value vs one that does. Maybe useful is not the right word?

jorgelainfiesta:
@cowboyd  how about something like this?

> That's not to say a React component rendering its props correctly isn't useful; it's just that a truly valuable test would act as insurance for your business, for instance, checking the workflow of correctly transferring money from one user account to another.

cowboyd:
> @cowboyd how about something like this?
> 
> > That's not to say a React component rendering its props correctly isn't useful; it's just that a truly valuable test would act as insurance for your business, for instance, checking the workflow of correctly transferring money from one user account to another.

I like it, although is that two uses of "truly valuable?"

jorgelainfiesta:
True! I updated the PR with this section text

> A test that is both fast and precise is not enough; it has to measure something of value as well. That's not to say a React component rendering its props correctly isn't useful; it's just that great tests would act as insurance for your business, for instance, checking the workflow of correctly transferring money from one user account to another. At Google they’ve [come to the realization][5] that "unit test," "functional test" and "end-to-end test" are arbitrary designations and what really matters is the relative size of a test. The larger a test, the more systems it integrates and thus the more business value it protects. The only hitch of course is that test size is in direct conflict with feedback loop speed.

> The BigTest project approaches every piece of its platform from the perspective of empowering development teams to master this tension and optimize testing feedback loops within the context of truly valuable tests.

cowboyd (approved):
Chef's kiss 👨🏻‍🍳 👌🏻 💋

PR #735, opened 2020-12-15
Fluid Interactors

jnicklas:
This explores the design space of an alternative API for creating interactors.

See: https://github.com/thefrontside/bigtest/discussions/737

### DIFF
--- a/packages/interactor/src/builder.ts
+++ b/packages/interactor/src/builder.ts
@@ -1,9 +1,19 @@
-import { Filters, Actions, InteractorBuilder } from './specification';
+import { Filters, Actions, LocatorFn, InteractorBuilder, InteractorSpecification, InteractorConstructor } from './specification';
+import { createConstructor } from './constructor';
 
-export function makeBuilder<T, E extends Element, F extends Filters<E>, A extends Actions<E>>(base: T): T & InteractorBuilder<E, F, A> {
+export function makeBuilder<T, E extends Element, F extends Filters<E>, A extends Actions<E>>(base: T, name: string, specification: InteractorSpecification<E, F, A>): T & InteractorBuilder<E, F, A> {
   return Object.assign(base, {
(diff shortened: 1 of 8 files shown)


packages/interactor/src/builder.ts
```diff
@@ -0,0 +1,19 @@
...
+    locator: (value: LocatorFn<E>): InteractorConstructor<E, F, A> => {
+      return createConstructor(name, { ...specification, locator: value });
+    },
+    filters: <FR extends Filters<E>>(filters: FR): InteractorConstructor<E, F & FR, A> => {
+      return createConstructor(name, { ...specification, filters: { ...specification.filters, ...filters } });
+    },
+    actions: <AR extends Actions<E>>(actions: AR): InteractorConstructor<E, F, A & AR> => {
+      return createConstructor(name, { ...specification, actions: Object.assign({}, specification.actions, actions) });
+    },
+  });
```
jnicklas, line 18: Once we remove backward compatibility, this will no longer need to be a separate thing

cowboyd (commented):
We should talk about what in the guides need to be updated with this new syntax. Obviously there are the 1:1 API changes, but this also changes the story around how to write your own interactor since this now enables deriving interactors from other interactors and the ability to to "extend" an existing interactor instead.

Let's update the 1:1 changes in the guides with this PR, and then we can have the discussion about what is the best way to write and organize your own interactors and your own interactor library. 

/cc @minkimcello @jenweber

jnicklas:
@cowboyd worth mentioning that as far as I am aware, this change is 100% backward compatible.

cowboyd:
Oh snap! Well in that case we can do the documentation upgrade later

cowboyd (approved):

jenweber:
Decoupling docs updates from this change makes sense to me. The new API experience seems great too.

PR #736, opened 2020-12-16
remove setMaxListeners from Atom and add configuration

dagda1:
I've removed `setMaxListeners` from the `Atom` interface and replaced it with an Atom configuration object that defaults to the following when calling `createAtom`:

```ts
export function createAtom<S>(init: S, { channelMaxListeners = DefaultChannelMaxListeners }: AtomConfig = {}): Atom<S> {
```

This is default object restructuring and the config object is optional for `createAtom`

The default `maxListeners` is currently `100000`.

### DIFF
--- a/.changeset/bright-points-fail.md
+++ b/.changeset/bright-points-fail.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/server": minor
+---
+
+add atom configuration and remove setMaxListeners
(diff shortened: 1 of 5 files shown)


cowboyd (approved):

cowboyd:
Getting better all the time!

PR #738, opened 2020-12-16
make reset private and remove from Atom interface

dagda1:
As discussed, remove `reset` from the `Atom` interface and access is now from a non-public `_reset`.

E.g. in `helpers` in `@bigtest/server` which is the only usage:

```ts
  (actions.atom as any)._reset((initial: OrchestratorState) => ({ ...initial, manifest, bundler }));
  ```

### CODE REVIEW

File: packages/server/test/helpers.ts [hunk 1]
```diff
@@ -124,7 +124,8 @@ after(async function() {
 });
 
 beforeEach(() => {
-  actions.atom.reset(initial => ({ ...initial, manifest, bundler }));
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (actions.atom as any)._reset((initial: OrchestratorState) => ({ ...initial, manifest, bundler }));
```
cowboyd: Can we perhaps extract this into a test helper that comes with the atom package? Like `resetAtom()` and then invoke it from here?

dagda1: @cowboyd I like that idea a lot.  On its way

cowboyd (approved):
nice! Optional: should we add a comment to `resetAtom` that this should be used for testing purposes only?

dagda1:
> nice! Optional: should we add a comment to `resetAtom` that this should be used for testing purposes only?

I've added the comment and removed the export from `src/index.ts` so you have to go to the `dist` directory to get it if you want it.

```ts
import { resetAtom } from '@bigtest/atom/dist/atom';
```

PR #739, opened 2020-12-18
dedupe the atom and slice functions

dagda1:
Previously to this PR, there was a duplication of getters, setters, update and other functions across the main `Atom` member and the `Slice<A>` instances that are created from the atom.  Duplicate subscriptions also existed.

This PR makes the top-level `atom` a slice that just happens to be at the top level.

```ts
  return {
    ...sliceMaker()() as Atom<S>,
    _channelMaxListeners: (states as any).bus._maxListeners,
    _reset: reset
  } as Atom<S>;
  ```

There are `getState` and `setState` functions that remain at the top level to mutate the `state` variable and retrieve it outside of the `sliceMaker` closure.

### CODE REVIEW

**Review by cowboyd — COMMENTED** [review]
Why do we need separate types now for `Atom`, `Slice`, and `Sliceable`. Can we not just get away with `Slice`?

**Comment by dagda1** [comment]
> Why do we need separate types now for `Atom`, `Slice`, and `Sliceable`. Can we not just get away with `Slice`?

Good point, we don't need `Atom`anymore, it is gone.

Slicable is the types for the overloaded slice function and it needs to stay.  This means we can keep the horrible overloads out of the code file.

If we upgraded to typescript 4.1, I am fairly sure I could get rid of the overloads

**Comment by cowboyd** [comment]
One final nit, if `Sliceable` is the type of the slice function, shouldn't it be `MakeSlice` or `Slicer`? In other words "verby" instead of "nouny"

**Comment by dagda1** [comment]
> One final nit, if `Sliceable` is the type of the slice function, shouldn't it be `MakeSlice` or `Slicer`? In other words "verby" instead of "nouny"

`MakeSlice` is definitely better.  `Sliceable` sounds more like an interface type.  Renamed

File: .changeset/four-colts-push.md [hunk 1]
```diff
@@ -0,0 +1,7 @@
+---
+"@bi
+"@bigtest/webdriver": minor
+"@bigtest/server": minor
```
cowboyd: webdriver and server should get a patch release automatically because they depend directly on atom. Then hitting patch on `bigtest` so that we can pull it in with a single bump.

```suggestion
"@bigtest/atom": minor
"bigtest": patch
```

cowboyd (approved):
Same functionality, a fraction of the code. 😎

amazing

Issue #740, opened 2020-12-18
Offset heading links


minkimcello:
When I click on a topic, the heading is covered by the navbar:

![Screen Shot 2020-12-18 at 2 45 03 PM](https://user-images.githubusercontent.com/29791650/102655193-e1aeb380-413f-11eb-898d-b85375975cf3.png)

I don't know if there's an offset option but it would be ideal to have it look like this so that the heading is in the view:

![Screen Shot 2020-12-18 at 2 45 15 PM](https://user-images.githubusercontent.com/29791650/102655228-f1c69300-413f-11eb-9336-667cb11efe57.png)

PR #746, opened 2020-12-21
Fix anchor scroll

jorgelainfiesta:
## Motivation

#740 

## Approach

This was a tricky one until I realized that the anchor was a hidden element within the heading.


minkimcello (approved):
👍

PR #747, opened 2020-12-21
Removes typedoc generated API from docusaurus

jorgelainfiesta:
## Motivation

The generated API ported to markdown wasn't great and was confusing so we remove them.

## Approach

I removed the dependencies we installed on docusaurus to integrate typedoc, as well as the settings for the integration plugin.


jbolda (approved):
It looks like you covered everything that I know of / added.

minkimcello (approved):

PR #748, opened 2020-12-21
Start discussion CTA

jorgelainfiesta:
## Motivation

As Jen pointed out, `Edit this page` isn't very useful so we want to replace it by a call to action that can be more helpful. The decision was to add a CTA at the end of the page prompting the reader to start a discussion if they found something confusing or didn't work for them. 

## Approach

1. I disabled the Edit this page by setting the edit url to an empty string.
2. "Swizzled" the `DocItem` component and added the little piece that we needed.

### DIFF
--- a/website/src/theme/DocItem/index.js
+++ b/website/src/theme/DocItem/index.js
@@ -1,0 +1,203 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+/**
+ * WARNING:
+ * DocItem is an internal component, and have a higher breaking change probability.
+ * Keep an eye on this file, specially when updating.
+ * The only customization is between lines 110-120 and 34-45 for the styles.
+ */
+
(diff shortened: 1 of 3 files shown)


website/src/theme/DocItem/index.js
```diff
@@ -0,0 +1,203 @@
+/**
```
cowboyd, line 1: Does Docosaurus support using the `.jsx` extension for JSX files?

jorgelainfiesta: Docusaurus always uses `.js` extensions, don't know why. If we want to use `.jsx` we'd have to update the other files too so I'd suggest we do it on a separate PR.

cowboyd (approved):

website/src/theme/DocItem/index.js
cowboyd: We should do this though since it 

1) plays nicer with editors
2) these files are technically not JavaScript, they are JSX


FB plays fast and loose with the truth when it comes to content type. They also called `Flow` files `.js` which was another case of something that is not true 🤷🏻‍♂️

PR #749, opened 2020-12-23
Docs - Polish the quick start and write your own sections

jenweber:
My goal was to simplify some of the explanations and tie up some loose ends. I think the quickstart is greatly improved by the airline example!

One thing I notice as I'm reading, we are a bit repetitive with regards to the "powers of interactors," but I think the official editorial pass will help us smooth that out.

One thing that is missing is an example of accessibility testing, but that could be added in a later version of the docs.


minkimcello (approved):
🙏

PR #750, opened 2020-12-23
Implement Project references

dagda1:
All of the packages which have a compiled output now use [typescript project references](https://www.typescriptlang.org/docs/handbook/project-references.html).

This cannot replace `prepack` yet because some of the  package`prepack` scripts do more than just build the typescript.

e.g. agent

```json
"prepack": "yarn prepack:app && yarn prepack:tsc",
```

The relevant `tsconfig.json` files now extend a root [tsconfig-base.json](./tsconfig-base.json) and have a `references` array of any dependencies, for example, the bigtest `tsconfig.json` now looks like this:

```json
{
  "extends": "../../tsconfig-base.json",
  // compiler options
  "references": [
    { "path": "../cli/tsconfig.build.json" },
    { "path": "../interactor" },
    { "path": "../suite" }
  ]
}
```

Each path in the array references a `path` to a directory with a `tsconfig.json` file or a config file if it is named differently as is the case with `cli` and the bundler packages that are named differently to stop the build config bleeding out into regular functionality.

There is a root [tsconfig.monorepo.json](./tsconfig.monorepo.json) that references all the projects:

```json
{
  "files": [],
  "references": [
    {
      "path": "./packages/eslint-plugin"
    },
    {
      "path": "./packages/atom"
    },
    // etc
```

The following npm/yarn scripts can build and watch the whole monorepo of packages for typescript changes:

```json
    "clean:tsc": "find . -path -o -name '*.tsbuildinfo' -delete",
    "prepack:tsc": "tsc --build ./tsconfig.monorepo.json",
    "watch": "yarn prepack:tsc && yarn tsc -b ./tsconfig.monorepo.json --watch"
```

Project references use incremental builds and typescript stores metadata in `*.tsbuildinfo` files.  

There is a `yarn clean:tsc` in the root to clean all *.tsbuildinfo files and each package now has a `yarn clean` script to ensure that a fresh build can be performed.

## TODOS

- [ ] verify impact on build pipeline perfomance. We currently store the result of a prepack so that it can be used once everywhere and with a separately uploaded artifact. Can we get away with using simple caching now? Major changes to the build can be deferred, but what we can't afford is blowing up the performance of all our builds.
- [ ] check the necessity of a pre-ordained build order now that building happens via dependency order because of project references


cowboyd (commented):
As I think i mentioned before the break, I'm very excited to get this in as our build is one of the most cumbersome and counter-intuitive bits of the project

I added some todos to the readme which may already be answered, but I want to make sure that we do our due diligence before merging in.

README.md
```diff
@@ -124,6 +124,16 @@ ultimately appear on NPM.
 2. Run `yarn prepack` to build all packages
 3. Run `yarn lint` to run linters
 
+### Typescript Project References
```
cowboyd: Let's get rid of this sub-heading and move the content of the section up into the "Building" section above since this is a natural extension of building.... If we do want to separate it into two parts, then maybe a "building during development"

However we do it, I think that these sections should be adjacent.

dagda1:
> I added some todos to the readme which may already be answered, but I want to make sure that we do our due diligence before merging in.

@cowboyd Due diligence is always very, very welcome :).  I don't want to rush this through and it goes belly up.

README.md
```diff
@@ -100,17 +100,28 @@ yarn bigtest test
...
 ```
 
-This will compile each package into the state in which it will
-ultimately appear on NPM.
+All of the packages which have a compiled output use [typescript project references](https://www.typescriptlang.org/docs/handbook/project-references.html) for faster build times and a better project structure.
+
+Any individual package can be built with the `prepack` script, eg. for `@bigtest/server`
+
+```bash
+yarn --cwd ./packages/server prepack
```
cowboyd:
```suggestion
yarn workspace @bigtest/server prepack
```

cowboyd:
I tried this out today, and it was big time saver. Although how can we avoid the custom builds because I noticed that those are what really slow down the subsequent build times:

It's whenever we have a custom thing like:

```
> @bigtest/globals
$ tsc --build && tsc --outdir dist/esm --module es2015 && tsc --outdir dist/cjs --module commonjs
```

That things get slow. Is there any way that we can avoid this, or is this what you were talking about addressing and making more consistent in subsequent PRs?

dagda1:
> or is this what you were talking about addressing and making more consistent in subsequent PRs?

yes, that was what i was talking about.

we could go over it on a call when you have time, just give me a shout when you are next online.

cowboyd (approved):

PR #754, opened 2021-01-04
Admonition styles + tweaks

### LINKED ISSUE #751: Links are difficult to click in a quote
This might be a docusaurus issue, I'm not sure. Links are fine in regular text but if we have a link inside of a quote:

```md
> [hello](google.com)
```

... the clickable area is limited. And that area changes depending on the responsive size of the website. Below are screenshots with the clickable area highlighted in green and unclickable in red:

Screen Shot 2020-12-23 at 3 33 17 PM
Screen Shot 2020-12-23 at 3 33 29 PM

See example from the preview build: https://deploy-preview-662--bigtest.netlify.app/bigtest/docs/interactors/

jorgelainfiesta:
## Motivation 

Docusaurus provides [admonition/callout blocks](https://v2.docusaurus.io/docs/markdown-features/#calloutsadmonitions) within markdown. So far, we had used `>` to mark notes and such but I think we can use these blocks for better semantics/styles.

## Approach

- This PR replaces all instances of notes with `>` on the Interactors guides with a `:::note` block. Can y'all check if they make sense? specially the little titles that I added @minkimcello @jenweber 
- I only made styles for `::note` and `:::info`, we can create styles for `:::tip`, `:::caution` or `:::danger` when we need them. If that's already something we want to use, let me know and I'll make them now.
- Small tweaks: fixes #751, also the jiggly topnav reported on #683 (before there was a 3px change on the link width caused by bold fonts, now links have a fixed width to prevent that).

[Link to page that uses two types of admonitions](https://deploy-preview-754--bigtest.netlify.app/bigtest/docs/interactors/write-your-own)

### DIFF
--- a/website/docs/interactors/1-quick-start.md
+++ b/website/docs/interactors/1-quick-start.md
@@ -8,7 +8,9 @@ BigTest's Interactors make your UI tests easier to write, faster, and more relia
 
 By the end of this quick start, you will be testing one of your own apps with Interactors.
 
-> If you need help or have questions along the way, please let us know in [the Discord chat](https://discord.gg/r6AvtnU).
+:::note We're here to help
+ If you need help or have questions along the way, please let us know in [the Discord chat](https://discord.gg/r6AvtnU) or [open a discussion](https://github.com/thefrontside/bigtest/discussions/) on Github.
+:::
 
 ## Prerequisites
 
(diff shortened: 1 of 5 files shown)


minkimcello (approved):
👍👍

PR #760, opened 2021-01-06
Enable sameorigin iframe embedding

jorgelainfiesta:
## Motivation 

The iframe on /platform cannot be loaded due to Neltify's default `DENY` policy. 

## Approach

I set `SAMEORIGIN` as policy for the file that we want to embed, the other routes keep the default policy.


taras (approved):
👍

PR #761, opened 2021-01-06
Fixes mobile menu cropping

jorgelainfiesta:
## Motivation 

The nav is broken on mobile.

## Approach

There was a problematic CSS that I introduced in the past that we no longer needed and had affected docusaurus mobile styles.

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -138,7 +138,6 @@ body {
   max-width: var(--ifm-container-width);
   margin-left: auto;
   margin-right: auto;
-  padding-top: var(--ifm-navbar-height);
 }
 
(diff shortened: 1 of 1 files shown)


jorgelainfiesta:
@cowboyd I only made a CSS change, yet I'm getting a webdriver test error. Is that expected?

jbolda:
> @cowboyd I only made a CSS change, yet I'm getting a webdriver test error. Is that expected?

It isn't unexpected. We don't currently scope our tests to only run on code changes. We may want to do that so they aren't all running on website changes that don't gain value by having the tests run.

jbolda (approved):
Cheers!

PR #762, opened 2021-01-07
Upgrade to typescript 4.1

dagda1:
This PR

- upgrades typescript to 4.1
- upgrades eslint to the latest....more on this below
- makes sure each package is on the latest `@frontside/tsconfig`.  Previously, it was getting clobbered in `@bigtest/project` which had a star refrence `"@frontside/tsconfig": "*"`.  This meant it was resolving to `0.0.1`.
- removes all `*` referecnes for things like `ts-node`.

This turned into a bit of saga due to the fact that `react-scripts` was clobbering  the`@typescript-eslint/eslint-plugin` and others.

No matter what version we referenced in any `pacakge.json`, `@typescript-eslint/eslint-plugin` was resolving to `2.10.0`.

I've added a `resolution` for now because `@bigtest/todomvc` will be toast soon and then we can remove the resolution.

Now we are on the latest `@typescript-eslint`, a lot of new rules are coming through from `plugin:@typescript-eslint/recommended` that were not there in `2.10.0`.

The main big change is [explicit-module-boundary-types](https://github.com/typescript-eslint/typescript-eslint/blob/v4.12.0/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md) which means that any module boundary function must have an explicit return type.

I personally like this rule but it might be annoying so we can turn it off if others find it too annoying.

The PR is a monster and I apologise for that but I think it needs done and we should just bite the bullet.

There are codeformatting rules now also and if you get your text editor set up to run `eslint --fix` on each save then you get code formatting which I love but again might be annoying for others.


.changeset/serious-suits-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigtest/cli": minor
+---
```
cowboyd, line 3: Since this will change the compiled code for every package in the entire workspace, I think we'll need to bump each  by a patch.

packages/agent/app/agent.ts
```diff
@@ -1,8 +1,9 @@
-import * as Bowser from 'bowser';
+import Bowser from 'bowser';
 import { QueryParams } from './query-params';
 import { Agent } from '../shared/agent';
 import { run } from './runner';
 
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
cowboyd, line 6: I'd rather add the type annotations than the disable comment.  I think the return type here would be `Operation<void>`

packages/cli/src/start-server.ts
```diff
@@ -11,7 +11,7 @@ interface Options {
 
 // TODO: this is what the server package should be doing in the first place
 // See: https://github.com/thefrontside/bigtest/issues/295
-export function* startServer(project: ProjectOptions, options: Options) {
+export function* startServer(project: ProjectOptions, options: Options): Generator {
```
cowboyd, line 14: Return type here should be `Operation<object>` I think?

cowboyd (commented):
You're right, this is quite the novella :) and it's a bit overwhelming to be able to sit down and review it in one go. How about we adopt this strategy for getting it merged in more digestible chunks:

let's do one PR with the structural changes to the dependencies first. In that PR we can globally disable all the new linting rules that require fixing. Then, we can add the linting rules back in one at a time (or in groups where appropriate).

This should be a quick review afterwards since there's much fewer _kinds_ of changes to keep track of in your head.

dagda1:
> You're right, this is quite the novella :) and it's a bit overwhelming to be able to sit down and review it in one go. How about we adopt this strategy for getting it merged in more digestible chunks:
> 
> let's do one PR with the structural changes to the dependencies first. In that PR we can globally disable all the new linting rules that require fixing. Then, we can add the linting rules back in one at a time (or in groups where appropriate).
> 
> This should be a quick review afterwards since there's much fewer _kinds_ of changes to keep track of in your head.


I like that idea and I normally do formatting changes as a single change but I did not realise that `yarn lint --fix` was doing that until late in the game.

I don't think it is possible for a human being to review a PR this size :).

I'll close this and fire out some smaller PRs.

packages/agent/app/agent.ts
dagda1: @cowboyd  the big union from `Operation` does not have the means to type narrow to the specific case which makes the type quite tricky to deal with.  I took the cowards way out :).

For v2 of affection, I would recommend some sort of conditional type to type narrow into the correct type in terms of its context.

There is nothing wrong with big unions, discriminated or not but they must be able to type narrow or their use is limited:

I had a quick hack a while ago at an `Operation<T>` after there was talk of `Task<T>` and came up with this, I'm definitely not saying to use this, just to illustrate that the union is narrowed

```ts
export type Operation<T> = T extends (task: Task<infer U>) => Iterator<infer R>
  ? (task: Task<U>) => Iterator<Operation<U>, R>
  : T extends PromiseLike<infer U>
  ? PromiseLike<U>
  : T extends Iterator<infer U, infer R, infer N>
  ? Iterator<U, R, N>
  : never;
```

dagda1:
closed in favour of #762 and #764

dagda1:
closed in favour of #764 and #766

PR #763, opened 2021-01-07
Configures BigTest GA id

jorgelainfiesta:
## Motivation

We'd like to know about our visitors behavior on the documentation site.

## Approach

We currently use Google Analytics in Frontside's website, and it was also used the previous BigTest website. Docusaurus provides built-in support for GA so we just need to configure the ID.  I also updated GA admin to point to the new URL.


jbolda (approved):
👍

PR #764, opened 2021-01-08
Upgrade typescript and eslint only

dagda1:
I've scaled things right back from the #762 so this should be more manageable:

I've disabled a bunch of rules that are coming through from the **real** `@typescript-eslint/eslint-plugin` which is versioned at `4.12.0`.

```json
{
  "extends": "@frontside",
  "rules": {
    "@typescript-eslint/ban-types": "off",
    "@typescript-eslint/explicit-module-boundary-types": "off",
    "eol-last": "off",
    "@typescript-eslint/no-extra-semi": "off",
    "no-multi-spaces": "off",
    "no-trailing-spaces": "off",
    "object-curly-spacing": "off",
    "semi": "off"
  }
}
```

I really like having the formatting done through `lint --fix` and have my text editor set up to do that on save.  These rules should probably go in `@frontside/eslint-plugin` so we are consistent across all projects.

I also like prettier and eslint-prettier so you don't end up spending time thinking about minutae but that might not be for everyone.

I've also kept `"esModuleInterop": false` where it was previously false to keep the number of changes down.

I'll enable it in a future PR.


cowboyd (approved):
Looks great!

PR #765, opened 2021-01-10
Match HTML description of interactors to heading

taras:
## Motivation

image

Posted that in Discord and found `Test your app as a person uses it` awkward to read. Then found a better version on the actual page.

## Approach

Update HTML description so it matches the description on the website.


cowboyd (approved):

jorgelainfiesta:
I need to add better meta `og:` tags to docusarus, like the images etc

PR #767, opened 2021-01-11
skip tests on documentation only edits

jbolda:
This should "scope" test running to when there is a real value in running them, ie when the code changes. If the documentation or website is the only change, there isn't a real need to run the tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,10 +3,10 @@ name: Test
 on:
   pull_request:
     branches:
-      - "v[0-9]+"
+      - 'v[0-9]+'
     paths-ignore:
-      - "website/**"
-      - "**.md"
+      - 'website/**'
+      - '**.md'
 
 jobs:
   prepack:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -44,7 +47,21 @@ jobs:
...
+            logging,
+            bundler,
+            project,
+            server,
+            suite,
+            interactor,
+            todomvc,
+            atom,
+            webdriver,
+          ]
```
cowboyd: is this related?

.github/workflows/test.yml
```diff
@@ -92,4 +109,4 @@ jobs:
         with:
           working-directory: packages/interactor/
           start: yarn bigtest-todomvc 3000
-          wait-on: 'http://localhost:3000' 
\ No newline at end of file
+          wait-on: "http://localhost:3000"
```
cowboyd: seems like a formatting change.

.github/workflows/test.yml
```diff
@@ -3,7 +3,10 @@ name: Test
 on:
   pull_request:
     branches:
-      - 'v[0-9]+'
+      - "v[0-9]+"
+    paths-ignore:
+      - "website/**"
+      - "**.md"
```
cowboyd: Will this match the readmes in the individual packages? If so, we should avoid those since that will actually change the content of what gets published to npm.

cowboyd (changes requested):
YAASSSSSSSSSS!!!! 

want!

cowboyd (changes requested):
YASSSS! WANT!

.github/workflows/test.yml
jbolda: ah, blast, I reconfigured things (in my personal setup) and must have left auto-formatting of yml on. I'll revert this and the package list change.

.github/workflows/test.yml
jbolda: Yes it will match those. True that it will change what gets published, but running the tests for a readme change isn't at all connected to publishing, is it?

cowboyd: That would be weird wouldn't it!

cowboyd (approved):

PR #766, opened 2021-01-11
Add missing eslint rules

dagda1:
Enables the following from `@typescript-eslint/recommended`:

- `@typescript-eslint/ban-types` this mainly affects the duplicitous `{}` type that really is a top type for all non-null types.
- `@typescript-eslint/explicit-module-boundary-types`

Enabling these rules will make things a bit stricter and a bit better.

PR #768, opened 2021-01-11
Remove Interactor class, simplify interactor types.

jnicklas:
This makes the inner workings of interactors a bit more straight-forward, in preparation for #737. It removed the `Interactor` class and replaces it with an interface. To create an interactor, we will create a one-off object for each time the interactor is intstatiated. This is more inefficient from a memory point of view, but it means we can avoid prototypal inheritance entirely, and instead use more "normal" javascript.


cowboyd (approved):
lgtm

packages/interactor/src/specification.ts
```diff
@@ -1,9 +1,130 @@
...
+   * ## Example
+   *
+   * ``` typescript
+   * await CheckBox('Accept conditions').is({ checked: true })
+   * ```
+   */
+  is(filters: F): ReadonlyInteraction<void>;
+}
+
+export type ActionFn<E extends Element> = (interactor: Interactor<E, {}>, ...args: any[]) => Promise<unknown>;
```
dagda1: one of the eslint rules that is coming in from #766 is to ban the use of the duplicitous `{}`. 

As you know, `{}` is not an empty object but all non-null values.

You can, of course, disable this rule in the same way as we do for `any` but if possible, could we use a different type here?

packages/interactor/src/specification.ts
```diff
@@ -49,7 +170,7 @@ export type InteractorSpecification<E extends Element, F extends Filters<E>, A e
 }
 
 export type ActionMethods<E extends Element, A extends Actions<E>> = {
-  [P in keyof A]: A[P] extends ((interactor: InteractorInstance<E, {}, {}>, ...args: infer TArgs) => Promise<infer TReturn>)
+ [P in keyof A]: A[P] extends ((interactor: Interactor<E, {}>, ...args: infer TArgs) => Promise<infer TReturn>)
```
dagda1: Same as before, if possible, use a different type than `{}`

dagda1 (commented):
I'm hugely in favour of removing the inheritance and class so I like this a lot.

I've added a couple of small comments.

packages/interactor/src/specification.ts
jnicklas: Oh yeah, I keep forgetting about this! Maybe `never` would be an appropriate choice here, I'll try it out and see if it works!

jnicklas: Okay, so I tried this with both `never` and `Record<string, any>` and neither of these work, unfortunately. I don't have any other ideas for this, unfortunately.

`never` I think doesn't work because it breaks the mapped types, `Record<string, any>` is too permissive, because due to the type mapping it will actually allow `myInteractors.completelyArbitraryMethod()`, so basically you can call *any* method on the interactor. Do you have any other ideas for types to try?

dagda1: @jnicklas you could try `Record<string, never>` which really is a type of empty object

`export type ActionFn<E extends Element> = (interactor: Interactor<E, Record<string, never>>, ...args: any[]) => Promise<unknown>;`

jnicklas: Interesting! This one did work for the constraints in `ActionFn` and `ActionMethods`, but it does *not* work as the default type in InteractorBuilder. I think this is because this type gets composed into an intersection type like this:

``` typescript
type Result = Record<string, never> & { some: string };
```

Which would effectively expand to something like:


``` typescript
type Result = { some: string & never };
```

Which doesn't make sense.

I'm guessing that using this for the constraints only would be better.

Once again, the lack of a proper "empty object" type in TypeScript is pretty annoying.

dagda1 (approved):

packages/interactor/src/specification.ts
dagda1: @jnicklas what should the defaults be for `InteractorBuilder`, would they really be an empty object.  Would they not be `never`?

PR #775, opened 2021-01-15
Adds a general meta image to website

jorgelainfiesta:
## Motivation 

When sharing bigtest website's the preview is a bit boring because we don't have a meta image set up as we do for Frontisde.

## Approach

I initially wanted to have a meta image for the Interactors part and a different one for the Platform side, and perhaps a third one for other pages. However, this was not a simple feat. Docusaurus allows us to set a general image and that's it.

Yes, I could've used helmet on the react pages (landing pages) to set specific meta images. But, I couldn't find a way of doing it for the docs for they're just markdown.

You could argue I could've swizzled the docs component, read markdown's meta tags and apply a conditional images based on that. However I don't think this is important enough to justify that work and ejecting more components out of docusaurus. 

So I'm just keeping it simple for now by creating a general image.


cowboyd (approved):

jorgelainfiesta:
I can't get to have the checks passing. It seems that the work that @jbolda did to only run those tasks outside of `website` isn't working anymore?

cowboyd:
@jorgelainfiesta It may be that what @jbolda did _does_ work, but those checks are required for a merge to master. So since they don't run, the merge is impossible.

jbolda:
> @jorgelainfiesta It may be that what @jbolda did _does_ work, but those checks are required for a merge to master. So since they don't run, the merge is impossible.

Ah jeez, that has got to be the issue. Oof, this system could really use some love from Github. Should we remove my change for the moment? I don't know that there is a good solution for the required checks with their built-in options.

cowboyd:
I can't tell how we could configure it to make it work other than making the these checks pass if the change is docs only which feels... *ugh*

jbolda:
Looks like conditional checks has been on feature requests for like 1.5+ years. Looks like there are a few GitHub Actions that can do conditional checks, maybe worth looking at that route?

cowboyd:
@jorgelainfiesta Now that we're on circle, let's just rebase.

cowboyd:
@jorgelainfiesta fire when ready!

PR #776, opened 2021-01-15
Extract Cypress as a separate package

### LINKED ISSUE #759: Extract cypress integration into separate `@bigtest/cypress` package in separate workspace
Our goal with cypress integration was to make using interactors as easy as possible. That's why our original design involved required nothing more than just importing interactors from the `bigtest` package:

```js
import { TextField } from 'bigtest';

describe('login', () => {
  cy.do(TextField("username").fillIn("cowboyd"));
})
```

While this was easy, it turns out not to be that simple, as it introduces a number of complexities.

1. There is no way to specify which version of cypress this works on. We don't want to introduce a "hard" dependency from `bigtest` -> `cypress` and so we just detect at runtime if cypress is there and then extend it. While that is kinda cool, what happens if the cypress API changes? We have no way of knowing until runtime. What if we want to support more than one version? This puts us in the unenviable position of having to dynamically link every possible version of cypress in every possible version of bigtest.
2. In order to develop and test we need to have cypress in our devDependencies. This adds a huge overhead to the install weight of `node_modules` and makes CI cycles significantly longer.

A solution to both of these problems is to move our integration into a separate `@bigtest/cypress` package. This package can depend on a specific version range of bigtest interactors and a specific version range of cypress. That way, if we need to change the cypress integration, we can just release a new version of `@bigtest/cypress` but anybody who needs to use the old integration is still good to go.

If we put the package in a separate `integrations` workspace, then it can be built and cached separately.

- [ ] create and publish `@bigtest/cypress`
- [ ] update README and interactor guides

minkimcello:
## ⭐ Primary Changes ⭐
- Moved cypress from `package/interactors/integrations` to `integrations/cypress`
- Named package `@bigtest/cypress` and include `cypress` in the dependencies
- Added README for `integrations/` and `integrations/cypress/`
- Changeset not included in the integrations monorepo
  - Because changeset only runs at root, using changeset cli in a monorepo that's not included in the workspace config of root would mean it's only good for versioning; which we might as well have a developer bump the versions manually themselves at that point (this is explained in the readme)
    - There should be a better solution to this workflow (I'll create an issue for it ⚠️)
- The install/prepack scripts of the root of bigtest monorepo will install/prepack for the integrations monorepo as well but keep the dependency tree separate
- Updated docs like so:
Screen Shot 2021-01-20 at 10 17 43 PM
Screen Shot 2021-01-20 at 10 18 06 PM
  - Updated all the cypress-specific examples throughout the guide
  - I thought these heads up (in the photos above) were necessary because cypress integration isn't introduced until the end of the docs so someone might follow along the guides and get frustrated unless we point this out for them at the very beginning? 🤷
- I'll need to update the sample app to use `@bigtest/cypress` instead of `bigtest` but I'll do that in a separate PR
- todomvc needs to be replaced with the sample app too in the integrations/cypress test

## Motivation
To resolve #759

## Approach/TODOs
- [x] Move all the cypress-relevant files from `./packages/interactor` to `./integrations/cypress`
### Removing Cypress from Interactors
- [x] Remove cypress-integration-related dependencies and irrelevant scripts from `interactor`'s `package.json`
- [x] Revert tsconfig to normal for `package/interactor`
- [x] Add changeset to remove cypress from next version of interactors
### Creating Cypress Package
- [x] Create `package.json` for `@bigtest/cypress`
- [x] Manually publish `@bigtest/cypress@0.0.1` to get it in the registry
- [x] Create README
- [x] Update Docs
- [x] Import/export interactors from within `@bigtest/cypress`
- [x] Consume/test-out new bigtest/cypress package in a practice repo
  - ~Update sample app with the new package~ ⚠️ will do this in a separate PR
### Workspace/Changeset Related
- [x] ~~Configure two directories for changeset config?~~ no, see top in the Primary Changes section
- [x] Added integrations workspace


integrations/cypress/cypress/integration/interactor.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { TextField, Heading } from '../../../../src/';
+import { TextField, Heading } from '../../src/index';
```
dagda1: why would you need index?

integrations/cypress/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "src/**/*",
+    "README.md"
+  ],
+  "dependencies": {
+    "@bigtest/globals": "^0.7.3",
+    "@bigtest/interactor": "^0.23.0",
+    "cypress": "^5.6.0"
+  },
+  "scripts": {
+    "start": "cd ../../packages/todomvc/ && yarn && yarn prepack && yarn bigtest-todomvc 3000",
```
dagda1, line 22: is todomvc not redundant now that sample application got merged?

integrations/cypress/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from '@bigtest/interactor';
+import './cypress';
```
dagda1, line 2: what is the import doing here?

dagda1 (changes requested):
This looks good!  I've left a few comments.  Nothing major

integrations/cypress/cypress/integration/interactor.spec.ts
minkimcello: Oh I guess we don't. I'll remove it.

integrations/cypress/package.json
minkimcello: I'm going to replace todomvc with the sample app in a separate pr because the sample app itself needs to consume @bigtest/cypress too.

integrations/cypress/src/index.ts
minkimcello: Isn't it necessary to import the functions into index so that they're applied in the environment that people install @bigtest/cypress into?

dagda1: oh right yes, I forgot about all that `Cypress.Commands.add` business

Can we move this into a separate .d.ts file

```ts
declare global {
  namespace Cypress {
    interface Chainable<Subject> {
      do(interaction: Interaction<void> | Interaction<void>[]): Chainable<Subject>;
      expect(interaction: ReadonlyInteraction<void> | ReadonlyInteraction<void>[]): Chainable<Subject>;
    }
  }
}
```

Also, can we not use 

```json
    "types": [
      "cypress",
      "node"
    ]
  ```
rather than

```ts
/// <reference types="cypress" />
```

minkimcello: I can give it a try but I think we've run into this problem before and ended up going in circles (for both separating the global declaration and avoiding reference types).

dagda1: @minkimcello I can do it on a separate PR if it proves problematic.  Don't waste too much time on it

dagda1: I can't think of any reason why it should not work but maybe I'm missing something.

minkimcello: oh okay 👌 so i'll resolve this conversation

integrations/README.md
```diff
@@ -0,0 +1,10 @@
+This directory contains BigTest integration for Cypress. The reason why there is a separate monorepo is so that we can separate the dependencies of this directory from the BigTest core components in order to provide a better developer experience.
+
+## Development
+### Install/Build
+To develop packages inside the integrations directory, you can run `yarn install` and `yarn prepack` from the root of the BigTest repository and it will install/prepack both the integrations and the BigTest core components.
+
+Or you can `cd` into `integrations/` and run `yarn install` or `yarn prepack` to install and prepack just the packages inside `integrations/`.
+
+### Publishing
+One feature we were not able to include in this duo-monorepo setup is its versioning and publishing workflow. The changesets github action will only run from the root of the repository so the changesets of packages that are not specified in the workspaces at the root will never consumed. So if you're adding features to any of the integrations packages, you will need to bump the package version manually in your pull request.
```
dagda1:
```suggestion
One feature we were not able to include in this duo-monorepo setup is its versioning and publishing workflow. The changesets github action will only run from the root of the repository so the changesets of the packages that are not specified in the workspaces at the root will never be consumed. If you are adding features to any of the integrations packages, you will need to bump the package version manually in your pull request.
```

integrations/README.md
```diff
@@ -0,0 +1,10 @@
+This directory contains BigTest integration for Cypress. The reason why there is a separate monorepo is so that we can separate the dependencies of this directory from the BigTest core components in order to provide a better developer experience.
+
+## Development
+### Install/Build
+To develop packages inside the integrations directory, you can run `yarn install` and `yarn prepack` from the root of the BigTest repository and it will install/prepack both the integrations and the BigTest core components.
+
+Or you can `cd` into `integrations/` and run `yarn install` or `yarn prepack` to install and prepack just the packages inside `integrations/`.
```
dagda1: rather than have instructions, can we not supply an npm script to do this?

```
### Install/Build

yarn install-integrations
```

could we use `"yarn install:integrations"`

minkimcello: Oh I already have `yarn install:integrations` so I guess I could say something like "if you only want to install/build the packages inside integrations, run `yarn install:integrations` and `yarn prepack:integrations`"

dagda1: > Oh I already have `yarn install:integrations` so I guess I could say something like "if you only want to install/build the packages inside integrations, run `yarn install:integrations` and `yarn prepack:integrations`"

I would have 1 command only.  Where possible, I would try and avoid giving users a list of things to do.  

You could give them a script that combined those 2.

so...

```md
if you only want to install/build the packages inside integrations, run 

yarn install-and-build-integrations
```

then you have a script like

```sh
"scripts":  {
   "install-and-build-integrations": "yarn install:integrations && yarn prepack:integrations"
}
```

minkimcello: Would it be misleading to just call it the build script? Install installs, prepack prepacks, build will install _and_ prepack?
```
yarn build:integrations
```

dagda1: > Would it be misleading to just call it the build script? Install installs, prepack prepacks, build will install _and_ prepack?
> 
> ```
> yarn build:integrations
> ```

I like that.  Nice one!

dagda1 (approved):
good work!

PR #777, opened 2021-01-19
Add `extend` method to builder API

jnicklas:
This makes it possible to both rename the interactor, and crucially to use a subtype of the element. To make this work we need to "forget" the element type of those actions and filters which have already been defined, otherwise we run into thorny variance issues. The type parameter `E` of `InteractorBuilder` must be *covariant* for this to work.

### DIFF
--- a/.changeset/shy-frogs-carry.md
+++ b/.changeset/shy-frogs-carry.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/interactor": minor
+"bigtest": minor
+---
+
+Add `extend` method to interactor builder API, to downcast to a subtype of the element
--- a/packages/interactor/src/builder.ts
+++ b/packages/interactor/src/builder.ts
@@ -16,5 +16,8 @@ export function makeBuilder<T, E extends Element, FP extends FilterParams<any, a
     actions: <AR extends Actions<E>>(actions: AR): InteractorConstructor<E, FP, AM & ActionMethods<E, AR>> => {
       return createConstructor(name, { ...specification, actions: Object.assign({}, specification.actions, actions) });
     },
+    extend: <ER extends Element = E>(newName: string): InteractorConstructor<ER, FP, AM> => {
+      return createConstructor(newName, specification) as unknown as InteractorConstructor<ER, FP, AM>;
+    },
   });
 }
--- a/packages/interactor/src/builder.ts
+++ b/packages/interactor/src/builder.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Filters, Actions, LocatorFn, InteractorBuilder, InteractorSpecification, InteractorConstructor, FilterParams, ActionMethods } from './specification';
 import { createConstructor } from './constructor';
+import { MergeObjects } from './merge-objects';
 
 export function makeBuilder<T, E extends Element, FP extends FilterParams<any, any>, AM extends ActionMethods<any, any>>(base: T, name: string, specification: InteractorSpecification<E, any, any>): T & InteractorBuilder<E, FP, AM> {
   return Object.assign(base, {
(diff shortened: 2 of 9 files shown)


jnicklas:
One downside of this PR is that previously we always used `F` and `A` as the type parameters which are constrained to `Filters` and `Actions` respectively. This was easy to understand. Now we sometimes use `Filters` and sometimes  `FilterParams`, likewise `Actions` and `ActionMethods`.

I think some of this can be simplified if we break backwards compatibility, since we could probably simplify the type of InteractorSpecification in that case, and therefore remove the need for `Filters` and `Actions` almost entirely, but as is, we're a bit stuck with this.

cowboyd (approved):
This looks good to me, although I'm not a typescript taste-maker like @dagda1. From the public API standpoint, it looks like we won't need to often use `ActionMethods` and `FilterParams` ourselves, so the added complexity of the typing doesn't bother me so much.

cowboyd:
> I think some of this can be simplified if we break backwards compatibility, since we could probably simplify the type of InteractorSpecification in that case, and therefore remove the need for Filters and Actions almost entirely, but as is, we're a bit stuck with this.

How about we release one version for forwards compatibility so that we can upgrade our current install bases, and then we'll break backwards compatibility before 1.0

jnicklas:
@dagda1 what do you think about this?

dagda1:
> @dagda1 what do you think about this?

This looks interesting!  I'll take a look now.  

Do you have an example of how this affects the existing API or what problems this introduces?

> we sometimes use Filters and sometimes FilterParams, likewise Actions and ActionMethods.

this sounds like it should be a conditional type

dagda1:
> @dagda1 what do you think about this?

Could `ActionMethods` and `Actions<E>` become 1 conditional type?

Same for `FilterParams` and `Filters`?

jnicklas:
> Could ActionMethods and Actions<E> become 1 conditional type?

They are pretty different. Basically an `ActionMethods` takes a an `Actions` as a parameter and transforms it.

So:

``` typescript
let actionDefinitions: Actions<HTMLElement> = { click: (interactor) => interactor.perform((element) => element.click() }
let actionImplementation: ActionMethods<HTMLElement, typeof actionDefinitions> = { click: () => interaction(actionDefinittions.click) }
```

So basically, `Actions` is an *input* to `ActionMethods`.

The main change in this PR is that previously we always did this conversion right at the *edge*, so we never actually passed around any `ActionMethods`. The problem with this is that an `Action` needs to be aware of the type of element it operates on, and this means we need to carry forward the element type as well, which causes problems in terms of variance. Since the element is used as a parameter in a function type, it makes the type parameter contravariant. With this PR we instead do this conversion earlier, but that means using `ActionMethods` in type signatures more. Both types already existed before.

dagda1:
@jnicklas so this will add the ability to extend interactors?  Are there tests for this?

One thing that jumps out at me and might be problematic is the return type of `makeBuilder`  `: T & InteractorBuilder<E, FP, AM> {`

If the goal is to merge objects then this might not work as you expect.  

[Here](https://www.typescriptlang.org/play?ssl=13&ssc=3&pln=1&pc=1#code/C4TwDgpgBAshBOBzCB5ARgKwgY2AZwB4BBAGigCEA+KAXigG8BYAKCigG0BpKASwDsoAawggA9gDMKAXQBcUAK58AJhHH8ISqBAAewCMrwUuUlmzYB+KNx16DQkRKhFTZi1ACi27ABt5KguTGZIoqanwa1AA+TsYurnKBnCasZgmxzAC+UABkDC5cvALCYpJEslZauvpKhsWO5HFQliGq6po21YaJya5unj5+EAFBCsqt4UpRMUmNbGkzKXPTyRkA3CwsoJBOtHmLUACGcnjA8PyI6-tox6fnlxkbzFvQ5LtM+0dQfPIAtmgIl1c1y+v3+8Huj2eUAAwrsiDkKJcWCofAd4NBsKI+CcoNg5NDLtgAHQHKAAejJXwgADcECxiWhyZTwrT4JDwNAACK7OBIVCYHD4YhkKhI5go7xojFYnFKOScsVKEmrNgUqAnM58RDIoloFVM9W3LXs7buHkIZDoLC4QjkMhESiKnCS9G4mXALRydxiiAkg0a84sX2MtUBrVAA) is a playground that shows the problem and a `MergeObject` type that might work better.

dagda1:
@jnicklas ignore that playground.  [here is a codesandbox](https://codesandbox.io/s/sm1b2)

jnicklas:
>  Are there tests for this?

Apparently not 😬 Hmm, I was sure I'd added some. Anyway, tests for both types and implementation have been added.

> If the goal is to merge objects then this might not work as you expect.

Yes, indeed, I added some tests for overriding filters and actions, and ended up using this type, which seems to work fine:

``` typescript
export type MergeObjects<A, B> = B & Omit<A, keyof B>;
```

dagda1 (approved):
Great work.  `extends` is seriously useful.

PR #782, opened 2021-01-21
Add `changeOrigin` flag to application proxy server

### LINKED ISSUE #781: proxy fails for `https` urls
When running bigtest server against a public url that is ssl results in the following error instead of :

```
Proxy error: Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames: Host: localhost. is not in the cert's altnames: DNS:dev.folio.org, DNS:*.dev.folio.org
```

This is because of this issue https://github.com/http-party/node-http-proxy/issues/782

We can fix this by setting the `changeOrigin` property on the proxy server.

To work around this before it is released, you can monkey patch `node_modules/@bigtest/server/dist/src/proxy.js` line `80`

```diff
--- let proxyServer = proxy.createProxyServer({ target: target, selfHandleResponse: true });
+++ let proxyServer = proxy.createProxyServer({ target: target, selfHandleResponse: true, changeOrigin: true });
```

cowboyd:
> resolves https://github.com/thefrontside/bigtest/issues/781

Motivation
-----------

Whenever we fetch from an https proxy, by default the origin of the request to the target will be made with the host of the proxy server. However, when verifying a public url over TLS it will see that the certificate does not have `localhost` as one of its altnames and so will fail. In other words a cert for `x.com` will validate content from neither `y.com` nor `z.com` and certainly not `localhost`. It will only work for  `x.com`.

Approach
----------

This sets the `changeOrigin` flag on the proxy server which causses the outgoing requests from the proxy server [to have the same origin as the target][1]. Iin the case above, would cause the proxy request to `x.com` to have that origin as well, so that TLS validation succeeds.

It's difficult to say what other impact this change will have, which is the only thing that worries me.

[1]: https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/test/lib-http-proxy-common-test.js#L291-L315

### DIFF
--- a/.changeset/good-bobcats-share.md
+++ b/.changeset/good-bobcats-share.md
@@ -1,5 +1,6 @@
 ---
 "@bigtest/server": patch
+"bigtest": patch
 ---
 
 Add `changeOrigin` flag to proxy server so that app urls served over SSL will work properly
--- a/.changeset/good-bobcats-share.md
+++ b/.changeset/good-bobcats-share.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/server": patch
+---
+
+Add `changeOrigin` flag to proxy server so that app urls served over SSL will work properly
--- a/packages/server/src/proxy.ts
+++ b/packages/server/src/proxy.ts
@@ -95,7 +95,7 @@ export const startProxyServer = (serviceSlice: Slice<ServiceState<ProxyStatus, P
 
   proxyStatus.set({ type: 'starting' });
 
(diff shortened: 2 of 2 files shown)


.changeset/good-bobcats-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigtest/server": patch
```
cowboyd, line 2:
```suggestion
"@bigtest/server": patch
"bigtest": patch
```

minkimcello (approved):
👍

Issue #772, opened 2021-01-15
Filter Matchers


cowboyd:
Filters are a great way to make assertions on the properties of a UI element, but they can sometimes feel awkward because there is no way to make an abstraction over the value you're checking against because they only support a strict equals comparison against a single value.

This is why we've had to provide outlets like assert to serve as an escape hatch so that we can make more bespoke assertions using our interactors. An example of this is making an assertion on the `className` of an element. Currently, because the actual names of the classes of a component are many, and subject to change, an assertion like:

``` typescript
component.has({ className: 'hidden' });
```

isn't really reasonable because you simply can't depend on knowing the fully reified values of some properties like `className`, and so we have to punch out to an external assertion library:

``` typescript
import expect from 'expect';

component.assert(el => expect(el.className).toContain('hidden'));
```

This works, but it's not nice for several reasons:

- it requires you to choose an off the shelf assertion library in addition to interactors
- it introduces a separate error reporting format into your test flow that force you to context switch between what type of error you're reading.
- it can't be used to lookup or filter items and so there's another general capability of our filters that is lost. One of the great things I think about filters is that finding and asserting are essentially the same which lightens the cognitive overhead substantially.

What if we had a way to pass some sort of matcher to a filter value? That way, we could support more abstract assertions, while at the same time maintaining a way to use those abstract assertions to narrow items, and we can also plug these into our error reporting paradigm.

Something like:

``` typescript
import { containing } from '@bigtest/interactor';

component.has({ className: containing('hidden') });
```

This would allow us to plug right into our error messaging table except using `className containing 'hidden'' as the expected value. Not only that, we could even use the matcher to limit the scope.

``` typescript
import { Paragraph, startingWith } from '@bigtest/interactors';

const taleOfTwoCities = Paragrah({ text: startingWith('It was the best of times. It was the worst of times.' })
```

This would be used not just to make an assertions, but also to actually filter the list of candidate paragraphs. I'd imagine that we'd want to be able to use it for locators as well, so we could even say:

``` typescript
Paragraph(startingWith('It was the best of times. It was the worst of times.' )
```

cowboyd:
This was closed by #797

PR #783, opened 2021-01-21
Add matchers to interactors

jnicklas:
Matchers can be used for more flexible matching in cases where equality is not a good candidate. This change does not ship with any matchers, but in the future we can ship these along with the library.

See #772

### DIFF
--- a/packages/interactor/src/match.ts
+++ b/packages/interactor/src/match.ts
@@ -3,6 +3,7 @@ import { Locator } from './locator';
 import { Filter } from './filter';
 import { Filters } from './specification';
 import { escapeHtml } from './escape-html';
+import { MaybeMatcher, isMatcher } from './matcher';
 
 const check = (value: unknown): string => value ? "✓" : "⨯";
 
@@ -52,7 +53,7 @@ export class Match<E extends Element, F extends Filters<E>> {
 
 export class MatchLocator<E extends Element> {
   public matches: boolean;
-  public expected: string | null;
+  public expected: MaybeMatcher<string> | null;
   public actual: string | null;
 
   constructor(
@@ -61,7 +62,11 @@ export class MatchLocator<E extends Element> {
   ) {
     this.expected = locator.value;
     this.actual = locator.locatorFn(element);
-    this.matches = this.actual === this.expected;
+    if(isMatcher(this.expected)) {
+      this.matches = this.expected.match(this.actual);
+    } else {
+      this.matches = isEqual(this.actual, this.expected);
+    }
   }
 
   formatActual(): string {
@@ -79,7 +84,7 @@ export class MatchLocator<E extends Element> {
 
 export class MatchFilter<E extends Element, F extends Filters<E>> {
   public matches: boolean;
-  public items: MatchFilterItem<E, F>[];
+  public items: MatchFilterItem<unknown, E, F>[];
 
   constructor(
     public element: E,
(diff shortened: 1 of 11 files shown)


packages/interactor/src/matcher.ts
```diff
@@ -0,0 +1,11 @@
+export interface Matcher<T> {
+  bigtestMatcher: true;
+  match(actual: T): boolean;
+  format(): string;
+}
+
+export type MaybeMatcher<T> = Matcher<T> | T;
+
+export function isMatcher<T>(value: MaybeMatcher<T>): value is Matcher<T> {
+  return value && (value as any).bigtestMatcher === true;
```
jnicklas: Is there any way to avoid this explicit `any`?

packages/interactor/src/match.ts
```diff
@@ -61,7 +62,11 @@ export class MatchLocator<E extends Element> {
   ) {
     this.expected = locator.value;
     this.actual = locator.locatorFn(element);
-    this.matches = this.actual === this.expected;
+    if(isMatcher(this.expected)) {
+      this.matches = this.expected.match(this.actual);
+    } else {
+      this.matches = isEqual(this.actual, this.expected);
+    }
```
cowboyd: I think we can achieve a greater symmetry if we just have an `equals` matcher which is the default. That way, we don't need to have separate codepaths depending on `isMatcher` except in one place:

```ts
import { equals } from './matchers';

//......

this.matcher = isMatcher(this.expected) ? this.expected : equals(this.expected);
```

Where the equals matcher would be something akin to:

```ts
export function equals(expected: unknown): Matcher<unknown> {
  return {
    match: (actual) => isEqual(actual, expected),
    format: () => JSON.stringify(expected)
  };
}
```

I think that this would also eliminate the need for the `MaybeMatcher` type since we could immediately convert any value into a matcher.

This asymmetry will become more pronounced if we extend the matcher interface which is a possibility if, for example, if we wanted to add an `diff()` method that was invoked for details when a match failed.

packages/interactor/src/matcher.ts
cowboyd: I think I'd prefer just to do a structural match on the methods of the object rather than requiring a `bigTestMatcher` attribute which feels awkward when creating matchers and would probably necessitate a special `createMatcher` function that would do nothing but add the `bigTestMatcher` attribute to the object. In other words, when we get down to defining our own matchers, or especially when users define their own matchers, it's a much simpler story to say:

```ts
export function longerThan(length: number): Matcher<string> {
  return {
    match: (actual: string) => string.length > length,
    format: () => `longer than ${length}`
  };
}
```
instead of 

```ts
import { createMatcher } from '@bigtest/interactor';

export function longerThan(length: number): Matcher<string> {
  return createMatcher({
    match: (actual: string) => string.length > length,
    format: () => `longer than ${length}`
  });
}
```

To answer your original question though, I think that if we just match on the structure of the object and not a boolean attribute, we should be able to just say:

```ts
export function isMatcher<T>(value: unknown): value is Matcher<T> {
  let matcher = value as Matcher<T>
  return matcher && typeof matcher.match === 'function' && typeof matcher.format === 'function';
}
```

cowboyd (commented):
With this enhancement in place, this is really going to make using interactors incredibly powerful.

It's very interesting indeed that the `Matcher` interface is completely non-specific to interactors.

packages/interactor/src/matcher.ts
jnicklas: My reasoning for this was that I wanted to avoid the rare edge case where we accidentally think something is a matcher, but this is probably not ever going to happen, so it doesn't really matter.

packages/interactor/src/match.ts
jnicklas: I think the `MaybeMatcher` type is still needed, because we can use it for the input type for filters and locators.

jnicklas: While the idea of a default matcher is nice, in practice it was actually slightly easier to just do conditionals.

cowboyd (approved):
> While the idea of a default matcher is nice, in practice it was actually slightly easier to just do conditionals

I hear you, and I don't think it will be an issue unless we decide to extend the matcher interface. In that case, we can always add a `matcherFor(value: MaybeMatcher): Matcher` that does the conversion for us.

jnicklas:
I ran into problems here, where matchers which are not `Matcher<string>` given as locators are allowed by TS, even though they should be rejected. Unfortunately I do not think we can fix this problem right now. See #792 for a more detailed explanation.

PR #788, opened 2021-01-23
spike playwright driver

cowboyd:
This demonstrates a playwright integration where playwright is used to create a bigtest agent:

It takes about 30 lines of code to integrate playwright into bigtest, and now we can playwright instead of webdriver to spin up bigtest web agents https://github.com/thefrontside/bigtest/blob/playwright-driver/packages/playwright/src/index.ts

The following demonstrates running a bigtest test suite in a playwright browser

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
       "packages/agent",
       "packages/driver",
       "packages/webdriver",
+      "packages/playwright",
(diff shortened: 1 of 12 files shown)


cowboyd:
This was a good spike, and we'll keep the branch with us

PR #789, opened 2021-01-25
Add a gitignore to the sample app

jenweber:
This PR adds a gitignore to the sample app that is created with `npx bigtest-sample`.

Although `npx bigtest-sample` does not initialize the project as a git repo, I think it's probably common that the users will. Without this, if you run `git init`, VSCode gets a bit upset.

### DIFF
--- a/sample/.gitignore
+++ b/sample/.gitignore
@@ -1,0 +1,4 @@
+node_modules
+.bigtest
+.cache
+dist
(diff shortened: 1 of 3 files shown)


minkimcello (approved):
👍 could you bump the `bigtest-sample` version?

jenweber:
Sure thing! Done!

PR #795, opened 2021-01-27
Add HTML and FormField as base interactors

jnicklas:
These are reusable base interactors which serve as a good starting point for constructing more advanced interactors. Also uses those interactors for the built in interactors.

### DIFF
--- a/packages/interactor/src/definitions/form-field.ts
+++ b/packages/interactor/src/definitions/form-field.ts
@@ -14,8 +14,8 @@ const FormFieldInteractor = HTML.extend<FieldTypes>('form field')
 
 /**
  * Use this {@link InteractorConstructor} as a base for creating interactors which
- * work with form field. This provides some basic functionality which is convenient
- * for most form fields.
+ * work with form fields. This allows you to use field labels as locators, and also provides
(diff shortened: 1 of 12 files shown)


cowboyd (commented):
This looks good, although we need to revamp now the section on how to write your own interactors since pretty much every interactor is, at the very least going to extend from HTML.

packages/interactor/src/definitions/form-field.ts
```diff
@@ -0,0 +1,50 @@
...
+    disabled: {
+      apply: (element) => element.disabled,
+      default: false
+    },
+  })
+
+/**
+ * Use this {@link InteractorConstructor} as a base for creating interactors which
+ * work with form field. This provides some basic functionality which is convenient
+ * for most form fields.
```
cowboyd:
```suggestion
 * work with form fields. This allows you to use field labels as locators, and also provides
 * some basic filters which are convenient for most form fields.
```

cowboyd (approved):

cowboyd (approved):
Awesome! Added #799 to track the guide for these changes.

PR #797, opened 2021-01-27
Built in matchers

jnicklas:
This adds some built in matchers which can be used with the new matcher feature. See #772.

The matchers provided by this are two basic matchers for strings:

- `including(subString)`
- `matching(regexp)`

Two matchers for iterables:

- `some(matcher)`
- `every(matcher)`

And three combinators:

- `and(...matchers)`
- `or(...matchers)`
- `not(matcher)`

Those were the ones that seemed the most useful to me. There are plenty more string matchers which could be useful, like `startingWith`, and similar, but I figured that it'd be better to wait how often these seem to be needed before including them.

Depends on #795


cowboyd (commented):
What do you think about naming these `including` and `matching` rather than `includes` and `matches` since I think it will read more nicely with our `is/has` assertion syntax:

```
HTML(including("Hello")).exists();
Link("Sign In").has({ id: matching(/signin/) })
Button("Submit").has({ className: including('disabled') });
```

jnicklas:
Two more matchers I just realized could be useful are `allOf` and `anyOf`, for example:

``` typescript
Button("Submit").has({ classes: anyOf('disabled') });
MultiSelect("Colors").has({ values: allOf(matching(/red/)) });
```

Where `classes` is an array.

cowboyd (approved):

Issue #705, opened 2020-12-04
Review & iterate on the locators-filters-actions page


jenweber:
To me, this page's draft still feels a little too vague to be useful beyond the initial learning experience. It could use some more real-world context and explanation.

cowboyd:
Key points to include based on my session with @jbolda yesterday:

Relationship between selector, locator and filters is particularly important.

`selector` -> used to find a  set of candidate nodes for an interaction.
`locator` -> **user readable** string that identifies a node from the list of candidates. It is important to think about this in the way a user would.
`filter` -> some value that can be matched against to _narrow_ the list of candidate nodes.

other points:

* all nodes selected by `selector` should represent the same starting point in a DOM hierarchy
* the locator, plus all filters and actions will be relative to 
* it is A-OK to dig into the DOM relative to the that base node. E.g. If you have a component whose current HTML looks like:

```html
<div class="text-field">
  <label for="password-field">Password</label>
  <input type="text" id="password-field">
  <span class="errors">Must be at least 8 characters</span>
</div>
```

It is encouraged to dig into to get your locator and filters:

```ts
export const TextField = createInteractor("text field")({
  selector: 'div.textfield',
  locator: element => element.querySelector('label')?.textContent,
  filters: {
    errorText: element => element.querySelector('.errors')?.textContent
  }
});
```

thus allowing you to say:

```ts
await TextField("Password").fillIn('too-short');
await TextField("Password").has({ errorText: "Must be at least 8 characters" });
```

jenweber:
Closed by https://github.com/thefrontside/bigtest/pull/798

PR #798, opened 2021-01-27
Updates Interactors docs with revised copy

jorgelainfiesta:
## Motivation

We want to make sure our copy is at its best so it's easier for people to follow it.

## Approach

I made some rounds of revisions with our editor on the Interactors docs. The Platform docs are still pending.

### DIFF
--- a/website/docs/interactors/1-quick-start.md
+++ b/website/docs/interactors/1-quick-start.md
@@ -150,12 +150,13 @@ Interactors have methods like `click` that mimic user actions. If you are using 
   </TabItem>
 </Tabs>
 
-If you are using the sample project, you can find these examples at `bigtest-sample/src/test/jest.test..js`,  `bigtest-sample/src/test/cypress.spec.js` and  `bigtest-sample/src/test/bigtest.test.js`.
+If you are using the sample project, you can find these examples at `bigtest-sample/src/test/jest.test..js`, `bigtest-sample/src/test/cypress.spec.js` and `bigtest-sample/src/test/bigtest.test.js`.
 
 Now run your tests as you usually would or use any of the sample project options. Congratulations – you used your first Interactor!
 
 :::note The BigTest Platform
- There's more to BigTest than just Interactors. BigTest can also run your tests on any _real_ browser – just like your users use. We have built a new integrated platform from the ground up to help you test more with less effort. And best of all it's free and Open Source! [Check it out](/platform), and let us know what you think..
+ There's more to BigTest than just Interactors. BigTest can also run your tests on any _real_ browser – just like your users use. We have built a new integrated platform from the ground up to help you test more with less effort. And best of all it's free and Open Source! [Check it out](/platform), and let us know what you think.
+
 :::
 
 ## Making test assertions
@@ -312,7 +313,7 @@ As you can see, Interactors not only make it simple to use DOM elements like Rad
 
 Try using more of the [Built-in Interactors](2-built-in-dom.md) within your tests such as `Link`, `Checkbox`, `TextField`, and more.
 
-You’ll quickly realize how much more powerful Interactors when combined with [Locators, Filters, and Actions](3-locators-filters-actions.md).
+You’ll quickly realize how much more powerful Interactors are when combined with [Locators, Filters, and Actions](3-locators-filters-actions.md).
 
 Once you are comfortable with those, you’ll want to [write your own Interactors](4-write-your-own.md) to explore just how useful they can be.
 ### Join the BigTest community
--- a/website/docs/interactors/1-quick-start.md
+++ b/website/docs/interactors/1-quick-start.md
@@ -4,7 +4,7 @@ title: Quick Start
 slug: /interactors
 ---
 
-BigTest's Interactors makes writing your UI tests easier, faster, and more reliable. You can use them across many different testing frameworks.
+BigTest's Interactors make writing your UI tests easier, faster, and more reliable. You can use them across many different testing frameworks.
 
 By the end of this page you will be testing one of your own apps with Interactors.
 
@@ -27,7 +27,7 @@ Run the following command to install a sample project which includes a simple ap
 $ npx bigtest-sample
 ```
 
-After you’ve installed the project, you’ll be able to run the test suite you desire. For example, for running the Jest suite you can use `npm run test:jest`. More details are available in the welcome message for the project.
+After you’ve installed the project, you’ll be able to run the test suite of your choice. For example, for running the Jest suite you can use `npm run test:jest`. More details are available in the welcome message for the project.
 
 ## Installation
 
--- a/website/docs/interactors/1-quick-start.md
+++ b/website/docs/interactors/1-quick-start.md
@@ -4,9 +4,9 @@ title: Quick Start
 slug: /interactors
 ---
 
-BigTest's Interactors make your UI tests easier to write, faster, and more reliable. You can use them across many different testing frameworks.
+BigTest's Interactors makes writing your UI tests easier, faster, and more reliable. You can use them across many different testing frameworks.
 
-By the end of this quick start, you will be testing one of your own apps with Interactors.
+By the end of this page you will be testing one of your own apps with Interactors.
 
 :::note We're here to help
  If you need help or have questions along the way, please let us know in [the Discord chat](https://discord.gg/r6AvtnU) or [open a discussion](https://github.com/thefrontside/bigtest/discussions/) on Github.
@@ -17,17 +17,17 @@ By the end of this quick start, you will be testing one of your own apps with In
 - [npm](https://www.npmjs.com/get-npm) or [yarn](https://classic.yarnpkg.com/en/docs/install)
 - A web app*
 
-*To follow along with this tutorial, you need an app that uses [Jest](https://jestjs.io/), [Cypress](https://www.cypress.io/), or BigTest. For the most interesting learning experience, you should use one of your own apps, or you can use the sample app described below.
+*To follow along with this tutorial, you need an app that uses [Jest](https://jestjs.io/), [Cypress](https://www.cypress.io/), or BigTest. For the most informative learning experience you should use one of your own apps. If you don’t have one available you can use the sample app below
 
 ### Sample app
 
-If you do not have an app of your own to use with this tutorial, run the following command to install a very simple app that has Jest, Cypress, and BigTest platform pre-installed and configured:
+Run the following command to install a sample project which includes a simple app with test suites implemented for Jest, Cypress, and BigTest:
 
 ```
 $ npx bigtest-sample
 ```
 
-This command will create a project and output instructions in the console on how you can run tests on any of the three testing platforms. Keep reading to learn about what is inside the sample app!
+After you’ve installed the project, you’ll be able to run the test suite you desire. For example, for running the Jest suite you can use `npm run test:jest`. More details are available in the welcome message for the project.
 
 ## Installation
 
@@ -86,11 +86,9 @@ If you are using Cypress, you will only need to install `@bigtest/cypress`:
 
 ## Import Interactors in your test suite
 
-Interactors can do a lot, but let's keep things simple and start with testing one of the most common user interactions - a button click.
+Interactors can do a lot, but let's keep things simple for now and begin testing one of the most common user interactions - a button click.
 
-In your codebase, find a test that already has a button click in it. In that test, import the `Button` interactor from `bigtest`, and use it to replace the click interaction. Interactors have methods like `click` that mimic user actions.
-
-Click on the tabs below to see some examples within different testing frameworks:
+Interactors have methods like `click` that mimic user actions. If you are using your own app, in your codebase find a test that already has a button click in it. In that test, import the `Button` interactor from `bigtest` and use it to replace the click interaction as exemplified below (make sure to substitute "Sign In" with your own button text):
 
 <Tabs
   defaultValue="jest"
@@ -152,19 +150,21 @@ Click on the tabs below to see some examples within different testing frameworks
   </TabItem>
 </Tabs>
 
-Make sure to substitute "Sign In" with your own button text.
+If you are using the sample project, you can find these examples at `bigtest-sample/src/test/jest.test..js`,  `bigtest-sample/src/test/cypress.spec.js` and  `bigtest-sample/src/test/bigtest.test.js`.
 
-Now, run your tests! Congratulations, you used your first Interactor.
+Now run your tests as you usually would or use any of the sample project options. Congratulations – you used your first Interactor!
 
 :::note The BigTest Platform
- There's more to BigTest than Interactors. BigTest can also run your tests on any _real_ browser! We have built a new integrated platform from the ground up to help you test more with less effort. And yes, it's free and Open Source too! [Check it out](/platform), and let us know what you think.
+ There's more to BigTest than just Interactors. BigTest can also run your tests on any _real_ browser – just like your users use. We have built a new integrated platform from the ground up to help you test more with less effort. And best of all it's free and Open Source! [Check it out](/platform), and let us know what you think..
 :::
 
 ## Making test assertions
 
 Interactors have methods to help you make test assertions, like checking if a Button is `absent` or if it `exists`.
 
-In this example, we are testing that the "Sign In" button disappears after a click, and a "Log Out" button appears:
+Let’s build on top of the example you used in the last section. When a user clicks a button, _something_ is expected to happen. We want to check if that something does happen after we use an Interactor to click the button.
+
+In the sample project, when you click the “Sign In” button it disappears and a “Sign out” button appears in its place. If you’re following along your own test suite, find a test that asserts whether an element exists. Writing this kind of assertion would look like this with Interactors:
 
 <Tabs
   defaultValue="jest"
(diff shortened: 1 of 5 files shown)


website/docs/interactors/1-quick-start.md
```diff
@@ -152,19 +150,21 @@ Click on the tabs below to see some examples within different testing frameworks
   </TabItem>
 </Tabs>
 
-Make sure to substitute "Sign In" with your own button text.
+If you are using the sample project, you can find these examples at `bigtest-sample/src/test/jest.test..js`,  `bigtest-sample/src/test/cypress.spec.js` and  `bigtest-sample/src/test/bigtest.test.js`.
```
minkimcello: There are extra spaces before cypress and bigtest.

website/docs/interactors/1-quick-start.md
```diff
@@ -152,19 +150,21 @@ Click on the tabs below to see some examples within different testing frameworks
...
 
-Now, run your tests! Congratulations, you used your first Interactor.
+Now run your tests as you usually would or use any of the sample project options. Congratulations – you used your first Interactor!
 
 :::note The BigTest Platform
- There's more to BigTest than Interactors. BigTest can also run your tests on any _real_ browser! We have built a new integrated platform from the ground up to help you test more with less effort. And yes, it's free and Open Source too! [Check it out](/platform), and let us know what you think.
+ There's more to BigTest than just Interactors. BigTest can also run your tests on any _real_ browser – just like your users use. We have built a new integrated platform from the ground up to help you test more with less effort. And best of all it's free and Open Source! [Check it out](/platform), and let us know what you think..
```
minkimcello: Should there be an extra `.` at the end? Or take one out?

website/docs/interactors/1-quick-start.md
```diff
@@ -306,20 +304,17 @@ Here's an example of what a test for an airline datepicker interface could look
...
 
-Try using more of the [Built-in Interactors](2-built-in-dom.md) within your tests, such as `Link`, `Checkbox`, `TextField`, and more.
-
-Once you are comfortable with those, you are ready to [write your own Interactors](4-write-your-own.md).
-
-### Try out more BigTest features
+Try using more of the [Built-in Interactors](2-built-in-dom.md) within your tests such as `Link`, `Checkbox`, `TextField`, and more.
 
-Once you are comfortable with those, you are ready to [write your own Interactors](/docs/interactors/write-your-own).
+You’ll quickly realize how much more powerful Interactors when combined with [Locators, Filters, and Actions](3-locators-filters-actions.md).
```
minkimcello: I think we're missing a word here.

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -3,25 +3,23 @@ id: locators-filters-actions
...
-## Locators
+## The Locator
 
-One benefit of Interactors is that they help you align your tests with how a user actually interacts with the app, starting with finding what to act upon.
+One benefit of Interactors is that they help you align your tests with how a user actually interacts with the app, starting with finding what to act upon. A locator is the simplest way to find a specific element in a user interface.
 
-Locators are the simplest way to find a specific element in a user interface.
-
-Whenever you use an Interactor in a test, you can pass it a string, like "Submit" in the example below. This string argument would be the Locator.
+When you use an Interactor in a test, you can pass it a string. This string argument would be the Locator - like "Submit" in the example below.:
```
minkimcello: Should it be either `:` or `.` but not both?

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -52,68 +50,81 @@ TextField('Username:', { id: 'username-id' }).exists();
  _See the source code of the TextField interactor [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts)_.
 :::
 
-You can think of locators as the "default filter" because filters and locators both serve the same functionality. The reason why we offer both solutions is for your convenience because having to pass in an object for each interactor can be repetitive.
+You can think of the locator as the "default filter" because filters and locators both provide the same functionality. The reason why BigTest offers both solutions is convenience, because having to pass in an object for each interactor can get repetitive – unless you use a filter.
```
minkimcello: > because having to pass in an object for each interactor can get repetitive
This was to point out that `Button('Submit')` is more convenient than having to use filters like `Button({ value: 'Submit' })` each time.

So adding `- unless you use a filter` causes confusion.

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -52,68 +50,81 @@ TextField('Username:', { id: 'username-id' }).exists();
...
-So how are Filters useful? Most forms have multiple textfields and if they do not have labels or if there are multiple labels with the same value, that's when you would need to use a filter. Below is an example of a form with textfields that do not have labels:
 ```html
 <form>
   <input id='username-id' type='text' placeholder='USERNAME'/>
   <input id='password-id' type='password' placeholder='PASSWORD'/>
 </form>
 ```
 
-As they do not have labels, we would not be able to use a locator in this scenario. Using `TextField()` would return two elements and therefore an error. We can narrow it down using either the `id` or `placeholder` filters provided by the BigTest `TextField` interactor:
+We cannot specify a locator based on the label, so using `TextField()` would return two elements and therefore produce an error. We can narrow  down from two TextFields to one using either the `id` or `placeholder` filters provided by the BigTest `TextField` interactor:
```
minkimcello: Extra space

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -52,68 +50,81 @@ TextField('Username:', { id: 'username-id' }).exists();
...
+The difference between this approach and using `exists()` is that `exists()` is less refined as it will succeed as long as there is at least one match. With the two textfields for the username and password, the assertion would fail on if we were to write a test like this:
 
 ```js
 TextField().has({ placeholder: 'USERNAME' });
 ```
 
-This assertion would fail on account of the textfield that has the placeholder value `PASSWORD`. So you will need to choose the assertion method that is most appropriate for your tests.
+It would fail on account of the textfield that has the placeholder value `PASSWORD`. You therefore need to choose the assertion method that is most appropriate for your tests.
+
+Lastly, there is also the `is()` method, which functions in an almost identical way to `has()`. The only difference is in the semantics so that your tests can read better.
```
minkimcello: It's not functionally almost identical but actually identical.

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -52,68 +50,81 @@ TextField('Username:', { id: 'username-id' }).exists();
...
 
-Lastly, there is also the `is()` method which is identical to `has()`. The difference is only in semantics so that your tests can read better. For instance, if we wanted to test if a textfield is visible, `has()` would work perfectly fine but we would write the test using `is()` instead:
+For instance, if we wanted to test if a textfield is visible, `has()` would work perfectly fine; but writing the test using `is()` would look slightly differently like this:
```
minkimcello: I think in this context maybe we should point out that it looks more "natural"?

website/docs/interactors/4-write-your-own.md
```diff
@@ -3,28 +3,28 @@ id: write-your-own
...
+There are four things to decide when creating an interactor:
 1. What to name and label the interactor
 2. Which HTML element or elements to target
-3. The locator and filters, which helps users be able to narrow down the element they want to reference.
-4. Actions that a test should be able to `perform` on that element, like `click`
+3. The locator and filters, which helps users be able to narrow down the element they want to reference
+4. The action or actions that a test should `perform` when using the interactor (like `click`)
 
-Putting this information together, we can make this new Interactor:
+In this case, let's call our Interactor ‘MyTextField’, and we’ll label it as ‘my-textfield-interactor’ to distinguish it from the build-in one. Let’s say that MyTextField will target a regular input with a custom class `.my-input`, which means its HTML selector would be `input[type=text].my-input`. For locator we’ll use the input’s placeholder, and we’ll define a value filter. Finally, we’ll define a `fillIn` action for MyTextField interactor. Putting this together, we get the following definition of an Interactor:
```
minkimcello: > build-in

built-in

website/docs/interactors/4-write-your-own.md
```diff
@@ -69,40 +70,44 @@ actions: {
...
 export const Form = createInteractor<HTMLFormElement>('form')({
   selector: 'form',
   actions: {
-    async submit(interactor){
-      await interactor.find(Button('Submit')).click();
-    }
+	async submit(interactor){
+  	  await interactor.find(Button('Submit')).click();
+	}
   }
```
minkimcello, line 92: i like the 2-space tabbing and that's what we've been using everywhere else in the docs

website/docs/interactors/4-write-your-own.md
```diff
@@ -69,40 +70,44 @@ actions: {
...
 ```
 
-Delegating the click action will save you the hassle of having to implement the click separately, and there is also the added benefit of making your tests easier to read:
+There’s two peculiarities about this example. First, notice we’re not using `perform` which means that we’re not performing an action on any element directly. Secondly, within the submit action definition you’ll notice that we’re using `find` to access another Interactor and calling its action. That’s what we mean with ‘delegating’ an action.
+
+Why would you want to delegate actions within an Interactor action definition? Well, it can help you avoid repeating yourself and to keep Interactors as black boxes. For example, let’s say we wanted to test that a user submits a form, without delegation we can reach in in the Form Interactor and issue a click action:
 
 ```js
-Form.find(Button('Submit')).click();
+Form.find(Button(‘Submit')).click();
```
minkimcello: > ‘Submit'

typo

website/docs/interactors/4-write-your-own.md
```diff
@@ -345,22 +353,24 @@ DatePicker().find(Button('31')).click();
...
 The first step is to see if you can go through the interaction yourself in the browser by using only [keyboard navigation](https://webaim.org/techniques/keyboard/).
-If you cannot get to the end successfully, then you just found a bug in your app!
+If you cannot get to the end successfully, then you just found a bug in your app.
 Although many people navigate an interface by sight and clicking,
-others may use assistive technology such as screen readers, and keyboard support is critical.
-For example, have you ever seen a click mistakenly attached to a `div` instead of a button?
-Those types of errors can make your app unusable to some people, and also difficult to test.
+others may use assistive technology such as screen readers and keyboard support is critical.
+For example, a click can mistakenly be attached to a `div` instead of a button.
+Those types of errors can make your app unusable to some people and also difficult to test.
```
minkimcello: Same here.

website/docs/interactors/4-write-your-own.md
```diff
@@ -345,22 +353,24 @@ DatePicker().find(Button('31')).click();
...
+When the built-in Interactors are not enough, we encourage you to write your own. They will help prevent duplicated logic in your tests, and if your interface changes, you only need to make changes to the Interactor instead of throughout the code.
 
-For example, let's say that you want to replace a custom datepicker with a popular third-party library instead.
-Although you may have many tests for flows with the date picker, only your Interactor needs to change.
+For example, let's say that you want to replace a custom datepicker with a popular third-party library instead. Although you may have many tests for flows with the date picker, only your Interactor needs to change.
 
 ### I have an interaction that is really difficult to test. What should I do?
 
 A good test suite helps you uncover hidden problems.
 Often, difficult UI tests are your early warning system for areas of your app that may have accessibility issues.
```
minkimcello: You can combine these two lines into one. Just a formatting thing. I think Jen likes to create new lines often when she works.

minkimcello (changes requested):

jorgelainfiesta:
Thanks Min! Most of those were my fault when putting back the new copy into the codebase, sorry.

website/docs/interactors/3-locators-filters-actions.md
minkimcello: @jorgelainfiesta i think you missed these three

website/docs/interactors/1-quick-start.md
```diff
@@ -4,9 +4,9 @@ title: Quick Start
 slug: /interactors
 ---
 
-BigTest's Interactors make your UI tests easier to write, faster, and more reliable. You can use them across many different testing frameworks.
+BigTest's Interactors makes writing your UI tests easier, faster, and more reliable. You can use them across many different testing frameworks.
```
cowboyd:
```suggestion
BigTest's Interactors make writing your UI tests easier, faster, and more reliable. You can use them across many different testing frameworks.
```

website/docs/interactors/1-quick-start.md
```diff
@@ -17,17 +17,17 @@ By the end of this quick start, you will be testing one of your own apps with In
...
 
-If you do not have an app of your own to use with this tutorial, run the following command to install a very simple app that has Jest, Cypress, and BigTest platform pre-installed and configured:
+Run the following command to install a sample project which includes a simple app with test suites implemented for Jest, Cypress, and BigTest:
 
 ```
 $ npx bigtest-sample
 ```
 
-This command will create a project and output instructions in the console on how you can run tests on any of the three testing platforms. Keep reading to learn about what is inside the sample app!
+After you’ve installed the project, you’ll be able to run the test suite you desire. For example, for running the Jest suite you can use `npm run test:jest`. More details are available in the welcome message for the project.
```
cowboyd:
```suggestion
After you’ve installed the project, you’ll be able to run the test suite of your choice. For example, for running the Jest suite you can use `npm run test:jest`. More details are available in the welcome message for the project.
```

website/docs/interactors/3-locators-filters-actions.md
cowboyd: I think I just realized that we provide `is()` so that boolean filters can read more naturally. Can anybody think of a non-boolean filter that reads better with `is()` I'm drawing a blank:

```js
TextField().is({ disabled: true });
TextField().is({ visible: false });
TextField().is({ valid: true });

TextField().has({ placeholder: 'address' });
```

If this is the case, then we should explicitly mention this so that folks know the rule.

website/docs/interactors/4-write-your-own.md
```diff
@@ -69,40 +70,44 @@ actions: {
...
 })
 ```
 
-Delegating the click action will save you the hassle of having to implement the click separately, and there is also the added benefit of making your tests easier to read:
+There’s two peculiarities about this example. First, notice we’re not using `perform` which means that we’re not performing an action on any element directly. Secondly, within the submit action definition you’ll notice that we’re using `find` to access another Interactor and calling its action. That’s what we mean with ‘delegating’ an action.
+
+Why would you want to delegate actions within an Interactor action definition? Well, it can help you avoid repeating yourself and to keep Interactors as black boxes. For example, let’s say we wanted to test that a user submits a form, without delegation we can reach in in the Form Interactor and issue a click action:
 
 ```js
 Form.find(Button('Submit')).click();
```
cowboyd:
```suggestion
Form().find(Button('Submit')).click();
```

website/docs/interactors/4-write-your-own.md
```diff
@@ -69,40 +70,44 @@ actions: {
...
+```
 
-// versus
+But if the Form component changes and now the button that submits is no longer `Submit` but `Send` we’d have to update all our tests to reflect that change. Instead we can encapsulate that detail in the `submit` action of `Form` and delegate the click action to the responsible party. Instead of reaching into internal component elements, who ever is using Form to test the form simply as:
 
+```js
 Form.submit();
```
cowboyd:
```suggestion
Form().submit();
```

website/docs/interactors/4-write-your-own.md
```diff
@@ -174,21 +179,19 @@ import TabItem from '@theme/TabItem';
...
 
 Check out the source code of [createInteractor()](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/create-interactor.ts) for more details.
 
-## Writing your second interactor
+## Writing a more complex interactor
 
-One of the greatest benefits of Interactors is that you can turn complex testing scenarios into readable assertions. Let's create an interactor for a table that navigates by row and column to make assertions about the table's data. Our goal is that when we are finished, we will be able to use it in our tests like this:
+One of the greatest benefits of Interactors is that you can turn complex testing scenarios into readable assertions. Let’s illustrate how that looks like with a commonly cumbersome UI piece: a table. We want to be able to easily assert the contents of our tables, and that means that we should be able to know the value in a cell given its column and row. Once we’re done creating a TableCell Interactor, we’ll be able to make that happen and have it look like this:
```
cowboyd:
```suggestion
One of the greatest benefits of Interactors is that you can turn complex testing scenarios into readable assertions. Let’s illustrate how that looks like with an historically cumbersome UI piece: a table. We want to be able to easily assert the contents of our tables, and that means that we should be able to know the value in a cell given its column and row. Once we’re done creating a TableCell Interactor, we’ll be able to make that happen and have it look like this:
```

website/docs/interactors/4-write-your-own.md
```diff
@@ -174,21 +179,19 @@ import TabItem from '@theme/TabItem';
...
 
 ```js
 TableCell({ columnTitle: 'Name', rowNumber: 2 }).has({ value: 'Marge Simpson' });
 ```
 
-First, let's look at the markup we are trying to filter through:
+First, consider a table markup defined as follows:
```
cowboyd:
```suggestion
First, consider some table markup defined as follows:
```

website/docs/interactors/3-locators-filters-actions.md
jorgelainfiesta: Isn't `is` inherently a boolean question?

minkimcello: But I'm worried saying `is() is for boolean assertions and has() for the rest` will make people think the functionality is different. So maybe we can phrase it so it's more of a suggestion?

minkimcello:
@jenweber Would you say this satisfies #705?

cowboyd:
@jorgelainfiesta We need to rebase this so that the newly improved CircleCI checks run.

website/docs/interactors/3-locators-filters-actions.md
minkimcello: @jorgelainfiesta @cowboyd

minkimcello (approved):

jenweber:
@minkimcello  yes I think so! I’ll close the issue.

website/docs/interactors/4-write-your-own.md
```diff
@@ -3,28 +3,28 @@ id: write-your-own
...
+There are four things to decide when creating an interactor:
 1. What to name and label the interactor
 2. Which HTML element or elements to target
-3. The locator and filters, which helps users be able to narrow down the element they want to reference.
-4. Actions that a test should be able to `perform` on that element, like `click`
+3. The locator and filters, which helps users be able to narrow down the element they want to reference
+4. The action or actions that a test should `perform` when using the interactor (like `click`)
 
-Putting this information together, we can make this new Interactor:
+In this case, let's call our Interactor ‘MyTextField’, and we’ll label it as ‘my-textfield-interactor’ to distinguish it from the built-in one. Let’s say that MyTextField will target a regular input with a custom class `.my-input`, which means its HTML selector would be `input[type=text].my-input`. For locator we’ll use the input’s placeholder, and we’ll define a value filter. Finally, we’ll define a `fillIn` action for MyTextField interactor. Putting this together, we get the following definition of an Interactor:
```
minkimcello, line 27: 🤔 i missed this. i don't know why the `custom class '.my-input'` part was added. it's not implemented in the example. nor is the `MyTextField` interactor name.

PR #803, opened 2021-02-01
ensure tsconfig config option is present and gets passed to the bundler

### LINKED ISSUE #800: tsconfig option not being defined when using bigtest inside a monorepo
An attempt was made by backstage developers at spiking a bigtest implementation (https://github.com/backstage/backstage/tree/59c234ed2ed3254164defcd8677170bee7a269f9). However, they quickly ran into a bundling issue where the default export of the testcase was not being included in the built manifest.

As a result, the manifest builder receives this error when trying to evaluate the manifest:

```
[manifest builder] build error:
Cannot read property 'map' of undefined
```

This is caused by the default export not being present. The test module looks like this:

```js
    var bigtest = /*#__PURE__*/Object.freeze({
        __proto__: null
    });
```
> Note: The module is named `bigtest` because in the example repo, the test case is in `src/bigtest/index.ts`. I tried changing the name of the test module and still encountered the same issue.

When it should look like

```js
    var bigtest = /*#__PURE__*/Object.freeze({
        __proto__: null,
        'default': the_actual_testcase
    });
```

This was tracked down to the `options.tsconfig` being undefined and so as a result not only was the bundle malformed, but also the entire monorepo was compiled in the `.bigtest/manifest/build` directory.

We need to make sure that we have a valid tsconfig when compiling typescript, and that we fail if one cannot be resolved, or worse if it resolves to a tsconfig in the root directory.

Reproduce
========

```
$ git clone https://github.com/backstage/backstage.git
$ cd backstage
$ git checkout 59c234ed2ed3254164defcd8677170bee7a269f9
$ yarn install
$ cd packages/e2e-test
$ yarn bigtest server
```

dagda1:
Fixes #800 where the `tsconfig` option of `bigtest.json` was not getting passed to the `manifest-builder` from the `orchestrator` and subsequently the `bundler`.

The option now gets passed from the `orchestrator`.

```ts
  yield fork(createManifestBuilder({
    watch: options.project.watchTestFiles,
    atom: options.atom,
    srcPath: manifestSrcPath,
    distDir: manifestDistDir,
    buildDir: manifestBuildDir,
    tsconfig: options.project.tsconfig
  }));
```
I struggled to write a test for this because there is so much going on in orchestrator and `undefined` is a valid value for `tsconfig` for non-typescript projects.

I added the following runtime check to `cli` which is a better check than a unit test IMHO.
 
 ```ts
export const ensureConfiguration = (config: ProjectOptions) => {
  if (typeof config.tsconfig !== 'undefined' && fs.existsSync(config.tsconfig) === false) {
    let errorMessage = `The \`tsconfig\` option of bigtest.json (\`${config.tsconfig}\`) is invalid.`;
    
    console.error(errorMessage);

    process.exit(1);
 }
```

### DIFF
--- a/packages/cli/src/ensure-configuration.ts
+++ b/packages/cli/src/ensure-configuration.ts
@@ -1,6 +1,5 @@
 import { ProjectOptions } from '@bigtest/project';
 import fs from 'fs';
-import path from 'path';
 
 export const ensureConfiguration = (config: ProjectOptions) => {
   if (typeof config.tsconfig !== 'undefined' && fs.existsSync(config.tsconfig) === false) {
--- a/packages/cli/src/ensure-configuration.ts
+++ b/packages/cli/src/ensure-configuration.ts
@@ -4,7 +4,7 @@ import fs from 'fs';
 
 export const ensureConfiguration = (config: ProjectOptions): void => {
   if (typeof config.tsconfig !== 'undefined' && fs.existsSync(config.tsconfig) === false) {
-    let message = `The \`tsconfig\` option of bigtest.json (\`${config.tsconfig}\`) is invalid.`;
+    let message = `The \`tsconfig\` option of (\`${config.tsconfig}\`) is invalid.`;
     
     throw new MainError({ exitCode: 1, message });
   }
--- a/packages/cli/src/ensure-configuration.ts
+++ b/packages/cli/src/ensure-configuration.ts
@@ -1,15 +1,11 @@
 import { ProjectOptions } from '@bigtest/project';
+import { MainError } from '@effection/node';
 import fs from 'fs';
 
 export const ensureConfiguration = (config: ProjectOptions): void => {
   if (typeof config.tsconfig !== 'undefined' && fs.existsSync(config.tsconfig) === false) {
-    let errorMessage = `The \`tsconfig\` option of bigtest.json (\`${config.tsconfig}\`) is invalid.`;
+    let message = `The \`tsconfig\` option of bigtest.json (\`${config.tsconfig}\`) is invalid.`;
     
-    console.error(errorMessage);
-
-    // is process.exit(1); sufficient from an effection standpoint or
-    // or should I use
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/cli/src/ensure-configuration.ts [hunk 1]
```diff
@@ -0,0 +1,15 @@
...
+export const ensureConfiguration = (config: ProjectOptions): void => {
+  if (typeof config.tsconfig !== 'undefined' && fs.existsSync(config.tsconfig) === false) {
+    let errorMessage = `The \`tsconfig\` option of bigtest.json (\`${config.tsconfig}\`) is invalid.`;
+    
+    console.error(errorMessage);
+
+    // is process.exit(1); sufficient from an effection standpoint or
+    // or should I use
+    // throw new Error(errorMessage);
+    process.exit(1);
```
cowboyd: You can throw a `MainError` from `@effection/node` See here for an example https://github.com/thefrontside/bigtest/blob/v0/packages/cli/src/start-server.ts#L23

packages/cli/src/ensure-configuration.ts
```diff
@@ -1,15 +1,11 @@
 import { ProjectOptions } from '@bigtest/project';
+import { MainError } from '@effection/node';
 import fs from 'fs';
 
 export const ensureConfiguration = (config: ProjectOptions): void => {
   if (typeof config.tsconfig !== 'undefined' && fs.existsSync(config.tsconfig) === false) {
-    let errorMessage = `The \`tsconfig\` option of bigtest.json (\`${config.tsconfig}\`) is invalid.`;
+    let message = `The \`tsconfig\` option of bigtest.json (\`${config.tsconfig}\`) is invalid.`;
```
cowboyd: small nit, but I believe we support loading different configuration files, so we can't assume that it will be named `bigtest.json`. If possible, we should point to the exact file.

cowboyd (approved):

Issue #801, opened 2021-01-30
malformed bundle is difficult to diagnose


cowboyd:
When encountering #800, this catestrophic misconfiguration only reported the realtively innocuous error message in the server output:

```
[manifest builder] build error:
Cannot read property 'map' of undefined
```

We should clearly indicate that bigtest could not process the bundle probably due to an internal error and that we should report the issue. It might even be panic-worthy.

dagda1:
I personally think insisting on a default export is problematic and I think we should work with named exports.  Default exports in typescript are not the norm which in someway has something to do with them initially being difficult to implement before `esModuleInterop` was added.

We don't appear to have any runtime checks and I don't think a generic error message is going to wash here.  

I created #530 which was refused and I do get the reasons but a generic error message with no indication that it is the default export will prove difficult to diagnose.

cowboyd:
I agree with you that in most cases a default export is not good, which we generally eschew it. However, in this case, not using a default export only sweeps the problem under the rug. We would need a "magic" named export that would have to be there, otherwise we couldn't discover and link the test. E.g. `export const theTest = test('My Test');` and then our bundler would have to look for an export called `theTest` instead of `default`. Same problem, different name.

All that aside, in this case, the problem wasn't the default export, right? The test did have one, but it was the typescript config was missing and so the default export wasn't being bundled. I field the issue so that we could have a placeholder to remind us that sometimes things will be borked, and we want to give as much of a diagnosis as possible.

dagda1:
> We would need a "magic" named export

I was forgetting about the reason why.  That is tricky and would possibly need some AST trickery to cure it.  The default export is good enough for now but we need really good validation that points to there not being a default export.

> All that aside, in this case, the problem wasn't the default export, right? The test did have one, but it was the typescript config was missing and so the default export wasn't being bundled. I field the issue so that we could have a placeholder to remind us that sometimes things will be borked, and we want to give as much of a diagnosis as possible.

Oh maybe there is more than one case.

This fails with the `.map` error also

```ts
import { test } from '@bigtest/suite';

export const t = test("No default export");
```

That is why I am being pedantic about this, if I get the big long error message that points to the github issue then I probably will not get that this is a basic validation error message.

I was thinking of adding something like this to `validate-test`:

```ts
  if ( [test?.assertions, test?.children].some(c => {
    return Array.isArray(c) === false;
  })) {
    throw new TestValidationError(`
Invalid Test: Test contains no assertions or children.
[...]

PR #804, opened 2021-02-01
fix linking of @bigtest/eslint-plugin rule to docs

dagda1:
While looking at #801 I remembered that there actually was an eslint rule for this.

The linking to the docs for the `@bigtest/require-default-export` is broken.

Eslint borks if it cannot find the definition.

This PR fixes it.

Maybe we should generate an `eslint.json` file as part of `bigtest init`?


dagda1:
closing in favour of #807

PR #806, opened 2021-02-01
Switch to using CircleCI instead of Github Actions

jnicklas:


jnicklas:
I think this is pretty much ready to go!

I tweaked the following:

- only running CLI tests on windows, this means we don't have to have a separate prepack step, this cuts the total runtime down from 15m to 10m.
- a single workflow for everything, linux/docker as the default platform
- only caching `~/.cache/yarn`, rather than `node_modules`. The performance gain from caching `node_modules` is not significant on either platform, and the complexity is significantly increased, due to us having to cache all the node_modules folders for all packages

jnicklas:
Even with these tweaks, windows tests are still by far the slowest. If we were to run linux tests only it would cut the time down from 10m to 5m.

cowboyd (approved):
We're just testing the CLI in Windows, which is A-OK with me 👍🏻

PR #807, opened 2021-02-01
fix @bigtest/eslint-plugin exports

dagda1:
def

I totally got the conventions of exporting the `@bigtest/eslint-plugin` totally wrong.

I created a test project and `eslint` could not find any of the rules

For example, we have a shared base config 

```ts
export = {
  parserOptions: { sourceType: 'module' },
  plugins: ['@bigtest'],
  rules: {
    '@bigtest/require-default-export': 'error'
  }
}
```

We have to use `export = {` rather than `export defult {`.

I finally got to the bottom of it after a lot of `yarn link`ing and profanity.

@cowboyd  We could generate an `eslint.config` as part of the `yarn bigtest init`?


cowboyd:
> @cowboyd We could generate an eslint.config as part of the yarn bigtest init?

@dagda1 that is a great idea, and I think we should definitely do this.

dagda1:
> > @cowboyd We could generate an eslint.config as part of the yarn bigtest init?
> 
> @dagda1 that is a great idea, and I think we should definitely do this.

As part of this PR or a separate one?

cowboyd (approved):

taras (approved):
👍

PR #808, opened 2021-02-01
ensure bundler errors go right up the stack

### LINKED ISSUE #801: malformed bundle is difficult to diagnose

dagda1:
Fixes #801 

This is to ensure bundler errors get raised right to the cli and we give the user a chance to create a github issue:

error

The original issue was caused by a user with a test file that did not have a default export.  

This fix still does not let the user know what the error was.

As I may have mentioned a few times :), I am not a fan of this default export requirement.

But as it stands we are not doing any runtime check.

We could potentially check in the manifest-generator that the files have a default export but doing this on a file by file basis is not great.

### DIFF
--- a/.changeset/twenty-wombats-switch.md
+++ b/.changeset/twenty-wombats-switch.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/server": patch
+---
+
+ensure bundler errors go right up the stack
--- a/packages/server/src/manifest-builder.ts
+++ b/packages/server/src/manifest-builder.ts
@@ -128,7 +128,6 @@ export function* createManifestBuilder(options: ManifestBuilderOptions): Operati
         } catch(error) {
           console.debug("[manifest builder] error loading manifest");
           bundlerSlice.update(() => ({ type: 'ERRORED', error }));
-          throw error;
         }
 
         break;
--- a/packages/server/src/manifest-builder.ts
+++ b/packages/server/src/manifest-builder.ts
@@ -128,6 +128,7 @@ export function* createManifestBuilder(options: ManifestBuilderOptions): Operati
         } catch(error) {
           console.debug("[manifest builder] error loading manifest");
           bundlerSlice.update(() => ({ type: 'ERRORED', error }));
+          throw error;
         }
 
         break;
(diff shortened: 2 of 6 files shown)


cowboyd (commented):
Do we want to kill the CLI if there is a bundler error? or is this just in the event that we can't actually link and load the bundle. It seems to me there are two things:

1. The CI command should definitely fail
2. the `server` should continue running, but it might make sense to crash if a certain class of errors (what amounts to linkage errors) percolate upwards.

dagda1:
> Do we want to kill the CLI if there is a bundler error? or is this just in the event that we can't actually link and load the bundle. It seems to me there are two things:
> 
> 1. The CI command should definitely fail
> 2. the `server` should continue running, but it might make sense to crash if a certain class of errors (what amounts to linkage errors) percolate upwards.

Good point, I think we have different types of exceptions.

I think we can say that a test that looks like this:

```ts
{
  test: {
    t: TestBuilder {
      description: 'No default export',
      steps: [],
      assertions: [],
      children: [],
      state: 'step'
    },
    fileName: 'manifest-54f435368391c9a6f8b73ea545058998181e4f5f0c28e868b684878125ca0ab8.js'
  }
}
```

A `test` that has no default export, i.e. none of the keys we expect (`description`, `steps`) etc. then we don't crash the CI.

For unhandled or unexpected exceptions, we crash the CI.

cowboyd:
How do we differentiate the two?

dagda1:
> How do we differentiate the two?

By the type of exception.  As we know, there are no multiple catch blocks in javascript so we could use a type guard or something like:

```ts
const isUnandledException(err: any) => /// using some tag on our custom Errors 
```

and then

```ts
        } catch(error) {
          console.debug("[manifest builder] error loading manifest");
          bundlerSlice.update(() => ({ type: 'ERRORED', error }));
          
          if (isUnandledException (error)) {
            return;
         }
          throw error;
```

dagda1:
closed in favour of #809

PR #809, opened 2021-02-02
Validate test implementation in validate-test

dagda1:
In response to #801 where a malformed test file resulted in a meaningless error message.

This PR verifies (in `validate-test`) that a test object must have

- At the very least a description filed
- assertions and children fields.

Descriptive errors are thrown if either of the above is not true.


packages/server/test/manifest-builder.test.ts
```diff
@@ -177,9 +177,37 @@ describe('manifest builder', () => {
...
+    });
+
+    it('should update the global state with the error detail', () => {
+      let bundlerState = atom.get().bundler;
+
+      assertBundlerState(bundlerState.type, {is: 'ERRORED'})
+
+      expect(bundlerState.error.message).toContain('Test contains no assertions or children')
+    });
+  });
```
jnicklas: There are tests specific for `validateTest` in the suite package here: https://github.com/thefrontside/bigtest/blob/8ffcba281845b2f7bc0334c40718f53a68386b1b/packages/suite/test/validate-test.test.ts

I think it would make sense to add those tests there?

dagda1: @jnicklas I agree, that is a better place.

packages/suite/src/validate-test.ts
```diff
@@ -19,6 +19,15 @@ export class TestValidationError extends Error {
...
 
+const every = Array.prototype.every;
+const some = Array.prototype.some;
+
+function validateTestKeys (test: Test, keys: (keyof Test)[], validationFn: typeof some | typeof every): boolean {
+  // the disable comment below is because eslint is not recognising k as used in !!test?.[k].
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  return validationFn.call(keys, (k: keyof Test) => !!test?.[k]);
+}
+
```
jnicklas: IMO this is a bit too clever. I've stared at this function for ten minutes and I'm still not sure what it does 😂

An alternative suggestion: how about changing the signature of `validateTest` to take an `unknown` instead, this actually meshes with how it is used, since we want to validate completely arbitrary input, we could then do something like:


``` typescript
// runtime validation and casting of the `Test` type
function validateTestType(test: any): Test {
  if(!test) {
     throw new TestValidationError("Invalid Test: the test is empty, did you export anything?")
  }
  if(!test.description) {
     throw new TestValidationError("Invalid Test: the test must have a description")
  }
  // ...
  return test as Test;
}

function validateTest(input: unknown) {
  let test = validateTestType(input);
  // ...
}
```

dagda1: > IMO this is a bit too clever. I've stared at this function for ten minutes and I'm still not sure what it does 😂
> 

haha, I'm not sure `clever` is the right word here if nobody gets what it does.   I think you have a point

You could have `Partial<Test>` rather than unknown:

```ts
function validateTestType(test?: Partial<Test>): Test {
```

jnicklas: I think that would be correct at runtime, but might be slightly less accurate from a typing perspective, since we could in theory export something like a number from a test file, and then we would end up calling this with a number, which I think should work at runtime, but would technically be not allowed by the TS typesystem.

dagda1: > since we could in theory export something like a number from a test file

Exporting a number would be incorrect though?

I would say we are stating here what the expected type is and not what could be passed.

Typescript has no runtime semantics

jnicklas: Yes, precisely. Since the input to this function is provided by the end user via whatever they export from their testfiles, it could be literally anything, even something silly like a number. The `validateTestType` function I sketched out above is a runtime implementation of type checking. It will check all properties of the type and only if they are satisfied will it return a properly typecast value.

dagda1: > Yes, precisely. Since the input to this function is provided by the end user via whatever they export from their testfiles, it could be literally anything, even something silly like a number. The `validateTestType` function I sketched out above is a runtime implementation of type checking. It will check all properties of the type and only if they are satisfied will it return a properly typecast value.

Makes sense, then what you are suggesting might be better as a type guard:

```ts
function validateTestType(test?: any): test is Test {
```

jnicklas: Yes, that totally works too, might be tricky though since it needs to throw in case of error, not return false.

packages/suite/src/validate-test.ts
```diff
@@ -11,14 +11,31 @@ export class TestValidationError extends Error {
...
     if(file) {
       this.loc = { file }
     }
   }
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function ensureIsTest (test: any, path: string[] = [], file?: string): test is Test {
+  if(!test) {
+    throw new TestValidationError("contains no description.\n\nDoes the test file contain a default export?", path, file)
```
jnicklas: I think this error message should not mention `description`?

packages/suite/src/validate-test.ts
```diff
@@ -11,14 +11,31 @@ export class TestValidationError extends Error {
...
+  }
+ 
+  if(!test.description) {
+    throw new TestValidationError("contains no description.\n\nDoes the test file contain a default export?", path, file)
+  }
+
+  if(!test?.children || !test?.assertions) {
+    throw new TestValidationError('contains no assertions or children.');
```
jnicklas: Can we have these as separate, more specific errors?

jnicklas (approved):

jnicklas:
The CI build should be fixed now, could you rebase this?

dagda1:
> The CI build should be fixed now, could you rebase this?

rebased and gone!

Issue #785, opened 2021-01-21
Replace todomvc with the sample app


minkimcello:
Currently, `cli`, `server`, and `integrations/cypress` relies on `@bigtest/todomvc` for its tests and we need to replace it with the BigTest sample app.

PR #825, opened 2021-02-04
Remove todomvc

jnicklas:
See #785

The todomvc app is not a good example on how to use bigtest and it is not neceassary anymore. We should get rid of it from our codebase.


jnicklas:
@minkimcello I'm not sure what the best way of replacing it is for the cypress tests

minkimcello:
Should we actually use the sample app to test cypress? I think we'll need actual interactions in the app to test for cypress so maybe that's the best way to go.

I can take care of the cypress test and remove todomvc package in a separate PR if you want to merge this one first.

jnicklas:
> I can take care of the cypress test and remove todomvc package in a separate PR if you want to merge this one first.

That sounds great!

cowboyd (approved):

PR #826, opened 2021-02-04
Don't wait for manifest builder to start orchestrator

jnicklas:
## Motivation

Currently we block the orchestrator readiness until the manifest builder has completed. This is actually unnecessary, since each individual test run also waits for this to complete. The orchestrator is perfectly usable before this, and doing this will allows us to use the orchestrator before the manifest is ready.

## Approach

Simply skip the step

There is a slight downside in that requesting a listing of the manifest might return a blank value. However with the changes outlines in #593, we will no longer provide this option anyway.

### DIFF
--- a/.changeset/large-horses-jog.md
+++ b/.changeset/large-horses-jog.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/server": minor
+"bigtest": minor
+---
+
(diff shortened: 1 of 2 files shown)


cowboyd (approved):
I'm all for this 👍🏻

PR #828, opened 2021-02-04
Remove integrations workspace and update scripts

### LINKED ISSUE #814: Implementing integrations for CI without mudding up the lock file
I remember now why I piggybacked integrations' `install` and `prepack` scripts to the main monorepo. It was for the preview and release workflows. Those workflows only run `yarn install` and `yarn prepack` at the root and do not have options to run specific commands inside the individual directories of packages.

We're trying to avoid putting it into the current workspace to separate the dependencies from `yarn.lock`.

minkimcello:
## Motivation

I originally had the install, prepack, and lint scripts piggybacking on the main workspace for the preview and release workflows (which was then removed in #806 when we were setting up CircleCI). We can't include integrations in the workspace configuration because it would combine the yarn.lock file.

Closes #814

## Approach

- Removed the workspace config _inside_ `integrations/`
- Updated the integration scripts at the root of the monorepo to specify cypress directory
  - `install:integrations`: `yarn --cwd integrations install` => `yarn --cwd integrations/cypress install`
    - once we have more integrations, we can combine the scripts together
- Added `lint` in the circleci workflow
- Added `prepack:all` script at root which runs `yarn workspaces prepack && install:integrations && prepack:integrations`
  - Running this command in `before_all` for both preview and release workflows


cowboyd (approved):
Looks good to me 👍🏻

PR #830, opened 2021-02-04
Add grammar guidelines in interactor docs for is and has

minkimcello:
## Motivation

To follow up on [this](https://github.com/thefrontside/bigtest/pull/798#discussion_r566732555) comment. In explaining the assertion methods of `is()` and `has()` in our Interactor docs, we say that the difference is for semantics but this may be too vague.

## Approach

I added another example right below the previous explanation to suggest using `is()` for adjective assertions and `has()` for nouns.

Screen Shot 2021-02-04 at 1 55 40 PM


jorgelainfiesta (approved):
Sweet!

PR #832, opened 2021-02-05
Run CLI tests on macOS on CircleCI

jnicklas:
Windows and MacOS are really our *main* platform, since most developers are not working on Linux desktops (sidenote, is 2021 the year of Linux on desktop? :trollface:). So we should run our tests on macOS in case there are any differences.

### DIFF
--- a/.changeset/gentle-gorillas-compare.md
+++ b/.changeset/gentle-gorillas-compare.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/webdriver": patch
+"bigtest": patch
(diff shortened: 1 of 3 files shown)


jnicklas:
This fails due to safaridriver not being enabled. Chrome is not installed on the VMs, and installing it via homebrew is very slow, so using safari would be much better.

jnicklas:
👊 👊 👊 👊

cowboyd (approved):

PR #835, opened 2021-02-05
Allow CLI tests to switch driver

jnicklas:
Currently, we're always running the CLI tests against chrome, this allows the tests to run against other drivers too.


cowboyd (approved):
Makes total sense

PR #836, opened 2021-02-06
Remove todomvc (part 2)

### LINKED ISSUE #785: Replace todomvc with the sample app

minkimcello:
## Motivation

Act II of removing `packages/todomvc` from the monorepo. Follow up to #825 and resolves #785.

## Approach

- Deleted `packages/todomvc`
- Removed todomvc from circleci test matrix
- Removed todomvc from monorepo and tsconfig settings
  - Noticed `packages/convergence` was in list of packages in `package.json` so I went ahead and removed that too
- Updated `integrations/cypress` test to use sample app instead
- `packages/bundler` was missing a `@types/babel__core` dependency
  - Error only appeared after todomvc was removed
- Didn't add changeset for 

## TODOs

- [x] Deprecate todomvc from npm?
- [x] Add changesets for `integrations/cypress` and or `packages/bundler`?


cowboyd (approved):
- I don't think we need to deprecate the npm package. Who knows, we may want to use it in the future
- I don't think we need a changeset for this because the only thing that changed was a _development_ dependency, not anything that would affect the content of the npm tarball

PR #842, opened 2021-02-08
Increase timeout of manifest builder tests

jnicklas:
These tests were randomly failing, presumably due to the timeout being too low.


cowboyd (approved):

taras (commented):
👍 not a very scientific way to do this but better than a failure

Issue #847, opened 2021-02-09
Add `className` and `classList` filters


cowboyd:
Especially when writing a test suite for components, you want to be able to make certain assertions based on the class name (`string`)or the class list (`string[]`)

```js
// when the button is being hovered.
await Button("Submit").has({ className: including('active') });
// alternatively
await Button("Submit").has({ classList: some(including('active')) });
```

PR #848, opened 2021-02-09
Deprecated focus/blur/focused helpers

jnicklas:
Just a little cleanup. It doesn't feel necessary to export these anymore, imo. Inline them into the HTML interactor instead. They can be reused by extending this interactor instead.

### DIFF
--- a/.changeset/forty-masks-teach.md
+++ b/.changeset/forty-masks-teach.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/interactor": minor
+"bigtest": minor
+---
+
+Deprecate focus, blur and focused helpers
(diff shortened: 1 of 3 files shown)


packages/interactor/src/index.ts
```diff
@@ -5,7 +5,6 @@ export { Page } from './page';
 export { App } from './app';
 export { perform } from './perform';
 export { fillIn } from './fill-in';
-export { focused, focus, blur } from './focused';
```
cowboyd: Can we deprecate these first instead of removing them? I say this because we're still in the process of upgrading the FOLIO interactors and the backwards compatibility is actually really nice since we're able to clearly say what is what. If we hold off breaking, then we can continue to enhance the test suite with features like #847 and consume new versions, while still not having everything converted to the new api.

cowboyd (approved):

PR #849, opened 2021-02-09
Improve HTML interactor

### LINKED ISSUE #847: Add `className` and `classList` filters

jnicklas:
Adds `text`, `className` and `classList` filters, also adds a selector which matches any element.

Closes #847

### DIFF
--- a/.changeset/honest-planes-notice.md
+++ b/.changeset/honest-planes-notice.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/interactor": minor
+"bigtest": minor
+---
+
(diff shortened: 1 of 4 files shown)


cowboyd (approved):
🎉 🥳

PR #852, opened 2021-02-10
Remove typings from createInteractor() examples

### LINKED ISSUE #823: Documentation - Show both JavaScript and TypeScript in JS examples
Some of the examples present in the Interactors docs do not work in the sample app because it does not use TypeScript. For example:

```
import { Button, createInteractor } from 'bigtest';

export const Form = createInteractor<HTMLFormElement>('form')({
  selector: 'form',
  ....
```

A  new learner has to modify the code before it can be used. Since BigTest provides full TypeScript support, we should show TypeScript by default, but also have a tab for JavaScript.

@cowboyd @minkimcello what do you think of this idea?

minkimcello:
## Motivation

To resolve #823.

## Approach

The were only two examples that had typings and they were both in the `writing-interactors` page so I just removed those.

Per the discussion in the issue mentioned above, the conclusion, I believe, is people who use Typescript will know how to type HTML elements and non-Typescript users might be thrown off by the syntax so we might as well just only show Javascript.

### DIFF
--- a/website/docs/interactors/4-write-your-own.md
+++ b/website/docs/interactors/4-write-your-own.md
@@ -23,7 +23,7 @@ import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
 <Tabs
-  defaultValue="npm"
+  defaultValue="javascript"
   values={[
     {label: 'Javascript', value: 'javascript'},
     {label: 'Typescript', value: 'typescript'}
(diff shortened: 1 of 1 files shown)


cowboyd (approved):

jnicklas:
One downside of this: I don't think it's entirely obvious how you would get this to typecheck correctly when using TypeScript. Most of the time, type parameters do not need to be given explicitly, but in this particular case they actually do. This is not common, and so it wouldn't be easy to guess that this is what you need to do.

Personally I'd go with different code samples for JS and TypeScript here.

minkimcello:
@jnicklas @cowboyd 
Screen Shot 2021-02-10 at 9 47 42 AM

PR #858, opened 2021-02-11
Use innerText over textContent where appropriate

### LINKED ISSUE #853: Use `innerText` instead of `textContent`
In our interactors we currently use `textContent`, but `innerText` would be a better choice. The differences between the two are explainer [here](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext).

jnicklas:
Uses `innerText` for built in HTML interactors. Still default to `textContent` for other elements.

Closes #853

### DIFF
--- a/website/docs/interactors/4-write-your-own.md
+++ b/website/docs/interactors/4-write-your-own.md
@@ -81,7 +81,7 @@ NoSuchElementError: did not find my-textfield-interactor "USERNAME"
 ```
 _An example of the console output when a test is unable to locate the interactor_
 
-Locators, filters, and actions are optional when creating your own interactor. While the locator for the `TextField` interactor offered by BigTest uses the `textContent` of the associated label (as we explained on the [Locators, Filters, and Actions page](/docs/interactors/locators-filters-actions#filters)), the example above has its locator configured as `element.placeholder`. This is just to demonstrate that you can set the properties of Locators to anything that suits your needs. If you create an interactor without a locator, it will default to `locator: element => element.textContent`.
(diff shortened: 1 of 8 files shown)


website/docs/interactors/4-write-your-own.md
```diff
@@ -81,7 +81,7 @@ NoSuchElementError: did not find my-textfield-interactor "USERNAME"
...
 
-Locators, filters, and actions are optional when creating your own interactor. While the locator for the `TextField` interactor offered by BigTest uses the `textContent` of the associated label (as we explained on the [Locators, Filters, and Actions page](/docs/interactors/locators-filters-actions#filters)), the example above has its locator configured as `element.placeholder`. This is just to demonstrate that you can set the properties of Locators to anything that suits your needs. If you create an interactor without a locator, it will default to `locator: element => element.textContent`.
+Locators, filters, and actions are optional when creating your own interactor. While the locator for the `TextField` interactor offered by BigTest uses the `innerText` of the associated label (as we explained on the [Locators, Filters, and Actions page](/docs/interactors/locators-filters-actions#filters)), the example above has its locator configured as `element.placeholder`. This is just to demonstrate that you can set the properties of Locators to anything that suits your needs. If you create an interactor without a locator, it will default to `locator: element => element.textContent`.
```
cowboyd, line 84:
```suggestion
Locators, filters, and actions are optional when creating your own interactor. While the locator for the `TextField` interactor offered by BigTest uses the `innerText` of the associated label (as we explained on the [Locators, Filters, and Actions page](/docs/interactors/locators-filters-actions#filters)), the example above has its locator configured as `element.placeholder`. This is just to demonstrate that you can set the properties of Locators to anything that suits your needs. If you create an interactor without a locator, it will default to `locator: element => element.innertText`.
```

cowboyd (commented):
I think you might have missed one.

cowboyd (approved):

PR #859, opened 2021-02-12
WIP Accessibility testing guide

jenweber:
This is in rough-draft state.

## Motivation

BigTest is really well suited to doing accessibility testing! We should teach it.

## Approach

This new section has two goals:

1. Show people how to do accessibility testing with BigTest/Interactors
2. Add to the body of working examples. The same patterns/principles are reusable for testing other things besides accessibility.

### Alternate Designs

This guide could be split up into multiple pages, or examples could go in a cookbook along with other examples. I am thinking that we start small and see where user needs/feedback push us towards.

I was unsure whether this belongs with Interactors or BigTest. I'm curious what the benefits/drawbacks are of both locations.

### TODOs and Open Questions

- [ ] Fill in code examples
- [ ] Determine if this coverage is MVP or if we want to add more topics before shipping. More possible topics include developing and documenting keyboard testing, an `enterText` action for inputs, an interactor for checking which element has focus, and running axe core
- [ ] determine if this is the right placement in the sidebar/site
- [ ] Editorial review
## Learning

This is greatly inspired by https://guides.emberjs.com/release/accessibility/ and https://reactjs.org/docs/accessibility.html

### DIFF
--- a/website/docs/platform/5-accessibility.md
+++ b/website/docs/platform/5-accessibility.md
@@ -1,5 +1,5 @@
 ---
-id: architecture
+id: accessibility
 title: Accessibility testing
 ---
 
--- a/website/docs/platform/5-accessibility.md
+++ b/website/docs/platform/5-accessibility.md
@@ -12,10 +12,18 @@ By the end of this guide, you will learn how to test some common accessibility f
 
 ## Inputs and labels
 
-[Every input in your app needs an associated label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#labels). One way you can ensure this is to use labels as the main way to locate inputs in your tests:
+[Every input in your app needs an associated label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#labels). One way you can ensure this is to use labels as the main way to locate inputs in your tests, instead of looking up an input by `id`.
+The following test would fail if the input was not connected to its label.
 
 ```js
-TODO
+export default test('Email input')
+  .step(Page.visit('/'))
+  .step(TextField('Email').fillIn('something@example.com'));
+```
+
+```html
+<label for="input-email">Email</label>
+<input type="text" id="input-email" />
 ```
 
 If you have a hard time finding an input with your interactors, you may have caught an accessibility bug!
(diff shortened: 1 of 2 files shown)


jenweber:
@cowboyd if you could give this a read through, the feedback I'm especially looking for at this stage is whether I have hit all the main selling points of BigTest. Let me know if you see gaps! Does this look like the right scope for a shippable MVP? Do you have thoughts on whether this belongs under Platform or Interactors?

Also, is there anyone in the BigTest community who we can interview to see what kinds of things they wish they could test for, and why?

jenweber:
Hi @cowboyd, this is ready for another look. It has code samples added. I ran into some issues with dependency versions on the sample app (not being able to use the new interactors syntax) and also the typescript issue. I'll open a PR to the sample app to fix this.

website/docs/platform/5-accessibility.md
```diff
@@ -0,0 +1,165 @@
...
+Write an interactor that looks for the buttons that control toggling the open
+and close of a menu:
+
+```js
+export const Toggle = createInteractor('toggle')
+  .selector('button')
+  .locator((element) => element.innerHTML)
+  .filters({
+    ariaExpanded: (element) => element.ariaExpanded
+  })
```
cowboyd, line 119: It will probably be easier to work with this filter if we convert it into a first-class boolean value
```suggestion
  .filters({
    ariaExpanded: (element) => element.ariaExpanded === "true"
  })
```

website/docs/platform/5-accessibility.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+Write an assertion using the `ariaExpanded` filter:
+
+```js
+export default test('Toggle to expand')
+  .step(Page.visit('/'))
+  .assertion(
+    Toggle('Expand', { ariaExpanded: "false" }).exists()
+  )
```
cowboyd, line 129: I feel like this assertion reads _slightly_ better using `is()`
```suggestion
  .assertion(Toggle('Expand').is({ ariaExpanded: true }));
```

But it's subjective.

cowboyd (commented):
@jenweber I gave this another read through and I think it's great! I rebased it on `v0` to remove the conflicts and get the tests passing again.

As for the content, my biggest questions are (some of which overlap with your TODO checklist)

- Where should this guide reside? In its current form, it's part of the platform docs. On the one hand, this makes sense because the code samples are using the bigtest platform, and this is something we want to tout universally, but fundamentally, it is the interactors that do the heavy-lifting when it comes to accessibility, and so part of me wonders if this should be attached to the interactors, and not the runner. This represents more work, because we'll need to make the code samples in both Jest and Cypress as well, but ultimately I think it will serve the marketing message better to since you don't need to swallow the whole whale in order to improve your accessibility coverage _right _now.

Finally, the question is, should we go with this before we have our keyboard issues resolved?

Issue #855, opened 2021-02-10
Website newsletter sign up form padding


minkimcello:
In both desktop and mobile format, there seems to be padding missing from the newsletter sign up form:

## Desktop
![Screen Shot 2021-02-10 at 10 30 49 AM](https://user-images.githubusercontent.com/29791650/107532353-d306da80-6b8b-11eb-9489-ecbc1e79f3b1.png)
- Join our newsletter image is cropped
- I think this section needs a bit more padding on the bottom so that it's more or less proportional to the rest of the page

## Mobile
![Screen Shot 2021-02-10 at 10 31 14 AM](https://user-images.githubusercontent.com/29791650/107532363-d4380780-6b8b-11eb-8a7a-169aa3676ca3.png)
- Missing left/right padding

jorgelainfiesta:
Fixed on #861

PR #861, opened 2021-02-12
Fix newsletter padding

jorgelainfiesta:
## Motivation

#855 

## Approach

Updated the CSS to have the newsletter form respond more smoothly to the window size.

Screenshot 2021-02-12 at 11 03 17


minkimcello:
Screen Shot 2021-02-12 at 10 37 10 AM

The picture still gets cropped in full desktop mode and we're still missing the bottom padding.

And I believe you need to rebase because you're introducing todomvc package.json back

jorgelainfiesta:
Thanks Min! should be good now

minkimcello:
@jorgelainfiesta looks good. just _one_ more in the `about` page. i think the body needs a bottom padding/margin?

Screen Shot 2021-02-13 at 10 05 37 AM

versus

Screen Shot 2021-02-13 at 10 05 22 AM

jorgelainfiesta:
Umm interesting, there's probably a break point that needs adjustment. I'll check it out, thanks.

Update: no, the problem is that that component is set within another component so the breakpoint is not working because it targets the window width... need some refactoring

minkimcello (approved):

PR #863, opened 2021-02-16
Change format of filter not matching error

jnicklas:
This changes the format of the `FilterNotMatchingError`. The table format has been pointed out as confusing by users. Also, both `ElementNotFoundError` and `FilterNorMatchingError` currently look visually similar, but serve different purposes. By making them more visually distinct, we make it easier to see these as separate errors with different failure modes.

Before:

Screenshot 2021-02-16 at 16 42 12

After:

Screenshot 2021-02-16 at 16 41 53

In this PR, colorizing the output is not included, since it requires more substantive changes, and this way we can split up this complex change into smaller chunks.

### DIFF
--- a/.changeset/fresh-bears-boil.md
+++ b/.changeset/fresh-bears-boil.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/interactor": patch
+"bigtest": patch
+---
+
+Improved format for filter not matching error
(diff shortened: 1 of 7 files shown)


cowboyd (approved):

cowboyd:
Looks good.

PR #864, opened 2021-02-16
Track orchestrator status in state

jnicklas:
Tracks the orchestrator status in the state instead of via a delegate. Removes the last of the delegates.


cowboyd (approved):
🔪 🔪  ✂️

PR #866, opened 2021-02-16
feat: website sync up code blocks

jbolda:
## Motivation

This change will sync up all code blocks. If someone is using, say `yarn`, they will likely prefer to see all of the code blocks using yarn instead of npm. This change allows selecting it once, and every other instance will also be changed as well (that matches the `groupId`).

## Approach

Add `groupId="text"` to any `Tabs`, this will sync up tabs that have that same `groupId`.


jorgelainfiesta (approved):
Wow I love it!

minkimcello (approved):

PR #868, opened 2021-02-18
Make interactors work out of the box in JSDOM

cowboyd:
Motivation
--------------
When we switched to `innerText` as the default locator, we actually broke the test suite because JSDOM does not actually implement `innerText` at all. To make up for this, we monkey-patched `innerText` in our test suite to fallback to `textContent`. However, this had one serious drawback. We want interactors to work out of the box with JSDOM, so every test suite using JSDOM as the document ended up breaking. So while we fixed our own test suite, we broke everybody else's that uses JSDOM🤦🏻‍♂️

Approach
----------
This change makes the `text` filter and locators that rely on `innerText` handle gracefully the case where the `innerText` property does not actually exist, using `textContent` and finally the empty string as a fallback. In practice this should be very rare with the only known exception being JSDOM.

### DIFF
--- a/packages/interactor/src/definitions/html.ts
+++ b/packages/interactor/src/definitions/html.ts
@@ -3,9 +3,9 @@ import { isVisible } from 'element-is-visible';
 
 const HTMLInteractor = createInteractor<HTMLElement>('element')
   .selector('*')
-  .locator((element) => element.innerText)
+  .locator(innerText)
   .filters({
-    text: (element) => element.innerText,
+    text: innerText,
     title: (element) => element.title,
     id: (element) => element.id,
     visible: { apply: isVisible, default: true },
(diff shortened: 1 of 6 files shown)


packages/interactor/test/create-interactor.test.ts
```diff
@@ -24,7 +24,7 @@ const Div = createInteractor('div')
 
 const Details = createInteractor<HTMLDetailsElement>('details')
   .selector('details')
-  .locator((element) => element.querySelector('summary')?.innerText || '')
+  .locator((element) => element.querySelector('summary')?.textContent || '')
```
cowboyd, line 27: In cases like these, I've just opted to use `textContent` directly since that way we don't have to couple anything more to the `innerText` helper, and because these interactors only exist inside our test suite which is known to be JSDOM

packages/interactor/src/definitions/html.ts
```diff
@@ -19,6 +19,16 @@ const HTMLInteractor = createInteractor<HTMLElement>('element')
...
   })
 
+// Because JSDOM does not have any concept of flow or layout, it cannot actually implement
+// `innerText` correctly, so they have opted not to implement it at all. So, to make things
+// work use `textContent` as a backup for all `HTMLElement` subtypes if `innerText` is
+// undefined
+//
+// See details: https://github.com/jsdom/jsdom/issues/1245
+export function innerText(element?: HTMLElement): string {
+  return (element?.innerText != null ? element?.innerText :  element?.textContent) || '';
```
jnicklas, line 29: I suppose this works because `innerText` will always return a string, right? So if it is `undefined` that means it actually isn't implemented?

jnicklas (approved):

packages/interactor/src/definitions/html.ts
cowboyd: Exactly!

Issue #625, opened 2020-10-15
pause() step


taras:
When debugging, it's useful to be able to stop at a specific step to see what's happening on the screen. I've been using

```
.step('stop here', () => new Promise(() => {}))
```

It would be nice if this was built in.

Thoughts?

cowboyd:
What about a `debug()` action that pauses in the debugger. The only problem with the pause is that your timeout is still running.

Issue #415, opened 2020-07-20
add `--no-exit` flag to CI command for debugging purposes


cowboyd:
When things don't work quite as expected in CI mode, you want to leave the whole things running so that you can inspect the browser developer console and environment. However, when bigtest exits it shuts down everything very efficiently, so it can be difficult to understand what exactly happens.

I propose adding a `--no-exit` flag to just leave everything on after a CI run.

PR #870, opened 2021-02-18
[WIP] Agent toolbar for `debug()` and `--no-exit`

minkimcello:
WIP #415 #625

## TODOs
- [ ] Get a button in the agent toolbar to invoke a function from `run-lane.ts`
  - This is so that later we can have a `step` and `play` button
- [ ] call `debug()` from test
    ```js
    import { debug } from 'bigtest';
    test => test .step().step(debug())
    ```
  - It's instantiated in `run-lane.ts` so it should expose it from there? maybe we need to go up higher to pass a callback function
- [ ] be able to control timebox
  - first console log out the value so that we can capture the value for later
  - be able to halt/pause the timebox for both `debug()` and `--no-exit`
    - i think the timebox already has `halt` and `resume` implemented
- [ ] `--no-exit` for `bigtest ci`
  - halt at the end of the run-lane
    - need to halt the timebox
    - add flag option within cli
- [ ] When all done I should add it to the bigtest runner docs probably?


minkimcello:
@cowboyd

So at the moment the agent of this PR looks something like this:
```html
<html>
  <head>...</head>
  <body>
    <AgentToolBar/> <!-- i added -->
    <div id="body">
      <splashimage/>
      <iframe id="test-frame">
        #document
          <html>
            <head>...</head>
            <body>
              <iframe id="app-frame">
                #document
                  <SampleApp/>
              </iframe>
              <script src="test-frame-main.ts"></script>
            </body>
          </html>
      </iframe>
      <script src="./main.ts"></script>
    </div>
  </body>
</html>
```

I would like to subscribe to 
```
document.querySelector('iframe[id=test-frame]').contentDocument
  .querySelector('iframe[id=app-frame]').contentWindow.location.href
```
from within `<AgentToolBar/>` but I'm not sure how I should go about it.

I first fiddled with `useEffect()` and its second argument but that was no use because we're trying to observe elements outside its component. 

Then I messed around with `MutationObserver` in hopes of maybe detecting for DOM changes in the individual iframe so that we can detect when the app inside the app-frame is loaded and I could tell the toolbar to re-evaluate the `location.href` but that I wasn't able to get too far with that approach.

taras:
@minkimcello I think we should just make it default to whatever the OS is set to.

cowboyd:
> now takes a third optional argument for the version of BigTest the server is running from.

Rather than take the version from the url string, we should add this information to the communication protocol. That way, we can easily add more data to transmit to the agent later without having to extend the url mechanism with more and more parameters. Also, if we use the connection between orchestrator and agent, we can transmit information at any point during the lifetime of the agent, not just at the moment of connection.

However, that means extending the protocol. Right now, the only message that the orchestrator sends to the Agent is `Run` which instructs it to run a test. https://github.com/thefrontside/bigtest/blob/v0/packages/agent/shared/protocol.ts#L90-L99

We could add a `Hello` command or a `ServerInfo` that transmits metadata about the bigtest server including the version number. 

>  my assumption is people who don't mind light theme won't mind dark theme

I prefer light theme in everything except my editors and terminals, and I experience mild annoyance when a non-development site is very dark, and while this is a development tool, so I'd probably be ok with it, the point is that there is no accounting for people's tastes, and so maybe we should develop this like any other webapp and take our cue from OS theme settings?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/bigtest">thefrontside/bigtest</a></strong></div><div>Tests that speed up development, not slow it down. Contribute to thefrontside/bigtest development by creating an account on GitHub.</div></blockquote>

PR #872, opened 2021-02-19
Use medium size windows executor

cowboyd:
## Motivation
We're paying $500 a month for our large windows executors.

## Approach
Switch to a medium VM since the pricing is half of the large and see what difference it makes in build time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,7 +164,7 @@ workflows:
       - test:
           executor:
             name: "win/default"
-            size: "large"
+            size: "medium"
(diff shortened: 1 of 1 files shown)


taras (approved):
👍

jnicklas:
$500 a month?!!??! What the hell?!

taras:
Windows large is very expense. It wasn’t $500. It’s about $300 but Windows large is 2/3 of that

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #879, opened 2021-03-02
Upgrade sample app with react router

minkimcello:
## Motivation
To split up #870 into smaller PRs.

I decided to extend the sample app by adding react router to it. One of the features I wanted to add to the `@bigtest/agent` is having the agent toolbar display the app's current address (e.g. `http://localhost:3000/aboutpage`). However, this benefit of being able to develop that feature is just a convenient side-effect. 

The actual focus of this PR is simply to enhance the sample app to add a bit more functionality to it because at the moment it just has one button that toggles back and forth between two states.

## Approach
- Added bigtest logo as favicon:
Screen Shot 2021-03-01 at 7 28 04 PM
- Two routes: `/` and `/about`. Each page just has a heading of `Home page` and `About page` respectively.
- New header bar with route links and the old sign in button that just toggles between `sign in` and `log out` (see pictures below).
- Updated the tests inside the sample app for all three testing platforms: `jest`, `cypress`, and `bigtest`. Previously the tests were only using the `Button` interactor but with these new changes we'll be using `Button`, `Link`, and `Heading` interactors in our sample tests.

Old sample app:
Screen Shot 2021-03-01 at 7 16 31 PM

New sample app:
Screen Shot 2021-03-01 at 7 16 25 PM


cowboyd (commented):
Does this affect the guides in any way?

minkimcello:
> Does this affect the guides in any way?

@cowboyd it does not. I've skimmed through the entire docs and it seems the only time we reference the sample app is when we say `In the sample project, when you click the “Sign In” button it disappears and a “Sign out” button appears in its place.` and the new sample app will still have that button.

sample/app/src/App.jsx
```diff
@@ -1,23 +1,55 @@
...
+              </button>
+            }
+            {isSignedIn &&
+              <button onClick={() => setIsSignedIn(false)}>
+                LOG OUT
+              </button>
+            }
+          </div>
+        </div>
+        <div className="body">
```
jenweber:
```suggestion
        <main className="body">
```

For semantics, I think this and the closing div tag could become `main`

sample/app/package-lock.json
```diff
@@ -7049,6 +7049,19 @@
       "resolved": "https://registry.npmjs.org/hex-color-regex/-/hex-color-regex-1.1.0.tgz",
       "integrity": "sha512-l9sfDFsuqtOqKDsQdqrMRk0U85RZc0RtOR9yPI7mRVOa4FsR/BVnZ0shmQRM96Ji99kYZP/7hn1cedc1+ApsTQ=="
     },
+    "history": {
```
jenweber, line 6861: Is it intentional that this has both a yarn lockfile and npm lockfile?

jenweber:
I like this idea! It's a nice basis for showing screenshots of things, and it adds a lot of polish. I ran this locally with Firefox and it looked good at desktop widths. However, I think it needs some minor mobile styling, if that's within scope for you.

sample/app/package-lock.json
minkimcello: Yeah. I'm not sure if it's a good idea or not but I thought it was necessary to offer both because we don't know if people prefer yarn or npm. Once we migrate over to npm completely, I can get rid of the yarn flag in the sample app command and the lock file as well.

minkimcello:
> I like this idea! It's a nice basis for showing screenshots of things, and it adds a lot of polish. I ran this locally with Firefox and it looked good at desktop widths. However, I think it needs some minor mobile styling, if that's within scope for you.

We'll scope that to a separate PR. Maybe we can tackle it when we start doing more work for bigtest mobile.

sample/app/src/App.jsx
minkimcello: 👍 I also replaced `<div className="header">` with `<header>`

pittst3r (approved):
:+1:

PR #881, opened 2021-03-03
Update copy to say 'Runner'

### LINKED ISSUE #880: Make it easier to find BigTest Architecture section
## Context

I'm talking to people about BigTest and the Architecture of BigTest is an important part of these conversations.
It would be helpful to me if it was easier to find the Architecture page from [BigTest website](https://frontside.com/bigtest/interactors).

## Possible Solution

Add a link to the Architecture page in the footer under the Platform column.

jorgelainfiesta:
## Motivation

Fixes #880 

## Approach

I realized that "platform" is used quite ambiguously across the site. For now I updated the copy to say Runner when we talk about the product that you use to write/run your tests on BigTest. Platform is reserved for talking about BigTest as a whole, encompassing Interactors, Runner, and Simulators.

## Next steps

Review copy more in depth and re-asses how we want to talk about things after recent developments


taras:
@jorgelainfiesta can you do a review of your changes to make sure that we didn't replace too aggressively? It's still helpful to talk about the platform but it's important to distinguish the runner from the platform. It might be as simple as changing Platform to Runner everywhere except one page where we have Platform Architecture. But we will need a page that talks about the Platform specifically cause there is the specific design that's very enabling that we need to talk about.

jorgelainfiesta:
I made no changes to the Architecture page. However it still a part of the Runner documentation. To my understanding, that's the architecture of the runner or not?

The only overview of the system/platform/project we have is in the /about page by Charles. We could have an Architecture page within the about page if that general architecture is available.

taras:
> To my understanding, that's the architecture of the runner or not?

No, runner is just a GraphQL client. Majority of the value is provided by the platform. These things are conflated because the orchestrator which is the brain of the platform is started by the runner when it's running locally.

jorgelainfiesta:
I think we need to rethink how we structure the pages, but as an intermediate solution I extracted the Architecture page and placed it within the about page. I added a section to the footer about the platform. What do you think?

taras (approved):
👍

jorgelainfiesta:
`test-agent`, `test-cli` and `test-webdriver` are not passing in this branch nor #884 although nothing in either PR seems to be related to those areas?

cowboyd:
@jorgelainfiesta you need to rebase onto `v0`, because of #308

Issue #882, opened 2021-03-03
CRASH: Cannot find module 'typescript'


zburke:
# Error Report

Following the instructions at https://frontside.com/bigtest/docs/platform: 
```
$ npm init
[...]
$ npm install bigtest --save-dev
$ npx bigtest init 
[...]
⌾ Do you want to write your tests in TypeScript? (yes/no)
   > no
$ cat bigtest.json 
{
  "port": 28000,
  "launch": [
    "default"
  ],
  "app": {
    "command": "npm start",
    "env": {
      "PORT": 28001
    },
    "url": "http://localhost:28001"
  },
  "testFiles": [
    "test/**/*.test.{ts,js}"
  ]
}
$ npx bigtest server --launch=chrome
[orchestrator] starting
😱😱😱 OH NO! UNEXPECTED ERROR! 😱😱😱
It looks like you've encountered a bug in BigTest that triggered an unexpected
shutdown. And yes, in case you're wondering, this is definitely on us. It would
help us a lot to improve BigTest if you'd take the time to report the problem.
[...]
```
This can be resolved by
```
npm i typescript --save-dev
```
but since I answered "no" to "Do you want to write your tests in TypeScript?" that seems odd. 

NB: The helpful-looking "Copy" button next to all the console instructions is slightly less helpful than it could be because it copies the leading `$ `, which you have to remove. 

<details><summary>diagnostics</summary>

Argv
----
server,--launch=chrome

Stack
-----
Error: Cannot find module 'typescript'
Require stack:
- /Users/zburke/temp/e2e/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js
- /Users/zburke/temp/e2e/node_modules/@bigtest/bundler/dist/bundler.js
- /Users/zburke/temp/e2e/node_modules/@bigtest/bundler/dist/index.js
- /Users/zburke/temp/e2e/node_modules/@bigtest/server/dist/src/manifest-builder.js
- /Users/zburke/temp/e2e/node_modules/@bigtest/server/dist/src/orchestrator.js
- /Users/zburke/temp/e2e/node_modules/@bigtest/server/dist/src/index.js
- /Users/zburke/temp/e2e/node_modules/@bigtest/cli/dist/start-server.js
- /Users/zburke/temp/e2e/node_modules/@bigtest/cli/dist/cli.js
- /Users/zburke/temp/e2e/node_modules/@bigtest/cli/dist/index.js
[...]

PR #884, opened 2021-03-04
Bugfix - sample app dependencies

jenweber:
Add typescript and pin bigtest version.

Partially addresses #882

## Motivation

I ran into two issues with a freshly generated sample app.

First, I got `selector is not a function` when using `createInteractor`. My sample app had bigtest `0.7.0` installed, which lacks the new syntax for interactors.

Second, I saw #882 after updating my bigtest dependency.

## Approach

- Pin the version of bigtest in app-pkg. An alternative would be to set it in the lockfiles only, but that seems easy to forget.
- Add typescript as a temporary dependency. This can be removed when #882 is resolved. This is a stopgap to make sure the main quickstart tutorial is back in working shape ASAP.

## After merging

This needs to be released after merging.

@minkimcello please feel free to push to this branch or create another PR if you have a different approach.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2006 @@
+{
+  "name": "bigtest",
+  "version": "0.0.0-monorepo",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/runtime": {
+      "version": "7.13.9",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.13.9.tgz",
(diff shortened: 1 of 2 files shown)


cowboyd:
It looks like they updated chrome version on the circle images. I'll need to push a fix, and then we'll have to rebase this.

minkimcello:
~I can't seem to replicate the typescript error on my end off of a fresh sample app install. 🤔~

cowboyd (approved):

PR #890, opened 2021-03-04
Introduce --only flag for framework-specific installs

minkimcello:
## Motivation

When there's a new release of a feature or if I'm trying to reproduce a bug, I'll often install a fresh new sample app and manually remove unnecessary jest/cypress files and dependencies to create a more tidy environment.

~~So this PR is to introduce a BigTest-only `-bo` flag to the `npx bigtest-sample` command so that it gives you a stripped down version without jest and cypress. And as an added bonus, with the `-bo` flag it appears the duration of the install gets shaved down around 15%.~~

This PR will introduce an `--only` flag with the options to specify `cypress`, `jest`, or `bigtest`.

To try out the `--only` flag, you can run `npx bigtest-sample@0.0.9-21505f00 --only [cypress|jest|bigtest]`.

## Approach ⭐ Updated ⭐
- Had to lock `jest-css-modules-transform` to `4.1.0` on account of the `parcel` bug.
- Added `dev:build` script for building into an ignored directory. Previously I had to keep modifying the target directory and manually deleting it to test out the build script.
- New migrate function:
  ```
  if(--only)
    merge(base-package-json, $framework-package-json)
    copy [...base-files, ...$framework-files] to target directory
    delete !framework test files by pattern
      // framework === cypress ?? delete tests that contain 'bigtest' and 'jest' in file names
  else
    merge(base, cypress, jest, bigtest)
    copy [...all-files] to target directory
  ```
  - To prevent the cypress template from installing bigtest (and webdriver) was to add bigtest into the bigtest and jest templates separately rather than the base template.
- Took the easy approach of updating the app's readme by deleting the section on how to run tests. Now the readme is applicable for all templates without requiring any edits.
- The success message at the end of `npx bigtest-sample` is dynamically generated based on template. If framework-specific then it only outputs the test command for that framework. If no `--only` flag is provided then it will output the commands of all three testing frameworks.
- Manually tested out all --only scenarios:

| | Success Message | test with npm | test with yarn | Error Message | All templates |
| :--- | :---: | :---: | :---: | :---: | :---: |
| `--only cypress` | ✅ | ✅ | ✅ |
| `--only jest`        | ✅ | ✅ | ✅ |
| `--only bigtest`  | ✅ | ✅ | ✅ |
| `--only foobar` | | | |  ✅ |
| No only flag | ✅ | ✅ | ✅ | | ✅ |

<details>
<summary>Details of original approach</summary>
<br>
- The new workflow of the `npx bigtest-sample` command looks something like this:

  ```
  create directory
  
  migrate files to the directory
  
  -bo flag ? remove all of the cypress/jest files and replace package.json
  
  install dependencies
  
  !(-bo flag) ? display instructional message
  ```
  - The new package.json takes `react`, `react-dom`, `parcel`, and `bigtest` version from the original package.json so we do not need to manually update those.
- I could've incorporated the `-bo` logic into the current `migrate` function but for legibility I decided it would be better if I refactored it into a separate function `bigtestOnly()` right after the file migration. This means we're taking the extra steps of removing files that we just moved into the new directory.
  - The alternative would be to wrap the if statement one level higher but then I'd be splitting off duplicate logic in the file migration into the `bigtestOnly()` function and that seems like it would make the code less robust?
- The `files` property of the root `package.json` has been set to `[*]` to include the `yarn.lock` files. I was unable to find why this is happening. The `npm` documentation does not make any mentions of it. It does say that `package-lock.json` is not included in the tarball when a package is published but ours has been getting published regardless. `npm publish` automatically ignores `node_modules` for us and the other directories we need to ignore would be `app/dist`, `app/.bigtest`, and `app/.cache` which I specify in the `.gitignore` nested inside `app/`.
- In addition to removing the jest/cypress files, I've also omitted the readme and the successful instructional output message in the console because the commands listed in the readme and the message will not be applicable to the stripped-down `-bo` version of the sample app.

</details>

## TODO

- [x] Merge #892 first to include the chromedriver fix and re-bump sample app version.
- [x] Update developer README (the one in `sample/` and not `sample/app/`) to instruct that new dependencies need to be added to the templates as well.
- [x] I have yet to update the README and the CLI success message. If someone goes by the README or the message output right now and tries to run `yarn test:cypress` from a `jest` template, they'll just get an error saying that command doesn't exist. So to tackle this issue, there are three options I can think of:
  - `Option 1`: Create four separate READMEs and CLI messages
  - `Option 2`: Generate the README and CLI success message based on the template name (`cypress`, `jest`, `bigtest`, or `undefined`)
  - `Option 3`: Keep all of the test scripts but have it output something along the lines of "Cypress is not available because this is a Jest template". This is probably the most complicated option of the three.


cowboyd:
For symmetry, should we have an `--only [bigtest|cypress|jest]` that lets us only focus on that framework?

jenweber (commented):
I am not sure how to test this locally. But I like the idea of adding a flag for this. It would be very helpful to me too. I do the same thing as you.

I like Charles' idea for the command syntax.

Also, nice branch name, haha.

minkimcello:
> I am not sure how to test this locally. But I like the idea of adding a flag for this. It would be very helpful to me too. I do the same thing as you.

@jenweber You can try it out with `npx bigtest-sample@0.0.8-f8334899 -bo`

You need to bump bigtest to 1.13.3 for the chromedriver fix, downgrade parcel to 1.12.3, and add typescript.

minkimcello:
> For symmetry, should we have an --only [bigtest|cypress|jest] that lets us only focus on that framework?

Would we need it? This `-bo` flag is really meant to be used for developers working _on_ bigtest.  It sort of acts as an interim playground until we have a real playground.

If we don't want to introduce a hidden feature to the `npx bigtest-sample` command, the alternative would be to add the flag one level higher in the `sample/` directory inside the `start-dev` script so that I can install/start the sample app within the bigtest repo without any of cypress or jest's dependencies. But this creates extra steps for the developers to have to undo the changes to the app's lockfiles.

And as you can see in the [`bigtest-only.js` file](https://github.com/thefrontside/bigtest/blob/mk/sample-body-odor/sample/bin/bigtest-only.js), the script isn't very future proof and requires upkeep whenever new dependencies are added. Writing scripts for `--only jest|cypress` would be even more complicated and be prone to breaking. While typing this out, I'm realizing maybe something this fragile shouldn't be added to the repository but instead just be a script I run locally for personal use.

cowboyd:
@minkimcello cypress and jest integration is definitely part of the bigtest story too, and so I was definitely imagining what would I do if I wanted to hack on or test the cypress integration. I see what you mean about not wanting to duplicate that script. My suggestion would be to use a template. For the package JSON which would mean that you don't need a separate file.

cowboyd (commented):
@minkimcello If you'd like to pair on this, let me know and we'll set up a time

sample/bin/template.js
```diff
@@ -0,0 +1,50 @@
...
+
+function processTemplate() {
+  let pkgjson;
+  let files;
+  let templateName;
+  if(process.argv.includes('--only')){
+    let onlyFlagIndex = process.argv.indexOf('--only');
+    templateName = process.argv[onlyFlagIndex + 1];
+    if(templateName !== 'cypress' && templateName !== 'bigtest' && templateName !== 'jest'){
+      throw new MainError({ message: `${formatErr(`The --only flag received an unrecognized framework.`)}` });
```
cowboyd: In order to make the error message helpful, we should provide what the the expected value and the actually value was. E.g.

```
expected the value of `--only` to be one of `jest`, `cypress`, or `bigtest`, but was `bligga-blorg`
```

minkimcello: Screen Shot 2021-03-18 at 10 33 29 AM

Too colourful?

cowboyd: 😆 perhaps. Is there a way to dial it down just a tad without losing the magic?

minkimcello: Screen Shot 2021-03-19 at 9 38 08 AM

We could just keep it all the same color

cowboyd (approved):

pittst3r (approved):
👍🏻

PR #891, opened 2021-03-05
feat: allow create inspector from interactor constructor

wKich:
## Motivation

For the Interactor inspector, we have to show all matched elements and be able to run actions for them. Interactors are lazy to resolve an element and expect to find exactly one element by a filter. So, the only way to execute an action from the inspector is to create an interactor with a unique key, like `selector`, but it's a very inefficient operation especially we already have all matched elements.

## Approach

We discussed with @cowboyd to expose the `createInspector` function that returns an object with the `all` method that returns all matched elements by interactor's selector, those wrapped to the interaction API.


packages/interactor/src/constructor.ts
```diff
@@ -190,7 +177,31 @@ export function instantiateInteractor<E extends Element, F extends Filters<E>, A
...
+      });
+    },
+
+    absent(): ReadonlyInteraction<void> {
+      return check(`${interactor.description} does not exist`, () => {
+        return converge(() => {
+          resolveEmpty(unsafeSyncResolveParent(options), options);
+        });
+      });
+    }
```
jnicklas, line 214: What about `has` and `is`?

packages/interactor/src/constructor.ts
```diff
@@ -100,9 +102,10 @@ function unsafeSyncResolveUnique<E extends Element>(options: InteractorOptions<E
   return resolveUnique(unsafeSyncResolveParent(options), options) as E;
 }
 
-export function instantiateInteractor<E extends Element, F extends Filters<E>, A extends Actions<E>>(
+function createInteractor<E extends Element, F extends Filters<E>, A extends Actions<E>>(
```
jnicklas: It's a bit confusing that there are now two functions called `createInteractor` which have different purposes. Can we rename this to something else?

jnicklas:
In general, I think this is a pretty nice and very clever approach. It basically gives us the best of both worlds. I am a bit worried about how this interacts with `find`? I think the behaviour could be a little strange.

Another downside of this is that it introduces an API which you really shouldn't use in normal test suites, and which is mainly useful for diagnostics, but some people might still end up finding and using this API and having a frustrating experience. One reason this worries me is that we had a similar method in Capybara, and it caused a bunch of frustration.

packages/interactor/src/constructor.ts
wKich: They defined in the `createInteractor` function. I can move them here if it makes more sense.

packages/interactor/src/constructor.ts
wKich: I agree. It is a little bit messy with `initInteractor` and `instantiateInteractor`. If I rename `createInteractor` to `instantiateBaseInteractor` will it be ok?

wKich:
@jnicklas `find` should works as expected because for `all`, we found the top-level elements. There is a case when you want to find all nested elements. So maybe we should allow pass `ancestors` as an optional parameter to `all` function.

I agree that this API must be used only for diagnostics. It's a good point. For my purpose, I need specifications only and the function where I could pass the element resolver function. Is it ok to export `createInteractor` function? The name definitely should say that that function is internal.

cowboyd:
@jnicklas @wKich Maybe have a separate `createInspector()` API which is a meta-api for doing meta-things and nobody ever really needs to use it?

```js
import { createInspector, Button, Checkbox } from '@bigtest/interactor';

let inspector = createInspector({ Button, Checkbox });

for (let inspectableButton of inspector.Button.all()) {
  // do the things.
}
```

We might even want to make the inspector an async iterable that did the polling loop for you and only gave you new inspector instances whenever needed:

```js
async function inspect() {
  let inspector = createInspector({ Button, Checkbox });
  inspector.start();
  for await (let reflection of inspector) {
    // Do stuff.
  }
}
```
```

wKich:
`createInspector` sounds good to me

wKich:
- Added `createInspector` function
- Implement `find` method to return inspector API for nested elements

jnicklas (approved):
Looks great! Very clever design!

PR #892, opened 2021-03-05
Update effection dependencies on all packages

minkimcello:
## Motivation

To update all of the effection dependencies to `v1`.

## Approach

As you can see in this effection [release PR](https://github.com/thefrontside/effection/pull/221), with the exception of `@effection/node` all of the other effection packages' dependencies from `v0` were all up to date and are identical to `v1.0.0`:
```
effection                    0.8.0 => 1.0.0
@effection/channel           0.6.8 => 1.0.0
@effection/events            0.7.9 => 1.0.0
@effection/fetch             0.1.3 => 1.0.0
@effection/subscription     0.12.0 => 1.0.0
```
And here's the [release PR](https://github.com/thefrontside/effection/pull/243) for `@effection/node@1.0.1` which was to include the [fix](https://github.com/thefrontside/effection/pull/238) for the `ctrlc-windows` package not including `not-windows.js` file.


cowboyd (approved):
I know it's just a version number, but why do do I feel 10x more legit?

PR #895, opened 2021-03-09
feat: add `specification` static prop for interactor

wKich:
## Motivation

Simplify getting interactor specification for composing/reuse actions/filters between interactors

## Approach

Add `specification` static prop for interactor's constructor

```ts
import { Button } from '@bigtest/interactor';

const HiddenButton = Button.filters({ visible: Button.specification.filters.visible.apply });
```


wKich:
I'm closing it, to think more wisely about how to deal with the element generic type that conflicts between two interactors

Issue #845, opened 2021-02-08
Add `HTML` and `FormField` to the builtin interactor list.


cowboyd:
With the advent of the the builder syntax, we have extracted the `HTML` interactor which is the parent of almost all of our interactors.  Not only is this handy for defining new interactors, but it is useful in its own right for making abstract assertions.

This is an example from a FOLIO test suite using RTL matchers to make a very broad assertion that the text "Not Authorized" is _somewhere_ in the document. There was no clear way on how to do this with interactors, so the fallback was to go back to using a known method from RTL.

```js
import { expect, getByText } from './test/helpers';

expect(getByText('Not Authorized')).toBeInTheDocument();
```

However, by using the generic `HTML` interactor combined with the `containing()` matcher we can achieve this elegantly:

```js
import { HTML, including } from '@bigtest/interactor';

HTML(including('Not Authorized')).exists()
```

PR #902, opened 2021-03-13
Update docs with guides on using matchers

### LINKED ISSUE #824: Update guides to show using matchers
With #797 we now have the capability to match powerfully on interactor filters, not just using `deepEquals`. This means that if you have html that is like:

```html
<p title="In which we greet the user"> Hello User! How are you today?</p>
```

We don't necessarily want to have to type all of those things in the test case. like:

```ts
import { Paragraph } from '@bigtest/interactor';

Paragraph("Hello User! Howe are you today?").has({ title: "In which we greet the user" });
```

Not only is this ugly, but it's brittle in the event that we want to do something like change the punctuation or capitalization. This is the kind of case that matchers were meant to solve. Matchers can be used both for locators and filters in finding interactors and also for making assertions. Thus, the above could be re-written as:

```ts
import { Paragraph, including, matching } from '@bigtest/interactor';

Paragraph(including("Hello")).has({ title: matching(/greet the user$/) });
```

Not only do we have the `including()`, and `matching()` matchers, but there are also combinators: `or()`, `and()`, `not()`:

```ts
or(containing("Red"), containing("Green"))
```

And quantifiers `some()`, `every()`  for working with filters that are iterable:

```ts
import { some } from '@bigtest/interactor';

Column("Username").has({ rows: some("cowboyd") });
```
We need to integrate these matchers into the guides. Maybe even under their own section.

minkimcello:
## Motivation

To update our docs with guides on how to use matchers. Resolves #824.

## Approach
[Here](https://deploy-preview-902--bigtest.netlify.app/bigtest/docs/interactors/assertions-matchers) is the new `Assertions and Matchers` page, but I also had to make several changes to the previous [`Locators, Filters, and Actions`](https://deploy-preview-902--bigtest.netlify.app/bigtest/docs/interactors/locators-filters-actions) page so please read through both pages.

## TODOs
- [x] ~Should `applyMatcher()` be included in the guides? In discussing how one can write their own matchers, I show how `including()` is implemented and provide a `greaterThan()` example neither of which use `applyMatcher()`.~
- [x] ~Need to add a sneak peak example for matchers in the quick start page.~ Later?
- [x] Not scoped to this PR but should mention that #845 is on my TODOs list after this PR is finished.


website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,228 @@
...
+
+It makes more sense to say "text field _is_ visible" rather than "text field _has_ visible". Generally, we recommend using `is()` if your assertion is an adjective, and use `has()` if your assertion is a noun. For example, if we wanted to assert for `visibility`, then it would sound more natural to use `has()` instead:
+
+```js
+TextField({ id: 'username-id' }).has({ visibility: true });
+```
+
+## Matchers
+
+The methods in which we use locators and filters, as we have illustrated in the previous page, are very explicit. There may be times when some flexibility in your interactors is necessary in order to write better tests. And that's where matchers come in. <!-- the paragraph above needs to be fact checked -->
```
minkimcello: 🤔

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,228 @@
...
+
+```js
+MultiSelect().has({ values: every(blueOrGreen) });
+```
+
+Composing a matcher using other matchers, like we did for `blueOrGreen()`, is convenient because it delegates most of the matcher's logic as well as the error message.
+
+To create your own matcher without the use of any of the preexisting ones, you will need to create a function that returns a `{ match(), format() }` object.
+
+The `match()` function is where you place all of the matcher logic. It takes an argument `actual` which represents the values from the interactors. Here's how the `including()` matcher is implemented:
```
minkimcello: 🤨

cowboyd:
> Should applyMatcher() be included in the guides? In discussing how one can write their own matchers, I show how including() is implemented and provide a greaterThan() example neither of which use applyMatcher().

`applyMatcher` is a private API, and so our docs shouldn't talk about it.

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,228 @@
...
+TextField({ id: 'username-id' }).has({ placeholder: 'USERNAME' });
+```
+
+The difference between this approach and using `exists()` is that `exists()` is less refined as it will succeed as long as there is at least one match. With the two textfields for the username and password, the assertion would fail on if we were to write a test like this:
+
+```js
+TextField().has({ placeholder: 'USERNAME' });
+```
+
+It would fail on account of the textfield that has the placeholder value `PASSWORD`. You therefore need to choose the assertion method that is most appropriate for your tests.
```
jorgelainfiesta: > You therefore need to choose the assertion method that is most appropriate for your tests.

Instead of listing caveats of each method, can we turn it around and list the use cases and discuss which methods are the appropriate ones for each case?

website/docs/interactors/4-matchers-assertions.md
jorgelainfiesta: I think it'd be better to formulate the definition of matchers from scratch, without referencing previous pages or concepts. 

Why are these functions called matches? From the way the docs are written, for me they look more like "Assertion helpers" or something like that. I think it could be beneficial to avoid introducing a new term like 'matchers', specially because it's not really used when using them. Of course, I may be missing understanding, which is why I ask the reason of the name.

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,228 @@
...
+
+:::note
+Matchers are meant to be used for the _values_ of locators and filters and they cannot be substituted for the actual filters:
+```js
+Heading().has({ id: or('foo', 'bar') }); // good
+Heading().has({ or(id: 'foo', id: 'bar') }); // bad
+```
+:::
+
+### including, matching
```
jorgelainfiesta: I think we'll have to make the two sections have a parallel structure. For instance, we'd have to separate the `Assertions` section in subsections by groups of functions just as we do here.

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,228 @@
...
+export const blueOrGreen = or(
+  including('Blue'),
+  including('Green')
+);
+```
+
+You can then import and use the new matcher in your tests like so:
+
+```js
+MultiSelect().has({ values: every(blueOrGreen) });
```
jorgelainfiesta, line 203: Perhaps we can change the example to a single value assertion such that we don't have to use `every` first? and then we can say something like "You can compose your own matches with other matchers" and use the multiselect example. Just so people understand that they don't need to use `every` to use their own matcher.

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+Composing a matcher using other matchers, like we did for `blueOrGreen()`, is convenient because it delegates most of the matcher's logic as well as the error message.
+
+To create your own matcher without the use of any of the preexisting ones, you will need to create a function that returns a `{ match(), format() }` object.
```
jorgelainfiesta, line 208: I think we should break this section in two subsections, one for "Composing helpers" and one for "Writing helpers from scratch"

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,228 @@
...
+
+The `match()` function is where you place all of the matcher logic. It takes an argument `actual` which represents the values from the interactors. Here's how the `including()` matcher is implemented:
+
+```js
+export function including(subString) {
+  return {
+    match(actual) {
+      return actual.includes(subString);
+    },
+    format() {
```
jorgelainfiesta, line 218: is `format` only used to format the error message? if so can we name it more explicitly like `formatErrorMessage`? format reads a bit confusing imo because it is also makes me think that I'll use format to check a value complies with a format or something like that

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,228 @@
...
+
+╒═ Filter:   id
+├─ Expected: including "foo"
+└─ Received: "bar"
+```
+
+Here is a simple example of how a `greaterThan()` matcher can be constructed:
+
+```js
+export function greaterThan(number) {
```
jorgelainfiesta, line 246: So this function is not included in the built in helpers right?

website/sidebars.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
     'interactors/quick-start',
     'interactors/built-in-dom',
     'interactors/locators-filters-actions',
+    'interactors/assertions-matchers',
```
jorgelainfiesta, line 6: The sidebar ends up being not too useful imo. It seems just two groups of concepts but tell nothing to the reader that don't know them. Can we name each page after the problem they can learn to solve in each page instead of the functions that we'll cover?  

Screenshot 2021-03-16 at 19 41 49

For instance, I'd go as far as saying "Asserting with Interactors" as the title for the second page. I'm not sure what'd be an accurate title for the first one.

jorgelainfiesta (commented):
Great work Min! I added some thoughts that come to mind

website/docs/interactors/4-matchers-assertions.md
cowboyd: What about something like `description()`?

website/sidebars.js
minkimcello: I think saying `Asserting with Interactors` will be redundant because this is the `Interactors` docs.

website/docs/interactors/4-matchers-assertions.md
minkimcello: It's not

website/docs/interactors/4-matchers-assertions.md
minkimcello: I don't think it'd be possible to separate `Assertions` into sections because of the way it's written. And removing `Matchers` subsections doesn't seem like a good idea either.

website/docs/interactors/4-matchers-assertions.md
jenweber: We changed this to:
> Matchers are methods that let you add logic or flexibility when you are searching for elements. They can be used within locators, filters, and assertions. In this section, you will learn how to use matchers to write better tests.

website/docs/interactors/4-matchers-assertions.md
jenweber: We updated this to show the positive advice about when to use each method first, and listing caveats second.

website/docs/interactors/4-matchers-assertions.md
jenweber: We have now split up the assertions section.

website/docs/interactors/4-matchers-assertions.md
jenweber: We ended up breaking this into 2 examples - one that is simple like `hasFoo` that asserts about a Heading, and the other more complex MultiSelect example.

website/docs/interactors/4-matchers-assertions.md
jenweber: This is now divided up.

website/sidebars.js
jenweber: @jorgelainfiesta what do you think about a sidebar like this? It would require minimal refactoring of the content itself. We want to include the terminology in the title, so that someone can say "That's a Matcher. There is a docs pages about them..." when someone is learning. But also, we can give a clue what they are for.

```
Quick start
Built in dom
Queries with Locators and Filters
User Actions
Assertions
Flexibility with Matchers
Write your own Interactor
Integrations
```

website/docs/interactors/4-matchers-assertions.md
minkimcello: Created an issue [here](https://github.com/thefrontside/bigtest/issues/906)

website/sidebars.js
minkimcello: Created issue [here](https://github.com/thefrontside/bigtest/issues/907)

minkimcello:
@cowboyd I'm trying to find an appropriate place in the [Quick Start](https://frontside.com/bigtest/docs/interactors) to include a matchers sneak peak but it feels to me that page goes over the big picture and matchers is more a micro feature.

If we were to include it somehow, I could sneak them into either the [`sign button should be absent`](https://frontside.com/bigtest/docs/interactors#making-test-assertions) example or into the [airline](https://frontside.com/bigtest/docs/interactors#a-look-at-real-world-interactors) example. But, including it into either of those examples would require us to address it saying "hey, this is a matcher. read more about it \[here](link)" but since we haven't introduced any other concepts yet at that point I think that's jumping the gun a little.

We could add a sentence in the [`Continue learning about Interactors`](https://frontside.com/bigtest/docs/interactors#continue-learning-about-interactors) section to include `Assertions and Matchers` but since we're considering reorganizing our content (issue #907), I think we can put it off until then.

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,267 @@
...
+
+- Text flexibility like `including()` and `matching()`
+- Conditionals like `and()`, `or()`, and `not()`
+- Iterables like `some()` and `every()`
+
+### When to use Matchers
+
+If your tests are written against a simulated database, it might not be important what the randomized users' names are. Maybe you just want to assert that a login success message includes the word "welcome", such as "Welcome, Taylor!" Or perhaps it does not matter what the name is, but you still want to assert that a name is displayed. These are situations where matchers would come in handy.
+
+### including, matching
```
jorgelainfiesta: Would it be possible to have a parallel structure in this section for subtitles as in the previous ```Using `has()` to assert for one result``` etc?

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,267 @@
...
+
+If there is no link that matches all three conditions, the test would fail on account of the interactor not returning any elements.
+
+In this next example, we will pass in one false value to `or()`:
+
+```js
+Link('Google').has({ 
+  href: or(
+    'https://google.com',
+    'https://twitter.com'
```
jorgelainfiesta: Just for the sake of making the example more realistic, maybe the second option could be `https://google.ca` or another domain (which makes me wonder, can `or` take an arbitrary amount of parameters?)

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,267 @@
...
+MultiSelect().has({ values: every(matching(/^Neon/)) });
+```
+
+In the two tests above we are passing in the `including()` and `matching()` matchers into `some()` and `every()`. Once again, `and()`, `or()`, `not()`, `some()`, and `every()` can take matchers as its arguments. This means you can chain them together multiple times to cater to your needs.
+
+Though the matchers are already ergonomic, you can make your tests even tidier and easier to read by creating your own matchers.
+
+## Composing matchers
+
+There are two ways you can write your own matcher: by piggybacking on preexisting matchers or you can create your own from scratch. We will cover both methods in this section.
```
jorgelainfiesta: Now we have two sections so I think 'cover both methods in this section' no longer apply?

jorgelainfiesta (commented):
Awesome work y'all! it's looking great!! I left two questions/comments for your consideration. Thanks!

website/docs/interactors/4-matchers-assertions.md
minkimcello: We can change up the example however you'd like. Any specific reason for `https://google.ca`?

`or` can take more than two arguments:

> These next three matchers: and(), or(), and not() are different from the first two matchers in that they can take multiple arguments and the arguments can be either a value or another matcher.

website/docs/interactors/4-matchers-assertions.md
minkimcello: Oops, I'll fix that.

website/docs/interactors/4-matchers-assertions.md
minkimcello: Could you elaborate?

website/docs/interactors/4-matchers-assertions.md
```diff
@@ -0,0 +1,267 @@
...
+
+```js
+MultiSelect().has({ values: every(blueOrGreen) });
+```
+
+## Creating matchers from scratch
+
+To create your own matcher without the use of any of the preexisting ones, you will need to create a function that returns a `{ match(), format() }` object.
+
+The `match()` function is where you place all of the matcher logic. It takes one argument, `actual`, which represents the values from the interactors. Here's how the `including()` matcher is implemented:
```
cowboyd, line 210: Can we be more specific. Instead of "the values from the interactors", actual will be the value of the locator or filter

cowboyd (commented):
matchers are mentioned in the assertion section, but they can also be used anywhere you would pass a locator or a filter. That includes for finding. Do we emphasize this enough? I seem to recall an todo item to tease matchers in the section on filters and locators.

minkimcello:
@cowboyd

> I seem to recall an todo item to tease matchers in the section on filters and locators.

I believe we discussed including it in the quick start. I've pointed it out in the comment [above](https://github.com/thefrontside/bigtest/pull/902#issuecomment-804967230).

I also skimmed through the `Locators, Filters, and Actions` page to see where we might be able to include matchers. But the examples we use are simple and the use cases for matchers are for not-simple scenarios. If we had to include matchers, it would either be "Why would you write a test that way?" or we would need to divert to a completely different example. I'll ask Jen what she thinks.

website/docs/interactors/4-matchers-assertions.md
minkimcello: > The `match()` function is where you place all of the matcher logic. It takes one argument, `actual`, which will be the value of the locator or filter.

☝️ is this okay?

We pondered over this very sentence. The reasoning behind why I said "values" instead of being more specific is that in the context of creating your own matchers it doesn't make a difference whether "actual" is going to be a locator or filter value.

website/docs/interactors/4-matchers-assertions.md
jorgelainfiesta: In the previous section, the functions were introduced with their function. For instance, the title is ```Using `absent()` for non-existant results``` instead of just `absent`

website/docs/interactors/4-matchers-assertions.md
jorgelainfiesta: No reason, just that a link to `Google` probably has a `google` href. How about this:

```
Link('Google').has({ 
  href: or(
    'https://google.com',
    'https://google.co.uk',
    'https://google.es'
...
```

website/docs/interactors/4-matchers-assertions.md
minkimcello: This is how it is now:
Screen Shot 2021-03-30 at 9 28 30 PM

And here's how it would look if we matched the pattern:
Screen Shot 2021-03-30 at 9 28 43 PM

Maybe there's some better text to put in there instead of `String matchers:` and et cetera. But unlike assertions, which we're all already very familiar with, I think matching the pattern for the matcher section might not be necessary in my opinion. Because `String matchers: ..`, `Conditional matchers: ..` and `Iterable matchers: ..` don't really tell the users much until they actually read the content.

I mean at the same time I think the descriptive headings for the assertions aren't necessary either. It's crowding the sidebar and making it look more complicated.

website/docs/interactors/4-matchers-assertions.md
minkimcello: Oh I see what you're saying. We're straying off into an unrealistic example. However, I do like that the two are very different and it makes it easier to get the message across. If anybody is skimming through the docs, they might not notice the difference between `.com` and `.ca`.

So when we take this example:
```js
Link('Google').has({
  href: or(
    'https://google.com',
    'https://twitter.com'
  )
})
```

It's very clear without any context that `or()` is a conditional that only needs one condition to return true. The reader doesn't even need to remember the HTML layout to understand the example:
```html
<a href="https://google.com">Google</a>
<a href="https://twitter.com">Twitter</a>
```

Whereas with:
```js
Link('Google').has({
  href: or(
    'https://google.com',
    'https://google.co.uk'
  )
})
```

The reader might go "why not?" It's a silly example but matchers are used in complex situations. Because interactors are already so intuitive that the need for matchers would, I assume, be in special cases. Meaning I think it's appropriate to show unrealistic examples as long as it helps the readers understand how they work in the quickest way possible.

website/docs/interactors/4-matchers-assertions.md
jenweber: There are a few considerations here.

For longevity, it's good to not list function names in headings, so that as more methods get added to the library, or others get deprecated, your section headings don't change. This is important for things like keeping blog post links stable. We can see an example of generic naming in the Jest matchers docs: https://jestjs.io/docs/using-matchers

On the other hand, including the function names gives you a sense of what matchers are, and lets someone quickly find the docs when they are using it as a reference material. It's a nice escape hatch for when there is not a neat, descriptive heading to use.

As a general rule in Ember docs, I go with option #1. In shorter guides/articles, keeping links stable matters less. These articles are fairly short, especially if matchers get to be their own article in the future, and someone will not be lost of the heading disappears.

I think a good case could be made for either option. If I had to choose, in this case I'd personally go with using labels like "Iterable matchers" and not list the functions in the headings, just to allow for future growth and to keep the table of contents neat. Someone can command-F to search for the function names, and they are listed at the beginning of the section. That list could be links. I do think the proposed heading names like "iterable" and "conditional" have useful meaning.

website/docs/interactors/4-matchers-assertions.md
jenweber: I like the example showing 3 TLDs. I think when there are three, it's easier to spot the difference. The other reason I like it is that I would not think to pass more than 2 parameters into this, and so it's teaching something useful about the API without introducing any new concepts/commentary.

This example is an opportunity to show off Frontside instead of google, potentially:
```
https://frontside.com/
https://www.frontside.com/
```

jenweber:
Regarding this last comment, I think putting matchers in their own file is the way to go here. Then you can show using matchers for assertions and other things, without trying to introduce the other concepts at the same time. Someone should already know what filters, assertions, etc. are.

website/docs/interactors/4-matchers-assertions.md
minkimcello: @cowboyd I'm going to assume the thumbs up are for the explanation for why I went with just "values". I'll resolve this for now?

website/docs/interactors/4-matchers-assertions.md
minkimcello: @jenweber 

> The other reason I like it is that I would not think to pass more than 2 parameters into this, and so it's teaching something useful about the API without introducing any new concepts/commentary.

The first conditional matcher example we use takes 3 arguments. I did this intentionally because I also assumed people would instinctively limit themselves to 2 arguments otherwise.

> I like the example showing 3 TLDs.

Are you saying our example should have three links?

```html
<a href="https://frontside.com">Frontside</a>
<a href="https://google.com">Google</a>
<a href="https://twitter.com">Twitter</a>
```

Three links is a bit overkill and over complicates the examples without adding more value in my opinion. For `and` and `or` we can get away with using just one link:

```js
and(
  true1(),
  true2(),
  true3()
)
```

```js
Link().has({
  or(
    'frontside.com',
    'whatever random href although taking from the second link would make the example clearer'
})
```

But for `not`, it makes more sense to have two different links. I'm alternating the usage of matchers between assertions and in the interactor so people don't get stuck using them for one or the other. The `and` example uses matchers for the filter of the interactor and `or` uses matchers for the assertion. It's not really practical or realistic to use `and` or `or` for the locator in these particular examples. But in order to use `not` in the locator, we need a second link or else the assertion becomes pointless:

```js
Link(not('Frontside')).has({ href: 'whatever' }); // this would fail because we don't have a second link
Link(not('whatever')).has({ href: 'https://frontside.com' }); // this would be a pointless test
```

So we can't get away with one and three is too many.

website/docs/interactors/4-matchers-assertions.md
minkimcello: Screen Shot 2021-04-01 at 10 12 51 AM

minkimcello: @jorgelainfiesta If that looks good, can you resolve this conversation?

website/docs/interactors/4-matchers-assertions.md
jorgelainfiesta: I don't think I follow your argument Min. I thought Jen was suggesting something like this:

```js
Link('Frontside').has({
  href: or(
    'https://frontside.com',
    'https://www.frontside.com',
    'https://frontside.io'
  )
})
```

minkimcello: @jorgelainfiesta For the `or()` matcher we just need to illustrate that one false statement can still make a test pass because it's an either-or situation.

```js
Link('Frontside').has({
  href: or(
    'frontside',
    'frontside',
    'frontside'
  )
})
```

☝️ isn't as clear as:

```js
Link('Frontside').has({
  href: or(
    'frontside',
    'not-frontside'
  )
})
```

Was the point I was trying to make.

And also do we actually add the different TLDs into the HTML example?

```html
<a href="https://frontside.com">Frontside</a>
<a href="https://www.frontside.com">Frontside</a>
<a href="https://www.frontside.io">Frontside</a>
```

Because with only those three links, it's hard to discuss `not()` - was the other argument I was making.

I suppose we could do `frontside` and `google` (just one of each) and still use the `or` example with three different TLDs. 🤷‍♂️

jenweber: I was thinking that the HTML example could shoot show a single link, and the “or” matcher includes three different URLs. International TLDs is a more realistic example than frontside urls, but the frontside URLs get the job done too. I think either example works here.

minkimcello: I still think the non-realistic example drives the point better of how `or()` is used than using `google.ca` and `google.co.uk`.

Because if I was reading this, I'll be thinking, "why not just use `href: including('google')`? what are the benefits of using `or()`?". And in what real life scenario do we need to ensure that it's just one of the three of all the TLDs? It would make sense to use `not()` if we wanted to specify that it should not be `google.ca`.

So by doing `or('google.com', 'twitter.com')`, I think we all know how silly the example is but we get the point much faster.

But anyway, I changed it. You can see the commit [here](https://github.com/thefrontside/bigtest/pull/902/commits/2d8872c7521ceb6247b0079eca363309c9b0b4b2). I didn't switch one of the links to Frontside yet because people will already be on the frontside website when they're reading this guide, and putting our Frontside link next to Google's might give off an inaccurate impression of how we perceive ourselves.

jorgelainfiesta: Ah gotcha, I see your point. Indeed `including` would've been a better choice indeed. What is the use of `or` then? I can't think of a use case but I guess there are plenty?

For me, reading incoherent examples is confusing. I'm like "why would they do that?"

jorgelainfiesta (approved):
Looks great already! I'd ship it

website/docs/interactors/4-matchers-assertions.md
minkimcello: The only real use case I saw was where someone wants to assert for a Button that may or may not be visible but still available in the DOM:
```js
Button({ visibility: or(true, false) }).is(whatever);
```

So we could change all the examples for `and`, `or`, and `not` so it uses a button instead but that boolean for visibility really only applies to `or` so there's a lot of switching back and forth in the examples even if we use the same Button.

Like you mentioned below, let's just brush this off for now and update it later.

PR #904, opened 2021-03-23
fix: add types restrition for test builder

wKich:
## Motivation

if you are trying to write an invalid sequence in tests, it will nice to get the feedback earlier than get an exception on runtime

## Approach

Solution pretty simple, add `Omit` for methods' return type

It's still possible to run these methods in runtime, but we throw an understandable error. So I think everything is ok.

### DIFF
--- a/packages/suite/src/dsl.ts
+++ b/packages/suite/src/dsl.ts
@@ -43,6 +43,9 @@ class TestStructureError extends Error {
 
 type TestBuilderState = 'step' | 'assertion' | 'child';
 
+type AssertionTestBuilder<C extends Context> = Omit<TestBuilder<C>, 'step'>
+type ChildTestBuilder<C extends Context> = Omit<TestBuilder<C>, 'step' | 'assertion'>
+
 /**
  * A builder API for constructing BigTest tests. This is usually created via the {@link test} function.
  *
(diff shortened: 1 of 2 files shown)


jnicklas:
IIRC we actually didn't add a typing like this somewhat intentionally. The thinking was that we're unable to attach an explanation *why* this happens to a type error, and so it would feel "broken" to a user, whereas with a runtime exception we're able to add an explanation why this occurs. We can totally change our stance on this, but that was the original thinking, for some background.

wKich:
I think what fails early is much better than we allow a user to write a bunch of tests and then fail in runtime. We definitely have to cover the order of operations in docs so a user can find it.

Btw, a custom error message might be implemented here https://github.com/microsoft/TypeScript/pull/40468

jnicklas:
This PR has been lingering here for a long time. Can we make a decision on whether or not to merge this? I still think that the original reasoning of being able to explain the error on failure makes a case for failing at runtime, but I don't feel super strongly about this. The TypeScript throw type proposal that @wKich linked looks like the absolute best solution, but it seems unlikely to be implemented anytime soon, unfortunately.

@cowboyd do you want to make a call here?

wKich:
Another way to show the user that he can't use steps in an assertion is by marking them as deprecated and adding some description. But it's an ugly workaround in my opinion.

But I'm ok with this implementation if we cover that moment in docs.

cowboyd:
I think I prefer the runtime messaging about _why_ something is impossible in this case since the type error is going to be really, really opaque.

Bear in mind that all of this is likely to be moot if end up re-visiting the entire synaxt which I think we will before 1.0

cowboyd:
So to be unambigous, I think we should not implement this at this time, and instead completely re-think our syntax in the context of what we've learned (and also in the context of cross-platform agents)

PR #909, opened 2021-03-25
✨  add `indeterminate` filter for checkbox

wKich:
## Motivation

The `checkbox` element could have the third possible state, `indeterminate`. Where it's not `checked` nor `unchecked`. You can read more about that state and check the examples here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes

## Approach

Add the `indeterminate` filter for `Checkbox` interactor


packages/interactor/test/definitions/check-box.test.ts
```diff
@@ -215,6 +215,24 @@ describe('@bigtest/interactor', () => {
...
+            <input type="checkbox" id="accept-field"/>
+          </p>
+        `);
+
+        // NOTE: No browser currently supports indeterminate as an attribute. It must be set via JavaScript.
+        // NOTE: See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes
+        await CheckBox('Accept').perform(e => e.indeterminate = true)
+
+        await expect(CheckBox('Accept', { indeterminate: true }).exists()).resolves.toBeUndefined();
+        await expect(CheckBox('Accept', { indeterminate: false }).exists()).rejects.toHaveProperty('name', 'NoSuchElementError');
```
cowboyd, line 234: For sanity check, can we also assert that both `checked` and `unchecked` are false?

wKich: The `checked` could be `true` or `false`. Think about it as the `indeterminate` state has a higher priority to UI view. Maybe we have to check `indeterminate == false` in a `checked` filter?

wKich: image

wKich: Hm... Looks like we also must reset `indeterminate` state on any `check/uncheck/toggle/click` actions. Ok, I'll fix it

cowboyd: Where are we on this?

cowboyd: Because I think this should be ready to go, right?

wKich: Yep, I remember about it. I just was a little busy with datepicker.

cowboyd: Ok, just wanted to make sure I didn't let it slip through the cracks

cowboyd (approved):
Looks good to me. I don't know if I've ever used this state before, but I guess you learn something every day.

packages/interactor/src/definitions/check-box.ts
```diff
@@ -12,9 +12,22 @@ const CheckBoxInteractor = FormField.extend<HTMLInputElement>('check box')
     },
   })
   .actions({
-    check: ({ perform }) => perform((element) => { if(!element.checked) element.click(); }),
-    uncheck: ({ perform }) => perform((element) => { if(element.checked) element.click(); }),
-    toggle: ({ perform }) => perform((element) => { element.click(); }),
+    check: ({ perform }) => perform((element) => {
+      element.indeterminate = false;
+      if(!element.checked) element.click();
```
cowboyd: If we are going to set the indeterminate state because that is default browser behavior, then we should we should ensure that the click was not cancelled.

wKich: Done

packages/interactor/test/definitions/check-box.test.ts
```diff
@@ -261,6 +261,40 @@ describe('@bigtest/interactor', () => {
         await expect(CheckBox('Accept', { indeterminate: false }).exists()).resolves.toBeUndefined();
         await expect(CheckBox('Accept', { indeterminate: true }).exists()).rejects.toHaveProperty('name', 'NoSuchElementError');
       });
+
+      describe("Don't reset indeterminate on canceled click", () => {
+        beforeEach(async () => {
+          await CheckBox('Accept').perform(e => e.onclick = function (e) { e.preventDefault() })
+        })
+
+        it("Reset indeterminate on toggle", async () => {
```
cowboyd:
```suggestion
        it("resets indeterminate on toggle", async () => {
```

packages/interactor/test/definitions/check-box.test.ts
```diff
@@ -261,6 +261,40 @@ describe('@bigtest/interactor', () => {
...
+          await CheckBox('Accept', { indeterminate: true }).toggle()
+
+          await expect(CheckBox('Accept', { indeterminate: true }).exists()).resolves.toBeUndefined();
+          await expect(CheckBox('Accept', { indeterminate: false }).exists()).rejects.toHaveProperty('name', 'NoSuchElementError');
+        });
+
+        it("Reset indeterminate on check", async () => {
```
cowboyd:
```suggestion
        it("resets indeterminate on check", async () => {
```

packages/interactor/test/definitions/check-box.test.ts
```diff
@@ -261,6 +261,40 @@ describe('@bigtest/interactor', () => {
...
+          await CheckBox('Accept', { indeterminate: true }).check()
+
+          await expect(CheckBox('Accept', { indeterminate: true }).exists()).resolves.toBeUndefined();
+          await expect(CheckBox('Accept', { indeterminate: false }).exists()).rejects.toHaveProperty('name', 'NoSuchElementError');
+        });
+
+        it("Reset indeterminate on uncheck", async () => {
```
cowboyd:
```suggestion
        it("reset indeterminate on uncheck", async () => {
```

packages/interactor/test/definitions/check-box.test.ts
```diff
@@ -261,6 +261,40 @@ describe('@bigtest/interactor', () => {
...
+          await CheckBox('Accept', { indeterminate: true }).uncheck()
+
+          await expect(CheckBox('Accept', { indeterminate: true }).exists()).resolves.toBeUndefined();
+          await expect(CheckBox('Accept', { indeterminate: false }).exists()).rejects.toHaveProperty('name', 'NoSuchElementError');
+        });
+
+        it("Reset indeterminate on click", async () => {
```
cowboyd:
```suggestion
        it("reset indeterminate on click", async () => {
```

Issue #907, opened 2021-03-23
Reorganize the docs' sidebar


minkimcello:
[Context](https://github.com/thefrontside/bigtest/pull/902#discussion_r595448846) from #902

I think for the same reason why React has a `Introducing Hooks` page instead of a `Using State and other React features without Class` page, I think it's important to include the actual names of functions/concepts we're introducing for each pages' titles. For example, if someone drops into our Discord channel and we show them how `filters` work, they might go back to the docs and look for `Filters` and not `Querying Interactors`.

However, that suggestion lead us to rethink how the content should be reorganized now that we have more content:

| Current | PR #902 | Possible Solution* |
| :-- | :-- | :-- |
| Quick Start<br>Built-in Interactors<br>Locators, Filters, and Actions<br>Writing Interactors<br>Jest & Cypress<br><br><br><br> | Quick Start<br>Built-in Interactors<br>Locators, Filters, and Actions<br>🆕 **Assertions and Matchers**<br>Writing Interactors<br>Jest & Cypress<br><br><br> | Quick Start<br>Built-in Interactors<br>Querying with Locators and Filters<br>User Actions<br>Assertions<br>Flexibility with Matchers<br>Writing Interactor<br>Jest & Cypress |

_* I made small iterations from Jen's original suggestion [here](https://github.com/thefrontside/bigtest/pull/902#discussion_r599531467) (such as keeping `Jest & Cypress` instead of renaming it to `Integrations`) but these are all up for discussion._

Basically...
- We would extract the `Actions` section to its own page and split up `Assertions` and `Matchers` since there's really no reason for those two concepts to be grouped together
- The pages are renamed to be a little bit more descriptive while still containing the terminologies

jorgelainfiesta:
I agree with the solution that you propose on the third column, great stuff!

I also think keeping "Jest & Cypress" is better than "Integrations" because those are keywords that people will look for.

PR #917, opened 2021-04-22
Reorganize the docs' sidebar

minkimcello:
## Motivation

To sort of address #907.

## Approach

In #907 I outlined that I'd separate out `Assertions and Matchers` into two separate pages (`Assertions` & `Matchers`) and split `Locators, Filters, and Actions` into `Query with Locators and Filters` and `User Actions`. However, I find we don't have enough content for `Actions` to have it warrant its own page _yet_:

Group

It would be a very short page. Once accessibility is fleshed out, we'll probably have more stuff to add to the actions page so let's reevaluate the split later.

Also, instead of `Flexibility with Matchers`, I just named the new page `Matchers`. "Flexibility with Matchers" isn't that descriptive and doesn't tell us anything more about matchers for someone who doesn't know what matchers are. And having that page with the only "descriptive" title wouldn't be very symmetrical.

For the `Assertions and Matchers` split, I updated all the intro and outro paragraphs in the new pages as well as in the previous page, `Locators, Filters, and Actions`. I updated all the links too that referenced the old `Assertions and Matchers` page.


jorgelainfiesta (approved):
Looks good, I think your rationale is on point. Thanks!

PR #918, opened 2021-04-22
Change matcher format() to description()

### LINKED ISSUE #906: Rename the format function for matchers
From a discussion in [PR #902](https://github.com/thefrontside/bigtest/pull/902):

> is `format` only used to format the error message? if so can we name it more explicitly like `formatErrorMessage`? format reads a bit confusing imo because it is also makes me think that I'll use format to check a value complies with a format or something like that

I think we could call it `error()` or `errorMessage()`.

minkimcello:
## Motivation

Currently if someone wants to construct their own matcher, they have to pass in an object that looks like this:
```js
{
  match(): { /* ... */ },
  format(): { /* ... */ }
}
```

But with this new refactor, we're renaming `format` to `description`:
```js
{
  match(): { /* ... */ },
  description(): { /* ... */ }
}
```

This will resolve #906.

## Approach

- Instead of changing `formatMatcher` to `descriptionMatcher`, I flipped it to say `matcherDescription` as that seems more descriptive in my opinion.
- Added changeset for minor bump since we're still on v0.
- Removed tests for built-in matchers that called `matcher().description()` as that's not something a user would be invoking themselves.
- Added test for the `matching` matcher with the old `format` API to ensure it still works (but will output a deprecation message in the console).
- Updated documentation.

## TODOs

- [x] ~I should probably keep `format()` in to output a message to say it's deprecated...?~ Did.

### DIFF
--- a/.changeset/interactor-format-bye.md
+++ b/.changeset/interactor-format-bye.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/interactor": minor
+---
+
+Deprecate matchers' format() for description()
--- a/packages/interactor/src/constructor.ts
+++ b/packages/interactor/src/constructor.ts
@@ -154,7 +154,7 @@ export function instantiateBaseInteractor<E extends Element, F extends Filters<E
           let element = resolver(options);
           let match = new MatchFilter(element, filter);
           if(!match.matches) {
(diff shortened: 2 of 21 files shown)


cowboyd (approved):
LGTM 👍🏻

PR #921, opened 2021-04-30
Update docs to replace `createInteractor` in favor of `extends`

### LINKED ISSUE #844: Update guides to show using `extend()` inside authoring interactors.
Most of the time when you're creating your own interactors, you actually want to extend an existing interactor. In fact, most of the interactors in the builtin list use:

```ts
HTML.extend()
```

That way, they inherit the `id`, `title`, `visible` and `focused` filters, and they get the `click()`, `focus()`, and `blur()` actions for free.

However, all of our content on creating your interactors shows using `createInteractor()` which in reality is something that should rarely be used unless they don't want those filters for some reason, or they are writing an interactor for an `SVGElement` or some other kind of XML element.

### LINKED ISSUE #845: Add `HTML` and `FormField` to the builtin interactor list.

minkimcello:
## Motivation

To resolve #844 and resolve #845. 

Screen Shot 2021-04-30 at 12 23 53 PM

## Approach

- Added `HTML` and `FormField` to [`Built-In Interactors`](https://deploy-preview-921--bigtest.netlify.app/bigtest/docs/interactors/built-in-dom) page and gave a brief description on what the `HTML` interactor is for
- Small updates to the [`Locators, Filters, and Actions`](https://deploy-preview-921--bigtest.netlify.app/bigtest/docs/interactors/locators-filters-actions) page to explain how filters are inherited from other interactors
- ❗Big updates to the [`Writing Interactors`](https://deploy-preview-921--bigtest.netlify.app/bigtest/docs/interactors/write-your-own) page to replace all instances of `createInteractor()` to `extends()`

### DIFF
--- a/website/docs/interactors/2-built-in-dom.md
+++ b/website/docs/interactors/2-built-in-dom.md
@@ -30,6 +30,7 @@ Follow the links above to get more information on how to use these interactors t
 If your app has unique interfaces that are not covered by these built-in tools, you are encouraged to [write your own interactors](/docs/interactors/write-your-own).
 
 ### Page
+
 The `Page` interactor is unique. Unlike Big Test’s other built-in interactors, it's not designed to target one specific element but rather the whole page. It is useful for asserting for the url or title in your test environment:
 
 ```js
@@ -51,9 +52,13 @@ export default test('BigTest Runner')
   .assertion(Page.has({ title: 'BigTest Example App'}));
 ```
 
-<!-- 
 ### HTML
-todo -->
+
+The `HTML` interactor is not meant to be used directly in your tests but for _composing_ other interactors. How this interactor is used will make much more sense once you read the `Write Your Own Matchers` page. But for now you can think of the `HTML` interactor as the basic building block for all other interactors.
+
+We took many of the common HTML properties and interactions, such as `className` and `click`, and added them into the `HTML` interactor. This provides the convenience of not having to respecify these properties over and over again for each of your interactors.
+
+Take a look at the [source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts) for the `HTML` interactor to see which filters and actions were added.
 
 ## Up Next
 
--- a/website/docs/interactors/2-built-in-dom.md
+++ b/website/docs/interactors/2-built-in-dom.md
@@ -11,7 +11,6 @@ These are the default interactors that are offered out-of-the-box with BigTest:
 - [CheckBox](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/check-box.ts)
 - [FormField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts)
 - [Heading](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/heading.ts)
-- [HTML](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts)
 - [Link](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/link.ts)
 - [MultiSelect](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/multi-select.ts)
 - [Page](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/page.ts)
@@ -24,8 +23,6 @@ As you might have seen on the [Quick Start](/docs/interactors/) page, you can im
 ```js
 import { Button, TextField } from 'bigtest';
 ```
-
-Follow the links above to get more information on how to use these interactors to test your app.
 
 If your app has unique interfaces that are not covered by these built-in tools, you are encouraged to [write your own interactors](/docs/interactors/write-your-own).
 
@@ -42,7 +39,7 @@ _The `Page` interactor is instantiated differently than the other built-in inter
 We introduced `.exists()` and `.absent()` in the [Quick Start](/docs/interactors/) section but there are also `.has()` and `.is()` Interactor assertion methods. We will discuss their details on the [Assertions](/docs/interactors/assertions) page.
 :::
 
-And when using BigTest runner, the Page interactor can be used to navigate between routes:
+And when using the BigTest runner, the Page interactor can be used to navigate between routes:
 
 ```js
 import { Page, test } from 'bigtest';
@@ -51,14 +48,6 @@ export default test('BigTest Runner')
   .step(Page.visit('/contact'))
   .assertion(Page.has({ title: 'BigTest Example App'}));
 ```
-
-### HTML
-
-You can think of the `HTML` interactor as the basic building block for all other interactors. The `HTML` interactor is not meant to be used directly in your tests but for _composing_ other interactors. How this interactor is used will make much more sense once you read the `Write Your Own Matchers` page.
-
-Many common HTML properties and interactions, such as `className` and `click`, are included in the `HTML` interactor. This provides the convenience of not having to re-specify these properties over and over again for each of your interactors.
-
-Take a look at the [source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts) for the `HTML` interactor to see which filters and actions were added.
 
 ## Up Next
 
--- a/website/docs/interactors/2-built-in-dom.md
+++ b/website/docs/interactors/2-built-in-dom.md
@@ -9,7 +9,9 @@ These are the default interactors that are offered out-of-the-box with BigTest:
 
 - [Button](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/button.ts)
 - [CheckBox](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/check-box.ts)
+- [FormField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts)
 - [Heading](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/heading.ts)
+- [HTML](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts)
 - [Link](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/link.ts)
 - [MultiSelect](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/multi-select.ts)
 - [Page](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/page.ts)
@@ -48,6 +50,10 @@ export default test('BigTest Runner')
   .step(Page.visit('/contact'))
   .assertion(Page.has({ title: 'BigTest Example App'}));
 ```
+
+<!-- 
+### HTML
+todo -->
 
 ## Up Next
 
--- a/website/docs/interactors/2-built-in-dom.md
+++ b/website/docs/interactors/2-built-in-dom.md
@@ -54,9 +54,9 @@ export default test('BigTest Runner')
 
 ### HTML
 
-The `HTML` interactor is not meant to be used directly in your tests but for _composing_ other interactors. How this interactor is used will make much more sense once you read the `Write Your Own Matchers` page. But for now you can think of the `HTML` interactor as the basic building block for all other interactors.
+You can think of the `HTML` interactor as the basic building block for all other interactors. The `HTML` interactor is not meant to be used directly in your tests but for _composing_ other interactors. How this interactor is used will make much more sense once you read the `Write Your Own Matchers` page.
 
-We took many of the common HTML properties and interactions, such as `className` and `click`, and added them into the `HTML` interactor. This provides the convenience of not having to respecify these properties over and over again for each of your interactors.
+Many common HTML properties and interactions, such as `className` and `click`, are included in the `HTML` interactor. This provides the convenience of not having to re-specify these properties over and over again for each of your interactors.
 
 Take a look at the [source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts) for the `HTML` interactor to see which filters and actions were added.
 
--- a/website/docs/interactors/3-locators-filters-actions.md
+++ b/website/docs/interactors/3-locators-filters-actions.md
@@ -49,7 +49,7 @@ The locator of the `TextField` interactor is the `innerText` of its associated l
 </label>
 ```
 
-_See the source code for the TextField interactor [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts). The TextField interactor extends the [FormField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts#L6) interactor which is from where it inherits the locator function._
+_See the source code for the TextField interactor [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts). The TextField interactor extends the [FormField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts#L6) interactor, so it has a locator function._
 :::
 
 You can think of the locator as the "default filter" since filters and locators both provide the same functionality. The reason why BigTest offers both solutions is convenience, because having to pass in an object for each interactor can get repetitive.
@@ -78,10 +78,10 @@ The filter object can take as many properties as you want it to:
 TextField({ id: 'username-id', placeholder: 'USERNAME', visible: true }).exists();
 ```
 
-The filters available are defined by each interactor, so look at the source code for the built-in interactors or the code for your own interactors to know what is available. For example, if you take a look at the [TextField source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts), you'll see that the TextField interactor provides two filters.
+The filters available are defined by each interactor, so look at the source code for the built-in interactors to know what is available. For example, if you take a look at the [TextField source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts), you'll see that the TextField interactor provides two filters.
 
 :::note
-Since the TextField interactor extends on the [FormField interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts) which extends from the [HTML interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts), the TextField interactor inherits all of the FormField and HTML interactor properties. So in total there are actually eleven filters available for the TextField interactor. 
+There are many filters available for the TextField interactor. The TextField interactor inherits all of the [FormField interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts) and [HTML interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts) properties.
 
 We will be exploring this concept more in-depth on the [Writing Interactors](/docs/interactors/write-your-own) page.
 :::
--- a/website/docs/interactors/3-locators-filters-actions.md
+++ b/website/docs/interactors/3-locators-filters-actions.md
@@ -17,7 +17,7 @@ Button('Submit').exists();
 
 A typical user identifies a button by the words printed across it, so in this example they would consider a button with the word 'Submit' on it as the "Submit" button. Interactors use locators to make that connection.
 
-What is going on behind the scenes? Just like the user, the built-in Button interactor provided by BigTest looks for a button with "Submit" on it. It uses [element.innerText](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/button.ts#L11-L12) to find the match. Or to put it another way, `Button('Submit')` returns a button element whose `element.innerText` is equal to `Submit`.
+What is going on behind the scenes? Just like the user, the built-in Button interactor provided by BigTest looks for a button with "Submit" on it. It uses [element.innerText](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/button.ts#L11) to find the match. Or to put it another way, `Button('Submit')` returns a button element whose `element.innerText` is equal to `Submit`.
 
 ### The locator is optional
 
@@ -49,7 +49,7 @@ The locator of the `TextField` interactor is the `innerText` of its associated l
 </label>
 ```
 
-_See the source code of the TextField interactor [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts)_.
+_See the source code of the TextField interactor [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts). The TextField interactor extends the [FormField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts#L6) interactor which is where it inherits the locator function from._
 :::
 
 You can think of the locator as the "default filter" since filters and locators both provide the same functionality. The reason why BigTest offers both solutions is convenience, because having to pass in an object for each interactor can get repetitive.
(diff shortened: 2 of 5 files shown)


website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -78,9 +78,13 @@ The filter object can take as many properties as you want it to:
...
 
-The [Assertions](/docs/interactors/assertions) page will cover how you can use filters to assert against Interactors in your tests. And later on the [Writing Interactors](/docs/interactors/write-your-own) page you will learn how to add your own filters to Interactors.
+:::note
+Since the TextField interactor extends on the [FormField interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts) which extends from the [HTML interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts), the TextField interactor inherits all of the FormField and HTML interactor properties. So in total there are actually eleven filters available for the TextField interactor. 
+
+We will be exploring this concept more in-depth on the [Writing Interactors](/docs/interactors/write-your-own) page.
+:::
```
minkimcello: The Assertions page is introduced again in the `Up Next` section only a few paragraphs after here so I removed it and replaced it with this side note to clarify how the filters are being inherited.

website/docs/interactors/6-write-your-own.md
```diff
@@ -17,7 +17,7 @@ There are four things to decide when creating an interactor:
 3. The locator and filters, which helps users be able to narrow down the element they want to reference
 4. The action or actions that a test should `perform` when using the interactor (like `click`)
 
-In this case, let's call our Interactor ‘MyTextField’, and we’ll label it as ‘my-textfield-interactor’ to distinguish it from the built-in one. Let’s say that MyTextField will target a regular input with a custom class `.my-input`, which means its HTML selector would be `input[type=text].my-input`. For locator we’ll use the input’s placeholder, and we’ll define a value filter. Finally, we’ll define a `fillIn` action for MyTextField interactor. Putting this together, we get the following definition of an Interactor:
+Putting this together, let's create a new Interactor called 'MyTextField' with a label of 'my-textfield-interactor'. We'll specify the selector as `input[type=text]` so that it targets all the text input elements, define a `value` filter, and provide a `fillIn` action. And to differentiate from the built-in TextField interactor, we'll configure the placeholder value as its locator:
```
minkimcello: This paragraph wasn't lining up with the example so I re-wrote it. `In this case` didn't really make any sense because the bullet points above it applies to _all_ cases. I think the rewrite is more concise.

website/docs/interactors/6-write-your-own.md
```diff
@@ -66,29 +66,42 @@ import TabItem from '@theme/TabItem';
...
+import { TextField } from 'bigtest';
+
+export const MyTextField = TextField.extend('my-textfield-interactor')
+  .locator((element) => element.placeholder)
+```
+
+This approach would allow your interactor to inherit the selector, locator, filters, and actions from the interactor it is extending from. Extending from another interactor provides the convenience of not having to rewrite common filters and actions. You can overwrite any of the inherited properties to suit your needs which is what we are doing with the locator in the example.
+
+If we didn't provide the placeholder value as the locator, it would inherit the locator from BigTest's built-in TextField interactor which itself is inheriting its locator from the [FormField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts#L6) interactor.
+
```
minkimcello: I'm not sure if this is an appropriate spot to discuss what I wrote here. Please take a look at the generated Netlify preview [here](https://deploy-preview-921--bigtest.netlify.app/bigtest/docs/interactors/write-your-own/) to see if it flows naturally.

website/docs/interactors/6-write-your-own.md
```diff
@@ -216,23 +229,28 @@ Let's get back to our example and add the new TextField interactor to a test. In
...
-The [TextField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts) interactor from BigTest can do a lot more than this, but the example we just wrote is a good place to start when it comes to understanding how to use `createInteractor`.
+By composing our own text field interactor, we are able to use it with its new locator value without which we would have needed to use a filter for each instance:
+
+```js
+TextField({ placeholder: 'email' }).fillIn('batman@gmail.com');
+MyTextField('email').fillIn('batman@gmail.com');
+```
 
-Check out the source code of [createInteractor()](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/create-interactor.ts) for more details.
+Although the built-in interactors may cover most use cases, the composability of interactors means there is no limit to how much you can optimize and tailor them to your needs.
```
minkimcello: New approach to wrapping up the first example.

website/docs/interactors/6-write-your-own.md
```diff
@@ -304,7 +322,9 @@ export const TableCell = createInteractor('table cell')
 This example uses [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) which is available in Node >=14.
 :::
 
-You'll notice we created `columnTitle` and `rowNumber` filters that will access its parent elements to get the appropriate value we're looking for. The locator was not specified, so it will default to `element.textContent`. We can now effectively use the TableCell interactor as we stated in the beginning of this page:
+Once again, by extending from the `HTML` interactor, our new TableCell interactor inherits all of the pre-defined filters and actions of the `HTML` interactor as defined [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts). If we needed to access a table cell's `id` property in our tests, we would not need to create a separate filter for it as it inherits it from the `HTML` interactor.
```
minkimcello: I don't know if this is needs to be reiterated. I might be fixating on the obvious too much.

website/docs/interactors/2-built-in-dom.md
```diff
@@ -49,6 +52,14 @@ export default test('BigTest Runner')
   .assertion(Page.has({ title: 'BigTest Example App'}));
 ```
 
+### HTML
+
+The `HTML` interactor is not meant to be used directly in your tests but for _composing_ other interactors. How this interactor is used will make much more sense once you read the `Write Your Own Matchers` page. But for now you can think of the `HTML` interactor as the basic building block for all other interactors.
+
+We took many of the common HTML properties and interactions, such as `className` and `click`, and added them into the `HTML` interactor. This provides the convenience of not having to respecify these properties over and over again for each of your interactors.
+
+Take a look at the [source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts) for the `HTML` interactor to see which filters and actions were added.
```
minkimcello: This one was tricky to discuss. I think a quick description here is warranted because it's unique from the other built-in interactors just like `Page`. ("Why is HTML there? Is Button and CheckBox _not_ HTML?")

But at this point in the docs, the reader hasn't been exposed to the interactor properties yet so they won't know what `filters` or `locators` are nor will they understand what `"extending" interactors` means so I felt like I needed to be as generic and vague as possible.

jorgelainfiesta: I think we should remove `HTML` from the built-in interactors because we don't want people to use it as an interactor. Making that change would enable us to get rid of this `HTML` section in this page, and return to it when people will use it (`Writing Interactors`).  

Also, to me `HTML` feels weird to me because the all caps notation makes me think it's a scalar constant (and likely, syntax highlighters will treat it as such).

website/docs/interactors/2-built-in-dom.md
```diff
@@ -49,6 +52,14 @@ export default test('BigTest Runner')
   .assertion(Page.has({ title: 'BigTest Example App'}));
 ```
 
+### HTML
+
+The `HTML` interactor is not meant to be used directly in your tests but for _composing_ other interactors. How this interactor is used will make much more sense once you read the `Write Your Own Matchers` page. But for now you can think of the `HTML` interactor as the basic building block for all other interactors.
```
jenweber: I think rearranging this will help readers with understanding its use better,

```suggestion
But for now you can think of the `HTML` interactor as the basic building block for all other interactors. The `HTML` interactor is not meant to be used directly in your tests but for _composing_ other interactors. How this interactor is used will make much more sense once you read the `Write Your Own Matchers` page. 
```

website/docs/interactors/2-built-in-dom.md
```diff
@@ -49,6 +52,14 @@ export default test('BigTest Runner')
...
+
+We took many of the common HTML properties and interactions, such as `className` and `click`, and added them into the `HTML` interactor. This provides the convenience of not having to respecify these properties over and over again for each of your interactors.
```
jenweber:
```suggestion
Many common HTML properties and interactions, such as `className` and `click` are included in the `HTML` interactor. This provides the convenience of not having to re-specify these properties over and over again for each of your interactors.
```

website/docs/interactors/2-built-in-dom.md
jenweber:
```suggestion
Take a look at the [source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts) for the `HTML` interactor to see which filters and actions are available.
```

I think there's a chicken and egg problem here - what order should you teach concepts in? Ideally, this information would be part of API docs and it could be glossed over in the guides. Until then, I think what you have is ok. You could refer to them as "features" if you wanted to be more general.

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -49,7 +49,7 @@ The locator of the `TextField` interactor is the `innerText` of its associated l
 </label>
 ```
 
-_See the source code of the TextField interactor [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts)_.
+_See the source code for the TextField interactor [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts). The TextField interactor extends the [FormField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts#L6) interactor which is from where it inherits the locator function._
```
jenweber:
```suggestion
_See the source code for the TextField interactor [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts). The TextField interactor extends the [FormField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts#L6) interactor, so it has a locator function._
```

Just simplifying some wording here.

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -78,9 +78,13 @@ The filter object can take as many properties as you want it to:
 TextField({ id: 'username-id', placeholder: 'USERNAME', visible: true }).exists();
 ```
 
-The filters available are defined by each interactor, so look at the API docs for the built-in interactors or the code for your own interactors to know what is available. For example, if you take a look at the [TextField source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts), you'll see that the TextField interactor provides eight different filters.
+The filters available are defined by each interactor, so look at the source code for the built-in interactors or the code for your own interactors to know what is available. For example, if you take a look at the [TextField source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts), you'll see that the TextField interactor provides two filters.
```
jenweber:
```suggestion
The filters available are defined by each interactor, so look at the source code for the built-in interactors to know what is available. For example, if you take a look at the [TextField source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts), you'll see that the TextField interactor provides two filters.
```

I think only pointing to the source code is ok here.

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -78,9 +78,13 @@ The filter object can take as many properties as you want it to:
...
 
-The [Assertions](/docs/interactors/assertions) page will cover how you can use filters to assert against Interactors in your tests. And later on the [Writing Interactors](/docs/interactors/write-your-own) page you will learn how to add your own filters to Interactors.
+:::note
+Since the TextField interactor extends on the [FormField interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts) which extends from the [HTML interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts), the TextField interactor inherits all of the FormField and HTML interactor properties. So in total there are actually eleven filters available for the TextField interactor.
```
jenweber:
```suggestion
There are many filters available for the TextField interactor. The TextField interactor inherits all of the [FormField interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts) and [HTML interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts) properties.
```

Once there are more than a few of something, it's best to leave numbers out so that they don't have to be updated when the codebase changes. 

I think it is ok to cut down the explanation some here. If someone doesn't understand inheritance, they would not understand the longer explanation. If they do, then a short explanation should be enough.

website/docs/interactors/6-write-your-own.md
```diff
@@ -66,29 +66,42 @@ import TabItem from '@theme/TabItem';
 </Tabs>
 
 :::note
-`fillIn` is a function exported by `bigtest`. See the implementation [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/fill-in.ts). You can use any of the functions defined by BigTest or implement your own.
+`fillIn` is a function exported by `bigtest`. See the implementation [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/fill-in.ts#L63-L76). You can use any of the functions defined by BigTest or implement your own.
 :::
 
+In the example above we're extending from the `HTML` interactor to compose the MyTextField interactor. If you take a look at the implementation of the built-in [TextField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts) interactor, you'll see that the `value` filter and `fillIn` action is already available.
```
jenweber:
```suggestion
In the example above, we're extending from the `HTML` interactor to compose the MyTextField interactor. If you take a look at the implementation of the built-in [TextField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts) interactor, you'll see that the `value` filter and `fillIn` action are already available.
```

website/docs/interactors/6-write-your-own.md
```diff
@@ -66,29 +66,42 @@ import TabItem from '@theme/TabItem';
...
+
+In a more realistic scenario, if you want a TextField interactor that uses its placeholder value as its locator, you would just extend from the pre-existing TextField interactor like so:
```
jenweber:
```suggestion
In a more realistic scenario, if you want a TextField interactor that uses its placeholder value as its locator, you should extend from the pre-existing TextField interactor:
```

website/docs/interactors/6-write-your-own.md
```diff
@@ -66,29 +66,42 @@ import TabItem from '@theme/TabItem';
...
+
+```js
+import { TextField } from 'bigtest';
+
+export const MyTextField = TextField.extend('my-textfield-interactor')
+  .locator((element) => element.placeholder)
+```
+
+This approach would allow your interactor to inherit the selector, locator, filters, and actions from the interactor it is extending from. Extending from another interactor provides the convenience of not having to rewrite common filters and actions. You can overwrite any of the inherited properties to suit your needs which is what we are doing with the locator in the example.
```
jenweber:
```suggestion
This approach would allow your interactor to inherit the selector, locator, filters, and actions from the interactor it is extending from. Extending from another interactor provides the convenience of not having to rewrite common filters and actions. You can overwrite any of the inherited properties to suit your needs, which is what we are doing with the locator in the example.
```

jenweber (dismissed):
Firstly, excellent meme!

I left a couple wording suggestions. I think the main challenge to solve is, how should inheritance be introduced to a reader who is unfamiliar?  

There are two major paths:
- Write the docs in such a way that someone does not need to understand inheritance
- Briefly explain inheritance once, and every time thereafter, assume someone has read the explanation

If there were API docs to link to, the methods available would be clear, and these guides would not need to leak as many implementation details. Someone could just see the list and not have to care how they got there. However, they would still need to have a mental model of extending from a class, in order to write clean & readable interactors. So I think although option one would be nice, option 2 is probably the way to go.

Can you think of a single place to briefly explain inheritance/extending? Then all the other explanations could be trimmed down.

What are your thoughts?

website/docs/interactors/6-write-your-own.md
jorgelainfiesta: I think using 'inherit' is getting too into the technical detail. I would just say 'use' instead of 'inherit' in casual conversation. I don't think we should be drilling inheritance into people, it's just one of the many mechanisms behind the APIs.

website/docs/interactors/6-write-your-own.md
jorgelainfiesta: I think it's ok

jorgelainfiesta (commented):
Great work Min!! (and meme haha)

About the inheritance topic. I'm more on the camp of not mentioning it, `extend` pretty much abstracts inheritance. If we do want to talk about it, maybe we can make it very brief. I'd also say that it should only be mentioned on `Writing your own`. 

Jen also brings up an important topic: API docs. Some months ago we said we'd take of them later, maybe it's time to retake that effort, or at least plan to schedule it sooner than later. At the moment, we have <10 visitors per day, and only ~3 get to the docs, but when we go on promotion mode it might be important to not sway away potential users.

website/docs/interactors/2-built-in-dom.md
minkimcello: Yeah, I don't feel good about mentioning filters or actions here when it hasn't been formally introduced yet.

I was trying to avoid the word `available` just in case people jump to the conclusion that these actions and filters are the _only_ ones they can use.

If we keep this `HTML` section in, should I just take out this particular paragraph?

website/docs/interactors/3-locators-filters-actions.md
minkimcello: Sounds good. But what should we do about the paragraph above it?

> `The filters available are defined by each interactor, so look at the source code for the built-in interactors to know what is available. For example, if you take a look at the [TextField source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts), you'll see that the TextField interactor provides two filters.`

Should we snip out the last sentence starting with `For example`?

website/docs/interactors/2-built-in-dom.md
minkimcello: That's true. HTML doesn't really fit here because `built-in` sort of implies it's something you can use "out-of-the-box" in your tests. Thoughts @jenweber?

minkimcello: @jenweber Also, in response to your comment, I agree that option 2 is the right way to go at the moment (without API docs). I skimmed through the docs and I think if we...
- Scrub all the instances of us saying "take a look at the source code" on `Built-In Interactors` and `Locators, Filters, and Actions` page
- Remove the `HTML` section on the `Built-In Interactors` page

... then we could add a section _before_ the construction of the custom interactor. So it would be like "Here are the parts that make up an interactor: 1, 2, 3, 4. Putting this together ... this is how you do it."

And before we show the actual example, we could say "But first, we need to explain how extending works."

At the moment I have the long explanation after the fact but I think that'll raise questions when they encounter it without knowing what it is so it would be better to explain it all before we show the example.

website/docs/interactors/6-write-your-own.md
minkimcello: Yeah, reading it through again now, I say `inherit` way too much. I'll rewrite it once we decide on what to do with `HTML` on the built-in page.

website/docs/interactors/2-built-in-dom.md
jenweber: Yeah I think that solves the problem. You could still link to the source code in the rest of the prose. And I see that you are considering taking out the HTML explanation altogether or moving it, since it's not something we want people to use in the same way as the other built-ins.

website/docs/interactors/3-locators-filters-actions.md
jenweber: Here's one way you could make these flow together:

> The filters available are defined by each interactor, so look at the source code for the built-in interactors to know what is available. For example, if you take a look at the [TextField source code](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts), you'll see that the TextField interactor provides two filters. In addition, TextField inherits all of the properties of the [FormField interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts) and [HTML interactor](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/html.ts).

jenweber:
I've been thinking about how to frame the "extends/inheritance" thing. I think perhaps all this needs is to say once something like this: "TextField extends from FormField. That means TextField has all the methods and properties of FormField, plus some of its own." Then for the rest of the times we bring up inheritance, we can use less precise, more approachable language, like "it has these methods" or "you can use any methods from FormField too. I agree with Jorge that making some of these explanations more casual is good!

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -49,7 +49,7 @@ The locator of the `TextField` interactor is the `innerText` of its associated l
 </label>
 ```
 
-_See the source code for the TextField interactor [here](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts). The TextField interactor extends the [FormField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/form-field.ts#L6) interactor, so it has a locator function._
+_This is just the way the built-in TextField interactor is configured. It is possible to modify pre-existing interactors such as this TextField interactor or create your own interactors from scratch to suit your needs. We will be going over how you can do all of this on the [Writing Interactors](/docs/interactors/write-your-own) page._
```
minkimcello: I removed the explicit mention of extend. I think at this point of the docs, people who wants to dive into the source code will do so, and for the others we should keep it more general.

website/docs/interactors/3-locators-filters-actions.md
```diff
@@ -98,8 +92,6 @@ Button('Submit').click();
 
 It’s that simple! Because Interactors can only match one element, there’s no ambiguity over which submit button you clicked. Additionally, the action will not occur if the Interactor is not there, so you don’t have to worry about checking if the button exists before issuing a click action.
 
-On the [Writing Interactors](/docs/interactors/write-your-own) page we'll also explain how you can construct your own actions to suit your needs when you create your own Interactors.
```
minkimcello, line 97: I removed this line as it felt repetitive. We reference the `Writing Interactors` page earlier on the same page.

website/docs/interactors/6-write-your-own.md
```diff
@@ -80,13 +88,17 @@ export const MyTextField = TextField.extend('my-textfield-interactor')
   .locator((element) => element.placeholder)
 ```
 
-This approach would allow your interactor to inherit the selector, locator, filters, and actions from the interactor it is extending from. Extending from another interactor provides the convenience of not having to rewrite common filters and actions. You can overwrite any of the inherited properties to suit your needs, which is what we are doing with the locator in the example.
+This approach would allow `MyTextField` to inherit the selector, locator, filters, and actions from the built-in `TextField`. You can overwrite any of the inherited properties to suit your needs, which is what we are doing with the locator in this instance.
```
minkimcello: I think this rewrite makes the point much more concise?

website/docs/interactors/6-write-your-own.md
```diff
@@ -65,13 +65,21 @@ import TabItem from '@theme/TabItem';
...
+:::note Cypress
+If you're using Cypress, all of the built-in Interactors and Interactor functions will need to be imported from `@bigtest/cypress` and not `bigtest`.
+:::
 
-In a more realistic scenario, if you want a TextField interactor that uses its placeholder value as its locator, you should extend from the pre-existing TextField interactor:
+### `extend()` method
+
+In the example above, we're extending from the `HTML` interactor to compose the `MyTextField` interactor, but if you take a look at the implementation of the [built-in TextField](https://github.com/thefrontside/bigtest/blob/v0/packages/interactor/src/definitions/text-field.ts) interactor, you'll see that the `value` filter and `fillIn` action are already available.
+
+You could reimplement the `value` filter and `fillIn` action from scratch like we did in the example, however, it would be more practical to just extend from the built-in TextField interactor instead:
```
minkimcello, line 82: Hmm...

minkimcello (commented):
Summarizing the [latest changes](https://github.com/thefrontside/bigtest/pull/921/files/81af13c3274810432b7a16886ee6f1e4892e9a65..2593cc8d5fdcbcb844f2bd035ac0c0fb01a328b5) here so it's easier to see:

### New Approach
- We've decided to remove `HTML` interactor from the `Built-In Interactors` page as it's more relevant to _writing_ interactors.
  - We're now also introducing `extend()` on the `Writing Interactors` page. I squeezed this in by splitting up `Writing Your First Interactor` section into multiple subsections:
    ```
    Writing your first interactor
      extend() method
      HTML interactor
      Interactor label
      perform() method
      Using your interactor
    Writing a more complex interactor
    ...
    ```
- And as Jorge suggested, I tried my best to rewrite a lot of the stuff around extend/inheritance.

website/docs/interactors/6-write-your-own.md
minkimcello: See new update down below.

website/docs/interactors/3-locators-filters-actions.md
minkimcello: This git diff isn't outdated. Github bug I think.

website/docs/interactors/6-write-your-own.md
minkimcello: This one isn't outdated either.

jorgelainfiesta (approved):
Looks great!

PR #925, opened 2021-05-11
Add ability to return value from `perform`

### LINKED ISSUE #914: Interactor `perform` should allow return value
Sometimes it'd be useful to return a value out of a `perform`. For example in material UI interactors, we had to do [this workaround](https://github.com/thefrontside/material-ui-interactors/blob/598a4fa038a3471316b5a63621e977e528d19bd5/src/calendar.ts). It would be pretty simple to enable this, so let's just add this option.

wKich:
## Motivation

https://github.com/thefrontside/bigtest/issues/914


cowboyd:
Finding out a way to compose filters asynchronously will be something we have to tackle, but for now, let's make the change of limited scope first by just making the ability to get a value from perform. 

Another alternative would be to make a "secret" function that is not part of the API, but rather a standalone function for getting the value of a filter. Like:

```ts
function read<T>(interactor, filterName: string): Promise<T>;
```

That can be used like so:

```ts
await read(button, 'id');
await read(button, 'disabled');
```

I'm just hesitant to commit to putting a property on the interactor interface because that is API real-estate we can never get back.

/cc @jnicklas

wKich:
Yeah, the separate function is also ok. I'll create separate PR for that.

wKich:
@cowboyd is anything that blocks this PR from merging?

cowboyd (approved):
@wKich Nothing blocking, just wanted to make sure we had everyone's buy in, and I think we do. 👍🏻

PR #929, opened 2021-05-18
Create previews only if preview label exists on PR

minkimcello:
## Motivation

To conserve resources. We're generating preview packages for every PR regardless of if the user needs them.

## Approach

Put an if conditional for the preview workflow so that it'll only run if developer puts a `preview` label on their PR.

The generated comment is from a separate commit. If you look at the checks now, you'll see the preview workflow was skipped.


cowboyd (approved):
YAS!

PR #930, opened 2021-05-20
change html interactor selector to not match svg elements

wKich:
## Motivation

HTML interactor match  by `*` selector any element, not only HTML elements. In general, it's not a big deal. But some API differs between HTML and SVG elements. And these differences lead to an exception `│ ERROR actual.includes is not a function` with that source code `HTML('locator', { className: including('class') }).exists()`. Because `className` of SGV element returns non-string object https://developer.mozilla.org/en-US/docs/Web/API/SVGElement and the `including` matcher waits a string for the `actual` value. So there could be other issues when `HTML` interaction finds a SVG element.

## Approach

Change a selector to exclude any SVG elements from query selection.

### Alternate Designs

Change the `className` filter behavior to `Array.from(element.classList).join(' ')`. It's not ideal, but at least doesn't break anything.

Changing a selector for the HTML interactor could be a breaking change.

### TODOs and Open Questions

Add `SVG` interactor https://github.com/thefrontside/bigtest/issues/939 And what API we expect from it?
I don't sure about a changeset level. Should it be `major` or `minor`?


packages/interactor/src/definitions/html.ts
```diff
@@ -2,7 +2,7 @@ import { createInteractor } from '../index';
 import { isVisible } from 'element-is-visible';
 
 const HTMLInteractor = createInteractor<HTMLElement>('element')
-  .selector('*')
+  .selector('*:not(svg), *:not(svg *)')
```
cowboyd: Can you have HTML embedded inside svg? If that is the case, then wouldn't the predicate `:not(svg *)` exclude potentially valid elements?

wKich: You're right, there is foreignObject that allows embed html inside svg https://developer.mozilla.org/en-US/docs/Web/SVG/Element/foreignObject

wKich: `*:not(svg), *:not(svg *:not(foreignObject *:not(svg)))` This can cover only one level of nesting.

wKich: Fixed

cowboyd (approved):

cowboyd (approved):

PR #931, opened 2021-05-20
Throw an error if an interactor has no label

### LINKED ISSUE #820: Improve error message for createInteractor
If you don't pass a string to `createInteractor`, you get the following message:

> ERROR Cannot read property 'length' of undefined

It would be better for this error to be caught and have a message like:

> "Please provide a description for your interactor, such as `createInteractor('username input')`

minkimcello:
## Motivation
To resolve #820

## Approach
Screen Shot 2021-05-18 at 7 41 02 AM

- Continuing from the discussion in issue #820, we needed to decide between either throwing an error and display a warning but I went with the former for two reasons:
  - 1). It formats for us. Especially for Jest (and most likely cypress and bigtest too), the error message is formatted _below_ the failing test and is color coded. If you take a look at this screenshot, where we throw a warning, the console log hits before the test completes so it almost looks like the warning is related to the previous test:
      118645043-6dd70c80-b7ac-11eb-92e5-dbb316a43823
  - 2). We should force people to use interactors properly. Otherwise, they'll be getting unhelpful error messages in their test suites and I'm afraid people might place the blame on BigTest.

### TODOs
- [x] ~Are we okay with the wording? Should it be "An interactor used in this test..."?~ Changed to "An interactor..."
- [x] Once a preview package is published, I want to run it inside Cypress and BigTest runner to double check the error format
  - Bigtest:
    Screen Shot 2021-05-21 at 6 24 03 AM
  - Cypress:
    Screen Shot 2021-05-21 at 6 21 36 AM

### DIFF
--- a/packages/interactor/src/constructor.ts
+++ b/packages/interactor/src/constructor.ts
@@ -38,6 +38,9 @@ export function findElements<E extends Element>(parentElement: Element, interact
 }
 
 function findMatches(parentElement: Element, interactor: InteractorOptions<any, any, any>): Match<Element, any>[] {
+  if (!interactor.name) {
+    throw new Error('The interactor used for this test was not given a label. Please provide a name for your interactor:\n\tHTML.extend(\'my interactor\') || createInteractor(\'my interactor\')');
+  }
   return findElements(parentElement, interactor).map((e) => new Match(e, interactor.filter, interactor.locator));
 }
 
--- a/packages/interactor/src/constructor.ts
+++ b/packages/interactor/src/constructor.ts
@@ -39,7 +39,7 @@ export function findElements<E extends Element>(parentElement: Element, interact
 
 function findMatches(parentElement: Element, interactor: InteractorOptions<any, any, any>): Match<Element, any>[] {
   if (!interactor.name) {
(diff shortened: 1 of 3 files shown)


packages/interactor/src/constructor.ts
```diff
@@ -38,6 +38,9 @@ export function findElements<E extends Element>(parentElement: Element, interact
 }
 
 function findMatches(parentElement: Element, interactor: InteractorOptions<any, any, any>): Match<Element, any>[] {
+  if (!interactor.name) {
+    throw new Error('An interactor used for this test was not given a label. Please provide a name for your interactor:\n\tHTML.extend(\'my interactor\') || createInteractor(\'my interactor\')');
```
cowboyd: It's possible to use interactors outside of testing because they are a general automation tool. For example, @jgnieuwhof suggested that they could be used as screen scrapers, or any other area that calls for application automation. Can we formulate an error message that includes those too? E.g "one of your interactors was created without a name" or some variation on that?

minkimcello: Changed it to `One of your interactors was created without a name. Please provide a label for your interactor: ...`

cowboyd (approved):

PR #941, opened 2021-06-07
Effection v2

minkimcello:
## Motivation
To upgrade BigTest's dependencies of effection to `2.0.0-beta.#`.

## TODOs
Failing tests:
- [x] `agent`
- [x] `cli`
- [x] `server`


minkimcello:
@jnicklas the two remaining tests that are failing are: [9748165](https://github.com/thefrontside/bigtest/pull/941/commits/97481653cb1a72e713498c9ac458926ede4cb6ce) and [bdd0518](https://github.com/thefrontside/bigtest/pull/941/commits/bdd0518cbc966434422f9f190af4197d8e3a668f)

minkimcello:
@wKich It's not a big problem but I thought we were going to upgrade to `beta.3` in a separate pull request after this one? Did you need to upgrade in order to fix the ci problem?

wKich:
I didn't know that we were going to update to beta.3 in the next PR. Ok, I'll move beta.3 update to separate pr

minkimcello:
@wKich It's fine. I think it's okay.

jnicklas:
This is now passing locally for me, but needs some adjustment on CI

.circleci/config.yml
```diff
@@ -138,7 +138,6 @@ workflows:
             parameters:
               package:
                 - agent
-                - atom
```
cowboyd, line 141: I guess we can finally kiss `@bigtest/atom` goodbye

.circleci/config.yml
```diff
@@ -176,6 +175,18 @@ workflows:
...
+                command: nvm install 16.4.0
+            - run:
+                name: "Use Node version"
+                command: nvm use 16.4.0
+            - run:
+                name: "Install yarn"
+                command: npm install --global yarn
             - run:
                 name: "Install Microsoft Edge"
                 command: choco install microsoft-edge
```
cowboyd, line 192: Note for the future: there must be some way we can cache this sequence. Given how slow the MS build is, it might speed things up considerably.

packages/agent/shared/agent.ts
```diff
@@ -1,74 +1,48 @@
...
+    *init(scope: Task) {
+      let subscription = createStream<Command, CloseEvent>((publish) => function*(task) {
+        task.run(on(socket, 'message')
+          .map((event) => event as MessageEvent)
+          .map(event => JSON.parse(event.data) as Command)
+          .forEach(publish));
+
+        let closed = yield once(socket, 'close');
+        return closed;
+      }).subscribe(scope);
```
cowboyd: I believe `Stream` is a resource that produces a subscription, so you can just yield to it instead.

```suggestion
      let subscription = yield createStream<Command, CloseEvent>((publish) => function*(task) {
        task.run(on(socket, 'message')
          .map((event) => event as MessageEvent)
          .map(event => JSON.parse(event.data) as Command)
          .forEach(publish));

        let closed = yield once(socket, 'close');
        return closed;
      });
```

packages/agent/shared/agent.ts
```diff
@@ -1,74 +1,48 @@
...
+  data?: Record<string, unknown>;
 }
 
-export class Agent implements AgentProtocol {
-  constructor(private socket: Socket & EventTarget, private options: AgentOptions) {}
+export function createAgent(socket: Socket & EventTarget, options: AgentOptions): Resource<AgentProtocol> {
+  return {
+    *init(scope: Task) {
+      let subscription = createStream<Command, CloseEvent>((publish) => function*(task) {
+        task.run(on(socket, 'message')
```
cowboyd: Is there a reason not to use the `spawn()` operation here?

packages/cli/src/report-coverage.ts
```diff
@@ -22,7 +22,7 @@ export function* reportCoverage(config: ProjectOptions, results: TestResults): O
   }
 }
 
-function * renderReports(config: ProjectOptions, data: string): Operation<void> {
+const renderReports = (config: ProjectOptions, data: string): Operation<void> => function*() {
```
cowboyd: Why not use function syntax here?

packages/client/test/client.test.ts
```diff
@@ -1,96 +1,112 @@
...
+  return {
+    *init() {
+      let connections = createQueue<Socket>();
+
+      let app: Express = yield express();
+
+      app.ws('*', (socket) => function*() {
+        connections.send(socket);
+        yield;
+      });
```
cowboyd, line 18: Is this going to be able to handle more than one connection at a time? Should it maybe spawn this?

packages/server/src/app-server.ts
```diff
@@ -18,31 +17,32 @@ export function* appServer(options: AppServerOptions): Operation<void> {
...
       env: Object.assign({}, process.env, options.env),
+      buffered: true,
     });
 
     yield spawn(function* () {
       let exitStatus = yield child.join();
-
-      options.status.set({ type: 'exited', exitStatus });
+      let stdout = yield child.stdout.expect();
+      let stderr = yield child.stderr.expect();
```
cowboyd, line 29: Is this is dangerous if the application is running for a long time and spewing lots of output before eventually exiting?

.circleci/config.yml
jnicklas: These steps are very fast!

Screenshot 2021-08-25 at 10 30 32

.circleci/config.yml
jnicklas: 🙌

.circleci/config.yml
cowboyd: Guess I shouldn’t have just assumed windows was bad at this.

packages/cli/src/report-coverage.ts
jnicklas: Absolutely no reason we can't!

packages/agent/shared/agent.ts
jnicklas: Yes, the effection layer in @bigtest/effection-express spawns internally, so this operation will be spawned once per opened socket.

packages/server/src/app-server.ts
jnicklas: It probably isn't ideal. A better solution would be to have an internal ring buffer, so we can keep just a limited amount of output, but this is significantly harder to implement, so I went with the easier solution here for now, and we can revisit when/if this becomes relevant.

jnicklas: See #990

cowboyd (approved):

PR #943, opened 2021-06-11
Reload test iframe on url hash changes

### LINKED ISSUE #936: Page.visit() doesn't reload url with changed hash part only
For example
```ts
export default test('Foo')
  .step(Page.visit('/#/posts'))
  .step(Page.visit('/#/login')) // <= Failed by timeout
  .assertion(Page.has({ url: including('/#/login') }))
```
The second step failed by timeout because the test iframe source URL isn't changed and only the hash part is changed and the load event isn't triggered.

Refs 
- https://github.com/marmelab/react-admin/blob/master/cypress/integration/auth.js#L15-L17
- https://stackoverflow.com/questions/21893361/force-iframe-to-reload-document-when-only-src-hash-changes

wKich:
## Motivation

Fix #936 

## Approach

Add search parameter `bigtest-page-interactor-number` as a workaround

### DIFF
--- a/packages/interactor/src/page.ts
+++ b/packages/interactor/src/page.ts
@@ -11,7 +11,7 @@ const PageInteractor = createInteractor('page')({
     url: (element) => {
       let url = new URL(element.ownerDocument.location.href);
       let search = new URLSearchParams(element.ownerDocument.location.search);
-      search.delete('bigtest-page-interactor-number');
+      search.delete('bigtest-interactor-page-number');
       url.search = search.toString();
       return url.toString();
     },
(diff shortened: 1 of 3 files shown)


packages/interactor/src/page.ts
```diff
@@ -2,12 +2,20 @@ import { interaction, Interaction } from './interaction';
...
 const PageInteractor = createInteractor('page')({
   selector: ':root',
   filters: {
     title: (element) => element.ownerDocument.title,
-    url: (element) => element.ownerDocument.location.href,
-  }
+    url: (element) => {
+      let url = new URL(element.ownerDocument.location.href);
+      let search = new URLSearchParams(element.ownerDocument.location.search);
+      search.delete('bigtest-page-interactor-number');
```
cowboyd:
```suggestion
      search.delete('bigtest-interactor-page-number');
```

packages/interactor/src/page.ts
```diff
@@ -23,7 +31,11 @@ const PageInteractorInstance = Object.assign(PageInteractor(), {
         throw new Error('no test frame defined');
       } else {
         let url = new URL(appUrl);
-        url.pathname = path;
+        let [pathname = '', hash = ''] = path.split('#');
+        url.pathname = pathname;
+        url.hash = hash;
+        url.searchParams.set('bigtest-page-interactor-number', String(visitCounter));
```
cowboyd:
```suggestion
        url.searchParams.set('bigtest-interactor-page-number', String(visitCounter));
```

cowboyd (approved):
Looks good 👍🏻

PR #946, opened 2021-06-23
Filter delegation

jnicklas:
This is a proof of concept for a way to implement filter delegation.

The API looks like this:

``` typescript
const Calendar = HTML.extend<HTMLElement>("calendar")
  .selector("div.calendar")

const Datepicker = HTML.extend<HTMLDivElement>("datepicker")
  .selector("div.datepicker")
  .locator(element => element.querySelector("label")?.textContent || "")
  .filters({
    open: Calendar().exists(),
    month: Calendar().find(Header()).text(),
  })
  .actions({
    toggle: async interactor => {
      await interactor.find(TextField({ placeholder: "YYYY-MM-DD" })).click();
    }
  });
```

There are a lot of open questions and inconsistencies with this implementation, but it does kinda-sorta work!

### DIFF
--- a/packages/interactor/src/constructor.ts
+++ b/packages/interactor/src/constructor.ts
@@ -12,15 +12,17 @@ import {
   Filters,
   Actions,
   FilterParams,
+  FilterMethods,
   InteractorSpecification,
   BaseInteractor,
+  ToFilter,
 } from './specification';
 import { Filter } from './filter';
 import { Locator } from './locator';
-import { MatchFilter } from './match';
+import { MatchFilter, applyFilter } from './match';
 import { formatTable } from './format-table';
 import { FilterNotMatchingError } from './errors';
-import { interaction, check, Interaction, ReadonlyInteraction } from './interaction';
+import { interaction, interactionFilter, check, checkFilter, Interaction, ReadonlyInteraction } from './interaction';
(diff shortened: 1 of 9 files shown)


wKich:
I'll try to make a PoC tomorrow of how it could be improved

The one thing that could be confusing is
```js
  .filters({
    open: Calendar().exists(),
    month: Calendar().find(Header()).text(),
  })
```
Because `Calendar().exists()` is an interaction. But under the hood, it isn't used as interaction. Instead, it's just used as a container of the filter function.

packages/interactor/src/constructor.ts
```diff
@@ -198,28 +220,32 @@ export function instantiateInteractor<E extends Element, F extends Filters<E>, A
...
 
-    exists(): ReadonlyInteraction<void> {
-      return check(`${interactor.description} exists`, () => {
+    exists(): ReadonlyInteraction<void> & ToFilter<boolean> {
+      return checkFilter(`${interactor.description} exists`, () => {
         return converge(() => {
           resolveNonEmpty(unsafeSyncResolveParent(options), options);
         });
+      }, (element) => {
+        return findMatchesMatching(element, options).length > 0;
```
wKich, line 229: ancestors are not used here

PR #948, opened 2021-06-24
add support to define an array of locators

wKich:
## Motivation

For some Material UI interactors would be nice to find an element by more than one source, for example, `Button` might have a text and can be located by `innerText` value, and also Button might have an icon as content, in that case, it should be located by `aria-label` attribute. The current implementation has a limitation where the result depends on order how locator sources are used.

## Approach

Adding the ability to pass an array to the locator helps us use more than one source.

```js
const MuiButton = Button
  .extends('Mui Button')
  .locator([
    element => element.innerText,
    element => element.getAttribute('aria-label') ?? ''
  ])

// If user has a component `<Button aria-label="upload">Upload a new image</Button>`
// It can be found by both parameters

MuiButton('upload').exists()
MuiButton('Upload a new image').exists()
```

### DIFF
--- a/.changeset/multiple-locators.md
+++ b/.changeset/multiple-locators.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/interactor": minor
+---
+
+Allow to define multiple locator functions for interactors
--- a/packages/interactor/src/constructor.ts
+++ b/packages/interactor/src/constructor.ts
@@ -25,7 +25,7 @@ import { Match } from './match';
 import { NoSuchElementError, NotAbsentError, AmbiguousElementError } from './errors';
 import { isMatcher } from './matcher';
 
-const defaultLocator: LocatorFn<Element> = (element) => element.textContent || "";
+const defaultLocator: LocatorFn<Element> = (element) => element.textContent ?? "";
 const defaultSelector = 'div';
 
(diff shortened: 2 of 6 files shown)


jnicklas:
I'm not necessarily *against* this, but just a note that Capybara has this feature, and we ended up abusing it a little bit, which wasn't great. For example `fill_in('name', with: 'Foo')` can find the `name` field by `id`, `label`, `title` and probably some other things as well, so it can be a bit challenging to understand what's going on.

wKich:
@jnicklas yeah, that's a pretty solid point. Does the improving error message or action description help to understand?

cowboyd:
It is nice, but is the magic worth it? I ask because we could write the above examples using simple filters:

```js
await MUIButton({ label: 'upload' });
await MUIButton({ text: 'Upload an image' });
```

I've wanted this feature myself too, but each time, I keep hesitating to pull the trigger because I think it is also very easy to just use a filter and there is zero ambiguity about what is and what is not matching.

wKich:
Yeah, multiple ways to do one thing is bad for API. I thought about using all filters as locator, like a short hand variant. But I can't say if it worth

wKich:
Close it because it brings unobvious API

PR #950, opened 2021-07-20
Update browserslist

cherewaty:
## Motivation
When running `yarn start` in the `website` directory:
```
Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db
```

## Approach
~- Remove `package-lock.json` from `sample`, since the rest of this repo is managed with `yarn`.~
- Run `npx browserslist@latest --update-db` in the root, `sample`, and `website`. Those three lockfiles contained all the instances of `caniuse-lite`.

### Alternate Designs
I'm kind of surprised `dependabot` didn't just handle this particular use case without any thought from us.

Could possibly break compatibility with some older browsers, but unlikely.

## Learning
https://github.com/browserslist/browserslist#browsers-data-updating

## Screenshots
Screen Shot 2021-07-20 at 2 27 54 PM

### DIFF
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -3510,20 +3510,10 @@ caniuse-api@^3.0.0:
     lodash.memoize "^4.1.2"
     lodash.uniq "^4.5.0"
 
-caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001154:
-  version "1.0.30001154"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001154.tgz#f3bbc245ce55e4c1cd20fa731b097880181a7f17"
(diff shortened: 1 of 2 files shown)


cherewaty:
TODO: Review whether we still want `package-lock.json` in the `sample` app. ~_Maybe._~ _Now that I read more about the sample app, probably._

jbolda:
> TODO: Review whether we still want `package-lock.json` in the `sample` app. ~_Maybe._~ _Now that I read more about the sample app, probably._

@minkimcello we did this intentionally so someone could run it and it would have a lock file regardless of package manager. Right?

cherewaty:
>we did this intentionally so someone could run it and it would have a lock file regardless of package manager. 

Gotcha. I've removed that change. This now only affects two `yarn.lock` files. The `yarn.lock` and `package-lock.json` within `sample/app` had no effects from running `npx browserslist@latest --update-db`; they were recent enough already.

cherewaty:
I don't _think_ I need a changeset; `caniuse-lite` only appears to be a dev dependency?

minkimcello (approved):

PR #951, opened 2021-07-20
Bump docusaurus to v2.0.0-beta.3

cherewaty:
## Motivation
Ran into some problems attempting to customize the webpack config on the `alpha` version of `docusaurus` 2.

## Approach
Upgrade.

This change introduced stricter use of `url-loader`, where use of `url()` in CSS files was breaking. I just slapped a `webpack` ignore on those for now, since we're exploring changing the CSS build tooling in the coming days.

## Learning
https://github.com/facebook/docusaurus/releases/tag/v2.0.0-beta.3

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,6 +3,9 @@
   command = "yarn build"
   publish = "public/"
 
+[build.environment]
+  NODE_VERSION = "12"
+
 [dev]
   command = "yarn start"
   publish = "public/"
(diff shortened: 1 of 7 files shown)


minkimcello:
Screen Shot 2021-07-22 at 7 22 14 AM
I think we need to add some new [config](https://github.com/facebook/docusaurus/pull/5052) to remove the banner?

cherewaty:
Thanks @minkimcello 🙌 fixed in new commit.

jorgelainfiesta (approved):
Looking good, thanks Jeffrey!

Issue #784, opened 2021-01-21
Creating better DX for integrations


minkimcello:
PR #776 creates(created?) a separate integrations directory outside the normal workspace of the bigtest monorepo.

The current workflow looks like so:
- User creates changeset for corresponding packages in their PR
- PR gets merged with those changesets, release action does not publish any packages because the packages' versions have not been modified
- Changeset action creates a new PR after consuming those changesets by bumping the versions of the appropriate packages
- When _that_ PR is merged, the release action will detect that the new package versions have not been published yet so it will publish those packages

Ideally it would've been great if we could configure changeset to include paths to packages that are not included in the workspace config. I was able to get changeset/cli configured in the integrations directory, however, the changeset action will only run from the root of the repository meaning the changeset/cli in `integrations/` is only good for running `changeset version` and `changeset publish` locally.

I didn't think that justified installing changeset/cli so I left it out in PR #776. I have included DX instructions in the readme that the devs are expected to bump the versions manually however there is probably a higher chance that the readme is overlooked and devs will find themselves wondering why they can't add changesets for the integration packages from the root of the monorepo.

I think covector(?) and or the release app will eventually solve these issues but for the time being one possible solution could be a new Github Action for generating a reminder comment when it detects that work has been done in any of the integration packages. I could tackle it in node and it'll give me a head start on the inevitable refactor of our current actions to javascript. And best yet we can finally ditch Danger and have @frontsidejack create these comments.

minkimcello:
The main reason we split the workspaces was to exclude cypress from the main package dependencies. But now that the cypress integration package has cypress as a peer and dev dependency, we could combine everything into one workspace... in the interactor monorepo.

PR #964, opened 2021-08-02
Move cypress to peer and dev dep

minkimcello:
## Motivation

I think I jot down my thoughts about creating a better dx environment for integrations in #784 but for now we need to keep bumping Cypress in our integrations' dependencies manually.

## Approach

Bumped interactors and globals to latest (just small new features) and added cypress ^8.0.0 to peer and dev dependencies.
  - _"Why are you bumping `@bigtest/cypress` manually?"_ - That's because changesets doesn't offer multi-workspace feature (last time I checked) so unfortunately this is how it must be done.
  
## TODOs
- [x] ~I need to update docs to say "you need to install cypress separately"~ Not necessary.

### DIFF
--- a/integrations/cypress/package.json
+++ b/integrations/cypress/package.json
@@ -15,8 +15,10 @@
   ],
   "dependencies": {
     "@bigtest/globals": "^0.7.6",
-    "@bigtest/interactor": "^0.29.0",
-    "cypress": "^7.7.0"
+    "@bigtest/interactor": "^0.29.0"
+  },
+  "peerDependencies": {
+    "cypress": "^8.0.0"
   },
   "scripts": {
     "start": "cd ../../sample/ && npm install && npm run start -- -p 3000",
--- a/integrations/cypress/package.json
+++ b/integrations/cypress/package.json
@@ -31,6 +31,7 @@
     "@frontside/eslint-config": "^2.0.0",
     "@frontside/tsconfig": "^1.2.0",
     "@frontside/typescript": "^1.1.1",
+    "cypress": "^8.0.0",
     "start-server-and-test": "^1.11.7",
     "ts-node": "^9.1.1"
   },
--- a/integrations/cypress/package.json
+++ b/integrations/cypress/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bigtest/cypress",
-  "version": "0.0.4",
+  "version": "0.0.5",
   "description": "Cypress Integration for BigTest Interactors",
   "main": "dist/index.js",
   "module": "dist/index.js",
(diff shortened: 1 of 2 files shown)


cowboyd:
Two questions:

1. Should cypress be a peer dependency?
2. How are we validating that this fixes the issue?

minkimcello:
>     1. Should cypress be a peer dependency?

This was my thought too but I think someone had a strong opinion that it should be packaged so that users only need to install `@bigtest/cypress` instead of both `@bigtest/cypress` and `cypress`.
 
>     2. How are we validating that this fixes the issue?

I'm not, unfortunately. I skimmed through the changelog and saw there was a [bug fix](https://github.com/cypress-io/cypress/pull/16744) around `.should()`. So I published a preview package hoping this might address it if anyone can give it a quick try. But not knowing the details of why importing our package is causing an error to an unrelated test, I'll need to investigate it further.

cowboyd:
@minkimcello Looking at the change, it is typescript typings only, which is not relevant to FOLIO since they do not actually use typescript at all, so it seems very doubtful that bugfix will solve our issue.

We need to actually reproduce this error before we start trying to find solutions. Otherwise, we'll be playing long-feedback whack-a-mole.

integrations/cypress/package.json
```diff
@@ -14,9 +14,11 @@
...
   ],
   "dependencies": {
-    "@bigtest/globals": "^0.7.5",
-    "@bigtest/interactor": "^0.28.2",
-    "cypress": "^7.2.0"
+    "@bigtest/globals": "^0.7.6",
+    "@bigtest/interactor": "^0.29.0"
+  },
+  "peerDependencies": {
+    "cypress": "^8.0.0"
```
cowboyd: Maybe we should make this `>= 8.0.0` so that it will be forward compatible?

minkimcello: I think it won't make much of a difference whether it's `^` or `>=` (given how little tests I've written for it), but it would be nice to do a manual check for breaking changes before we tell people it's okay to use `9.0.0`. 🤷‍♂️ 

I think the bigger question might be if we should lower peerDependencies back down to `>= 6.0.0` now that we know it's not a recent change that introduced the cypress bug. So maybe something like `>=6.0.0 <9.0.0`

cowboyd: We can also say `"^6.0.0,^7.0.0.,^8.0.0"`

cowboyd:
I think we need a changeset, and then we're good to go.

minkimcello:
> I think we need a changeset, and then we're good to go.

Since we're still using changesets, we can't use it for `integrations/*` since it's not in the workspace config.

cowboyd:
How do we release it then?

minkimcello:
> How do we release it then?

It gets released when we bump the version ourselves. So when this PR is merged, the release workflow will publish the new package instead of waiting for a `version-packages` pr.

minkimcello:
@cowboyd I thought I would have to update the docs in case it said something like "the bigtest/cypress package comes with cypress as a dependency" but it looks like there's no mention of anything like that.

Should I add the warning about cypress' bug? And that `@bigtest/cypress` should only be used with interactor assertions?

cowboyd:
@minkimcello I don't think that needs to be coupled to this PR

cowboyd (approved):

PR #968, opened 2021-08-06
override lookup filters to asstertion filters

### LINKED ISSUE #966: Remove filters appearing in assertion from the lookup filters?
Per the discussion in https://github.com/folio-org/stripes-testing/pull/113

> If you extend HTML for your purposes, but you don't override the visible filter, attempting to use 

```js
ddMenu.is({ visible: false })
```

> will give you the error message that it couldn't find something matching visible = true (via the nicely presented console tables.)

In other words, you are making the assertion that visible is false, but you cannot find the element to make the assertion because visibility is false. The workaround here is that you would use

```js
Menu({ visible: false }).exists()
``` 

But that doesn't read that well, and it leaves in place the gotcha.

In other words, _we have a set of filters that we use for resolving the element, and a set of filters upon which we make assertions_

So when we say `Button('New')`, what we're really doing is saying behind the scenes `Button('New', { visible: true, disabled: false })` and when we say `Button('New').is({ visible: false })` what this (currently) says behind the scenes is `Button('New', { visible: true, disabled: false }).is({ visible: false })` which is obviously an incoherent assertion.

**Would it not be a better API to make it impossible to write such incoherent assertions?**

 In other words, because we detect a value for the `visible` filter in the assertion, we then remove the filter value from the lookup. If we did that, then `Button('New').is({ visible: false })` would be shortand for `Button('New', {disabled: false}).is({ visible: false })`

wKich:
## Motivation

Fix #966 

## Approach

Override lookup filters to filters from asserion

### DIFF
--- a/packages/interactor/src/constructor.ts
+++ b/packages/interactor/src/constructor.ts
@@ -108,7 +108,12 @@ function description(options: InteractorOptions<any, any, any>): string {
   return ancestorsAndSelf.reverse().map(ownDescription).join(' within ');
 }
 
-function getLookupFilter<E extends Element, F extends Filters<E>>(filter: Filter<E, F>, filters: FilterParams<E, F>): Filter<E, F> {
+/**
+ * Removes any default values for a filter from the lookup if that filter is present in the 
+ * assertion. Otherwise, it is not possible to make an assertion on a filter that might conflict
+ * see https://github.com/thefrontside/bigtest/issues/966
+*/
+function getLookupFilterForAssertion<E extends Element, F extends Filters<E>>(filter: Filter<E, F>, filters: FilterParams<E, F>): Filter<E, F> {
   let lookupFilter = new Filter(filter.specification, Object.assign({}, filter.filters));
   let specFilters = lookupFilter.specification.filters;
   for (let key in specFilters) {
@@ -166,7 +171,7 @@ export function instantiateBaseInteractor<E extends Element, F extends Filters<E
       let filter = new Filter(options.specification, filters);
       return check(`${this.description} matches filters: ${filter.description}`, () => {
         return converge(() => {
-          let element = resolver({...options, filter: getLookupFilter(options.filter, filters) });
+          let element = resolver({...options, filter: getLookupFilterForAssertion(options.filter, filters) });
           let match = new MatchFilter(element, filter);
           if(!match.matches) {
             throw new FilterNotMatchingError(`${description(options)} does not match filters:\n\n${match.formatAsExpectations()}`);
--- a/packages/interactor/src/constructor.ts
+++ b/packages/interactor/src/constructor.ts
@@ -108,6 +108,20 @@ function description(options: InteractorOptions<any, any, any>): string {
   return ancestorsAndSelf.reverse().map(ownDescription).join(' within ');
 }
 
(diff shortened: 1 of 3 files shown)


packages/interactor/src/constructor.ts
```diff
@@ -152,7 +152,9 @@ export function instantiateBaseInteractor<E extends Element, F extends Filters<E
       let filter = new Filter(options.specification, filters);
       return check(`${this.description} matches filters: ${filter.description}`, () => {
         return converge(() => {
-          let element = resolver(options);
+          let lookupFilter = Filter.copy(options.filter)
+          lookupFilter.overrideFilterValues(filters)
```
cowboyd: Rather than have a static function that copies, and then we mutate it. Can we just have a single, module-private function that copies and overrides in one go?

cowboyd (commented):
@wKich What should we do in the case of:

```js
await TextField({ text: 'hello' }).has({ text: 'goodbye' });
```

Currently, won't this change be equivalent to:

```js
TextField({}).is({ text: 'goodbye'});
```

Which could fail because of non-uniqueness..

I think we had discussed un-applying default filter values when that filter appeared in an assertion.

wKich:
@cowboyd yeah, you are right. Fixed

packages/interactor/src/constructor.ts
```diff
@@ -108,6 +108,20 @@ function description(options: InteractorOptions<any, any, any>): string {
   return ancestorsAndSelf.reverse().map(ownDescription).join(' within ');
 }
 
+function getLookupFilter<E extends Element, F extends Filters<E>>(filter: Filter<E, F>, filters: FilterParams<E, F>): Filter<E, F> {
```
cowboyd:
```suggestion
/**
 * Removes any default values for a filter from the lookup if that filter is present in the 
 * assertion. Otherwise, it is not possible to make an assertion on a filter that might conflict
 * see https://github.com/thefrontside/bigtest/issues/966
*/
function getLookupFilterForAssertion<E extends Element, F extends Filters<E>>(filter: Filter<E, F>, filters: FilterParams<E, F>): Filter<E, F> {
```

packages/interactor/src/constructor.ts
```diff
@@ -152,7 +166,7 @@ export function instantiateBaseInteractor<E extends Element, F extends Filters<E
       let filter = new Filter(options.specification, filters);
       return check(`${this.description} matches filters: ${filter.description}`, () => {
         return converge(() => {
-          let element = resolver(options);
+          let element = resolver({...options, filter: getLookupFilter(options.filter, filters) });
```
cowboyd:
```suggestion
          let element = resolver({...options, filter: getLookupFilterForAssertion(options.filter, filters) });
```

cowboyd (commented):
Looks good. I added some comments, referenced the original issue, and expanded on the name of the function because this is _very_ edge-case and not obvious if you're coming to this for the first time.

packages/interactor/test/create-interactor.test.ts
```diff
@@ -474,6 +474,29 @@ describe('@bigtest/interactor', () => {
       await expect(TextField('Password', { enabled: false }).exists()).resolves.toBeUndefined();
     });
 
+    it('can override default values', async () => {
```
cowboyd: I have just one last suggestion to make the test case name _very_ specific so those coming later know what's going on.
```suggestion
    it('the values in an assertion override the default values for a filter', async () => {
```

cowboyd (approved):

PR #974, opened 2021-08-10
pin effection packages in bundler

dagda1:
## Motivation

pin packages in `@bigtest/bundler`

## Approach

remove `^` from dependencies


cowboyd (changes requested):
oops! My bad. Only versions inside the monorepo should be pinned. Let's leave the `@effection` deps as-is.

cowboyd (approved):

PR #976, opened 2021-08-10
Update interactors landing page

jorgelainfiesta:
## Motivation

Update the landing page about interactors

## Approach

Trying out some typographical styles that are part of the new design, and other updates

## TODO 

- [ ] Incorporate feedback from Charles into copy


cherewaty:
@jorgelainfiesta I'll copy this over to the new repo for you.

PR #977, opened 2021-08-12
Fix broken `cy.should` assertion if `@bigtest/cypress` is used

### LINKED ISSUE #962: Importing bigtest/cypress breaks Cypress' `.should()`
@axelhunn found this in https://github.com/folio-org/stripes-testing/pull/116/files#r679459664

Expected behavior: there should be no adverse impact on Cypress by importing `@bigtest/cypress` package

wKich:
**Should be merged after release #978**

## Motivation

fix #962 

## Approach

Added `isInteraction` symbol to identify the interaction object. And then use it to choose what assertion to use.


packages/interactor/src/interaction.ts
```diff
@@ -1,3 +1,5 @@
+export const isInteraction = Symbol('interaction')
```
cowboyd: We want this to be a global symbol, not local to the module. Otherwise, if the symbol is eval'd twice (which can easily happen with the way `node_modules` works) then you'll have many copies of the symbol floating around

```suggestion
export const isInteraction = Symbol.for('interaction');
```

wKich: Didn't know about that, thanks.

cowboyd (commented):
One problem is that our integrations are not released with this repo and need to have their versions bumped manually. As such, we need to release interactors _first_ and then manually bump the `@bigtest/cypress`

.changeset/fix-cy-should.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigtest/cypress": patch
```
cowboyd:
```suggestion
"@bigtest/interactor": patch
```

wKich: Oh, yes. Exactly ) I missed, that I made changes in interactors too

cowboyd (commented):
@wKich I think we need to release the interactors with the symbol first, and then we can release cypress, so this has to be broken up into two PRS, otherwise it will release the cypress integration before the version with the symbol (which doesn't get released until the version PR is merged)

wKich:
Ok 👍

cowboyd (commented):
Ok, this has been released, so you should be able to integrate it 👍🏻

Just remember that this does not require a changeset, but instead a manual patch to `package.json`

cowboyd (approved):
@wKich let's get this rebased and merged since this is a pretty important fix

wKich:
@cowboyd, sorry, completely forgot about it :(

PR #978, opened 2021-08-13
add `isInteraction` symbol to identify interactions

wKich:
## Motivation

Prep for #962 

## Approach

Added `isInteraction` symbol to identify interaction object:
```ts
import { isInteraction } from '@bigtest/interactor';

/*... */

if (action[isInteraction]) {
  // action is interaction object
}
```

### DIFF
--- a/.changeset/add-interaction-symbol.md
+++ b/.changeset/add-interaction-symbol.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/interactor": patch
+---
+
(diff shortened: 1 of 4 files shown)


cowboyd (commented):
let's export a function `isInteraction` as a typescript predicate. That way, we don't have to import the symbol and use it to lookup.

wKich:
@cowboyd  done

cowboyd (approved):

PR #980, opened 2021-08-17
Rename interactor package

cherewaty:
## Motivation
The Interactors concept and package are not really coupled to the BigTest runner. Because interactors also work with Cypress, Jest, etc., they would benefit from a branding decoupling.

We can start publishing to the `@interactors` scope instead.

## Approach
_WIP_

We need to make sure migrations are as comfortable as possible. I need to put some thought into what the right order of operations is here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,7 +102,7 @@ jobs:
           at: .
       - run:
           name: Running tests
-          command: yarn workspace @bigtest/<<parameters.package>> test
+          command: yarn workspace <<parameters.package>> test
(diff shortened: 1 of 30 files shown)


cherewaty:
This needs a discussion issue open, with a solid order-of-operations plan.

cherewaty:
https://github.com/thefrontside/bigtest/issues/981

cherewaty:
Replaced by https://github.com/thefrontside/bigtest/pull/982, to improve clarity around Circle job names.

PR #982, opened 2021-08-18
Rename interactor package

cherewaty:
## Motivation
The Interactors concept and package are not really coupled to the BigTest runner. Because interactors also work with Cypress, Jest, etc., they would benefit from a branding decoupling.

We can start publishing to the `@interactors` scope instead.

## Next steps after merge
~- [ ] update which Circle jobs are required, since they changed names a bit to deal with multiple scopes~
- [x] create PR to switch dependencies within this monorepo to `@interactors/html`
- [x] `npm deprecate` the `@bigtest/interactor` package

## The next steps after that
- Switch `cypress` package scope
- Change `material-ui-interactors` publish path
- Create separate `interactors` monorepo
- Move `material-ui-interactors` to the new `interactors` monorepo

### DIFF
--- a/.changeset/tough-bees-brake.md
+++ b/.changeset/tough-bees-brake.md
@@ -1,0 +1,5 @@
+---
+"@interactors/html": patch
+---
+
+Rename @bigtest/interactor package to @interactors/html
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,7 +102,7 @@ jobs:
           at: .
       - run:
           name: Running tests
-          command: yarn workspace <<parameters.package>> test
+          command: yarn workspace @<<parameters.package>> test
   lint:
     executor: <<parameters.executor>>
     parameters:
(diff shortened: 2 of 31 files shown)


cherewaty:
🤔 I added the `preview` label and manually re-ran the "Publish Preview Packages" workflow job, but I'm not getting a preview package published.

@minkimcello you seem like the best person to ask about this. Am I missing something I need to do to get an `@interactors/html` preview published?

minkimcello:
> 🤔 I added the `preview` label and manually re-ran the "Publish Preview Packages" workflow job, but I'm not getting a preview package published.
> 
> @minkimcello you seem like the best person to ask about this. Am I missing something I need to do to get an `@interactors/html` preview published?

I think you might need to push another commit (after applying the label) for it to register and run.

minkimcello (approved):

cherewaty:
Commits added that keep the Circle CI jobs all named the same as before.

cowboyd:
We need to move these to a separate repo right?

cherewaty:
@cowboyd correct. I wanted to get all the packages publishing to their preferred names in the npm registry first, then we can rearrange repositories.

cowboyd:
We're already down this path, but I was thinking the path of less resistance would be to set up the packages in the other repos as just wrappers, and get the publishing set up and then move the code itself. E.g. for starters the `@interactors/html` package index would just be:

```js
export * from '@bigtest/interactor';
```

That way, the only changes to make in the `bigtest` repo is to delete the packages

PR #984, opened 2021-08-22
Update dependencies on @bigtest/interactor

cherewaty:
## Motivation
Sequel to https://github.com/thefrontside/bigtest/pull/982

## Approach
Update `@bigtest` packages that had a dependency (or `devDependency`) on `@bigtest/interactor` to new `@interactors/html` name.

## Next step
- Update the `sample` app after merging/publishing these changes.

## Question
Is the `cypress package` not hooked up to the `changesets` workflow? I manually bumped the patch version here.

### DIFF
--- a/.changeset/chilly-kids-yawn.md
+++ b/.changeset/chilly-kids-yawn.md
@@ -1,0 +1,6 @@
+---
+"@bigtest/agent": patch
+"bigtest": patch
+---
+
+Update @bigtest/interactor dependencies to @interactors/html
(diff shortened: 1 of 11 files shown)


minkimcello (approved):
> Is the cypress package not hooked up to the changesets workflow? I manually bumped the patch version here.

It is not so you're supposed to bump it yourself. Changesets doesn't have support for multi-workspaces and we needed to separate out the two back when the integration package included a dependency of cypress.

Issue #981, opened 2021-08-18
Decouple interactor packages from bigtest scope


cherewaty:
The Interactors concept and package are not really coupled to the BigTest runner. Because interactors also work with Cypress, Jest, etc., they would benefit from a branding decoupling.

We can start publishing to the @interactors scope instead.

This issue seeks to set out a plan for changing the publish path from `@bigtest/interactor` to `@interactors/html`. We would then repeat these steps for other packages, like `cypress`. This issue does not yet try to move these packages to a new repository.

1. Publish `@interactors/html` for the first time
2. Deprecate `@bigtest/interactor`: https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions
3. Modify other packages' dependencies from `@bigtest/interactor` to `@interactors/html`.

For step 2: the smoothest route might be to authenticate to NPM as the frontsidejack account on my local machine, instead of attempting to run `npm deprecate` in CI.

Hanging question: is there anything tricky I'm missing about step 1?

cowboyd:
@cherewaty I'd go ahead and set up the the other repository to move `interactors` to its own repo as soon as we can. That way we can set up publishing there and it will be decoupled from the versioning in this repo because interactors is basically ready for 1.0, but bigtest isn't.

Also, it lets us move `@interactors/with-cypress` to use changesets/covector which it does not currently.

minkimcello:
And what should we do about the interactor-related issues and PRs?

cowboyd:
> And what should we do about the interactor-related issues and PRs?

Those should be moved over to the new repository. I believe that is possible as long as they are in the same org?

cherewaty:
> I believe that is possible as long as they are in the same org?

https://docs.github.com/en/issues/tracking-your-work-with-issues/transferring-an-issue-to-another-repository

Yep, I confirmed I see the "Transfer issue" button here with `interactors` as one of the repos I can transfer to.

<blockquote><img src="https://github.githubassets.com/images/modules/open_graph/github-logo.png" width="48" align="right"><div><img src="/assets/images/site/favicon.svg" height="14"> GitHub Docs</div><div><strong><a href="https://docs.github.com/en/issues/tracking-your-work-with-issues/transferring-an-issue-to-another-repository">Transferring an issue to another repository - GitHub Docs</a></strong></div><div>To move an issue to a better fitting repository, you can transfer open issues to other repositories.</div></blockquote>

cherewaty:
Success.

PR #988, opened 2021-08-23
Rename @bigtest/cypress to @interactors/with-cypress

cherewaty:
## Motivation
Part of https://github.com/thefrontside/bigtest/issues/981

## Approach
`@bigtest/cypress` -> `@interactors/with-cypress`

I really like Next.js's use of the `with-` prefix in their examples: https://github.com/vercel/next.js/tree/canary/examples

### Alternate Designs
On Discord, we have also discussed `@interactors/within-cypress`, `@interactors/integration-cypress` and `@interactors/cypress` as possibilities.

### DIFF
--- a/integrations/cypress/README.md
+++ b/integrations/cypress/README.md
@@ -1,6 +1,6 @@
 ## Getting Started
 ```
-npm install @bigtest/cypress
+npm install @interactors/with-cypress
 ```
 
 This package includes both `cypress` and `interactors`. Please refer to our [Interactor Docs](https://frontside.com/bigtest/docs/interactors) to learn how to use Interactors with Cypress.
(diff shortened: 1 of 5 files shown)


cherewaty:
@cowboyd probably want your blessing of this name before we proceed.

cowboyd:
I'd do this name-changing once we set up publishing in the other repo. That's because publishing `with-cypress` doesn't use changesets so iterating on it is a pain. Also, it means that you can't publish the `with-cypress` along with other packages in a single step. Instead, you have to publish the core packages first, and only _then_ publish `with-cypress`

cherewaty:
> I'd do this name-changing once we set up publishing in the other repo.

Alright, I'll switch to getting that repo set up.

PR #989, opened 2021-08-25
Remove interactor package

cherewaty:
## Motivation
Sibling to https://github.com/thefrontside/interactors/pull/2. Part of https://github.com/thefrontside/bigtest/issues/981.

## Approach
- Remove `packages/interactor` directory.
- Remove related Circle job (`test-interactor`). I also removed it from repo branch protection rules.
- Remove type reference paths to `packages/interactor`.
- Backfill missing type dependencies.

## Future TODO
~`packages/webdriver/test/remote.test.ts` imported `converge()` from the `interactor` `src` internals, but `converge()` doesn't appear to be exported from `@interactors/html`. As a band-aid, I copied `converge()` and `wait()` directly into that test.~ No longer an issue.


cowboyd:
What about the cypress integration? Doesn't that depend on the interactor package?

cherewaty:
> What about the cypress integration? Doesn't that depend on the interactor package?

It depends on `@interactors/html` now, so that should be sufficiently decoupled for this specific PR.

cherewaty:
~Couldn't resolve conflicts on my machine; blocked by https://github.com/thefrontside/bigtest/pull/992~

Worked around this by temporarily removing `volta` stanza from `package.json` to run `yarn`. No longer blocking, but should still proceed with setting consistent node versions.

cherewaty:
Rebased and ready for reviews.

minkimcello (approved):

cowboyd:
I wonder if we should release a "last" version of the package that basically is a pass through to `@interactors/html` and a deprecation message saying to use that package instead.

cherewaty:
>I wonder if we should release a "last" version of the package that basically is a pass through to @interactors/html and a deprecation message saying to use that package instead.

I ran `npm deprecate` for `@bigtest/interactor`: https://www.npmjs.com/package/@bigtest/interactor

1. This message shows up in the registry UI:
Screen Shot 2021-08-30 at 8 40 30 AM

2. That message will also show up when you try to `npm install` any version of `@bigtest/interactor`.

3. `@bigtest/interactor` no longer appears in search results on npmjs.com.

Is that enough, or do we also need a "runtime" deprecation warning, not just at install?

<blockquote><div> npm</div><div><strong><a href="https://www.npmjs.com/package/@bigtest/interactor">@bigtest/interactor</a></strong></div><div>Interactors for working with applications</div></blockquote>

minkimcello:
Screen Shot 2021-08-30 at 9 48 12 AM

The deprecation warnings do not pop out enough in my opinion but since our docs have already been updated I think it would be fine for the most part? 🤷‍♂️

PR #992, opened 2021-08-26
Set node to 14 and yarn to 1.22

cherewaty:
## Motivation
When trying to run `yarn` after https://github.com/thefrontside/bigtest/pull/941 merge, I ran into:

`error @parcel/source-map@2.0.0-rc.6: The engine "node" is incompatible with this module. Expected version "^12.18.3 || >=14". Got "12.11.1"
error Found incompatible module.`

I'm running `volta` to manage my local tools, but it had `node` set to the incompatible `12.16.0`.

Additionally, Circle was running node 15 in some instances, and 16 in others.

Netlify was running node 12.

## Approach
Run the same `node` and `yarn` everywhere.

I chose to stick to `node` 14, the current LTS release.

## Alternate Designs
I doubt everyone here is using `volta`. Should we be considering other tool version tools, like `asdf`?

### TODOs and Open Questions

- [ ] Is there a particular reason we need to be running node 16 anywhere?
- [ ] Do we need to cut releases of all the packages now delivering a different `@types/node`? Patch? Minor?


cowboyd (approved):

cowboyd (approved):

cherewaty:
Experiencing timeouts on Windows Circle jobs trying to install from `chocolatey`. 😓 I'll come back to it in a bit and see if that issue resolves itself.

PR #995, opened 2021-08-27
Add introduction for `mui` and add `Existing Interactors` page

minkimcello:
## Motivation

Add mui to docs, showcase `folio` interactors, and adapt to `interactor` becoming its own package.

## Approach
- Changed `Built-in DOM` page to `Predefined Interactors`
  - `Why?` - In the process of trying to find an appropriate place to add interactor-adapters examples (`mui`/`folio`), I figured the best page would be where we list the "built-in" ones. And it made me think that maybe "built-in" has a "it's set in stone" connotation when in reality it can be modified and reshaped to users' needs. "Predefined? Cool, I'm going to RE-define _that_ to my liking".
      - This is also my reason why this is clumped together in one PR instead of two.
- ~~Changed `Storybook` page to `Storybook & Material UI`~~ Created `Existing Interactors` page
  - Introduce folio interactors from [this](https://github.com/folio-org/stripes-testing/tree/master/interactors) repo as examples
  - Talk about how we wrote interactors for mu components so that people can write tests for their mu app.
  - Talk about how mui can be used to write stories in storybook.

- Many of our examples were importing interactors from the `bigtest` package. And we also had many "Bigtest's interactors". So i've read through the entire docs to update the imports and also reword some of the phrasing to treat interactors as its own package.

## Questions
- [ ] I omitted the preview deployment of our story. I'm not entirely sure how much value it will provide to someone already familiar with storybook and material ui. And the code used for the preview deploy is not too well organized. See [here](https://github.com/thefrontside/material-ui-interactors/blob/storybook/stories/interactors.stories.tsx). I think the timeout used in those examples might throw people off.
- [x] Is it safe to assume bigtest will always include a dependency of interactors? Because in some examples of the bigtest runner, we could have both the interactors and runner modules being imported from `bigtest` but at the moment I've separated them out so that runner-specific ones are being imported from `bigtest` and interactors from `@interactors/html`.
- [x] ~~I left the `@bigtest/cypress` stuff as is until [this](https://github.com/thefrontside/interactors/pull/35) is complete.~~
- [x] Need to add section for `mui` and `folio` in `predefined interactors`
  - [x] The call-to-action heading is `Show Us Your Interactors!` which could sound a bit provocative depending on _how_ you say it. But if there's a better heading for it, we should change it.

### DIFF
--- a/website/docs/interactors/1-quick-start.md
+++ b/website/docs/interactors/1-quick-start.md
@@ -90,7 +90,7 @@ If you are using Cypress, you will only need to install `@interactors/with-cypre
 
 Interactors can do a lot, but let's keep things simple for now and begin testing one of the most common user interactions - a button click.
 
-Interactors have methods like `click` that mimic user actions. If you are using your own app, in your codebase find a test that already has a button click in it. In that test, import the `Button` interactor from `bigtest` and use it to replace the click interaction as exemplified below (make sure to substitute "Sign In" with your own button text):
+Interactors have methods like `click` that mimic user actions. If you are using your own app, in your codebase find a test that already has a button click in it. In that test, import the `Button` interactor from `@interactors/html` and use it to replace the click interaction as exemplified below (make sure to substitute "Sign In" with your own button text):
 
 <Tabs
   groupId="runner"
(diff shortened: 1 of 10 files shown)


taras (approved):
👍 looks and reads great

cowboyd:
@minkimcello I do not think that `bigtest` will include a dependency for interactors. The reason is because `@interactors/html` only makes sense when the agent is a browser. However, we're going to be rolling with multi language agents soon, and so it won't make sense in that context.

PR #997, opened 2021-09-01
Mention that regex is also acceptable for locator

minkimcello:
## Motivation

[PR #34](https://github.com/thefrontside/interactors/issues/34) from [thefrontside/interactors](https://github.com/thefrontside/interactors)

## Approach

This one was a lot simpler than anticipated. We only mention strings+locators once when we introduce locators as a concept. Nowhere else do we bring up strings+locators. So I reworded it to say "string _or_ a regular expression" and added a second example for the locator (one string, one regex).

### DIFF
--- a/website/docs/interactors/3-locators-filters-actions.md
+++ b/website/docs/interactors/3-locators-filters-actions.md
@@ -9,10 +9,11 @@ Whether they are built-in or written by you, all interactors have some things in
 
 One benefit of Interactors is that they help you align your tests with how a user actually interacts with the app, starting with finding what to act upon. A locator is the simplest way to find a specific element in a user interface.
 
-When you use an Interactor in a test, you can pass it a string. This string argument would be the Locator - like "Submit" in the example below:
+When you use an Interactor in a test, you can pass it a string or a regular expression. This argument would be the Locator - like for "Submit" in the examples below:
 
 ```js
-Button('Submit').exists();
+Button('Submit').exists(); // string
+Button(/Submit/).exists(); // regex
 ```
 
 A typical user identifies a button by the words printed across it, so in this example they would consider a button with the word 'Submit' on it as the "Submit" button. Interactors use locators to make that connection.


cowboyd (commented):
I don't know how technical we want to get, but precisely, the `locator` of an interator is _still_ only ever a string. It's just that we will match the regex against that locator by default now. In other words, `Button(/Submit/)` is sugar for `Button(matching(/Submit/))`

minkimcello:
> I don't know how technical we want to get, but precisely, the `locator` of an interator is _still_ only ever a string. It's just that we will match the regex against that locator by default now. In other words, `Button(/Submit/)` is sugar for `Button(matching(/Submit/))`

I think from the users' perspective, they just need to know the two options for the locator because it doesn't change the way they would use it.

We could say "alternatively, you could use regex which is just a shortcut for a matcher (you'll learn about matchers later)" but I'm afraid this could potentially feel overwhelming especially from the first page.

Or if we were to mention it in the matchers page and say "oh, remember how you can use regex for the locator? That's actually wrapped in _this_ particular matcher" but at this point of the docs, it seems irrelevant.

cowboyd:
I'm just worried that we're being incorrect. A regex is not a locator, the locator is always a string. The distinction isn't important when you use interactors, but when you write them, it definitely is. because you have the `locator()` function in the build which must only ever return a string.

minkimcello:
> I'm just worried that we're being incorrect. A regex is not a locator, the locator is always a string. The distinction isn't important when you use interactors, but when you write them, it definitely is. because you have the `locator()` function in the build which must only ever return a string.

Is it incorrect to say `... you can pass it a string or a regular expression. This argument would be the Locator ...`? Later in the `creating your own interactors` page, we never explicitly state that it must return a string. We just imply it by giving examples of `element.placeholder` and `element.innerText`.

I could clarify by adding "although the argument can be string or regex, remember that the locator function must always return a string" but I'm fairly certain that'll cause more confusion. The modified page is fairly early on in the docs and i don't think people are ready to be distinguishing between _using_ and _creating_ interactors.

As in I don't think we have to worry about people trying to write a locator that returns regex.

cowboyd:
My specific nitpick is this phrase "This argument would be the Locator " because the locator is not the argument, it is the string returned by the locator function. I don't want to confuse things, but I also don't want to say something that isn't true that will confuse them later on when understanding locators because they are a very subtle, very powerful, and very unique to interactors.

minkimcello:
Oh, right. We've been down this road before. Jen and I couldn't find a good way to write this.

"This argument of your interactor is the value that's passed into the locator function" will be _very_ confusing. This is early on in the docs where readers will not be even thinking about constructing locator functions or how interactor works under the hood.

I think in this context, even though it's not technically technically correct, it gets the right message across. And anyone that's read through the entire docs won't come back and go "hey, what? this is technically incorrect!".

We did ponder over that very sentence but I'm open to any suggestions.

minkimcello:
From https://github.com/thefrontside/bigtest/blame/8e86068a9a2ae8b6073f8021373d7b4d624a00ea/website/docs/interactors/3-locators-filters-actions.md#L24-L30

> Whenever you use an Interactor in a test, you can pass it a string, like "Submit" in the example below. This string argument is used by the Locator.

`... While "This string argument is used by the Locator" is technically correct, the follow up is then: what is a locator, and that is not really explained. Even though this is somewhat confusing, I would almost advocate to refer to the actual *value* as the locator, and so the function which is passed to the specification is a locator function and it evaluates to the locator. I think this would be easier to understand.`

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/thefrontside/bigtest">bigtest/website/docs/interactors/3-locators-filters-actions.md at 8e86068a9a2ae8b6073f8021373d7b4d624a00ea · thefrontside/bigtest</a></strong></div><div>Ensure your React/Vue/Ember/anything app works perfectly across browsers. - bigtest/website/docs/interactors/3-locators-filters-actions.md at 8e86068a9a2ae8b6073f8021373d7b4d624a00ea · thefrontside...</div></blockquote>

cowboyd (approved):

cowboyd:
@minkimcello 
> “This argument of your interactor is the value that's passed into the locator function" will be very confusing.”

For the record, This is also not true (but I don’t know if it would change the outcome though). The locator is the value returned by the locator function, full stop. 

The argument is what is compared against the locator.

PR #1000, opened 2021-09-02
Update cypress package name in sample app

cherewaty:
## Motivation
Sequel to https://github.com/thefrontside/bigtest/pull/996

## In the future
Should the `sample` app still include Cypress and Jest examples in this repo? Maybe it should only include examples with the BigTest runner?

### DIFF
--- a/sample/app/app-pkg.json
+++ b/sample/app/app-pkg.json
@@ -18,7 +18,7 @@
     "@babel/core": "^7.12.10",
     "@babel/preset-env": "^7.12.11",
     "@babel/preset-react": "^7.12.10",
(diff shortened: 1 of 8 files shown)


minkimcello:
> Should the sample app still include Cypress and Jest examples in this repo? Maybe it should only include examples with the BigTest runner?

I think interactors-sample might be the one that should come with both jest and cypress? And the bigtest sample app (if and when that happens) can just include bigtest... i think. 🤷‍♂️

minkimcello:
It may or may not be able to publish a non-existent package. So you might need to manually publish an empty package first.

minkimcello (approved):

Issue #1005, opened 2021-09-08
ChromeDriver doesn't match Chrome version in Circle browsers image


cherewaty:
`test-agent`, `test-cli`, and `test-webdriver` jobs on Circle CI are failing.

`Error: session not created: This version of ChromeDriver only supports Chrome version 91`

It appears the `circleci/node:14-browsers` image now includes Chrome 93.

cowboyd:
We need a durable solution for this. Most other frameworks just require chromedriver to be installed as a dependency. If we want to auto-install, we should probably first detect the version of chrome that is _actually installed_ and then work backwards and download the version of chromedriver needed to control it, because it's such a nagging problem.

#308 #885 #887 #894 #935

PR #1004, opened 2021-09-08
Remove interactors docs

cherewaty:
## Motivation
https://github.com/thefrontside/bigtest/issues/981

## Approach
- Remove `interactors` pages.
- Modify the platform landing page to site at the root.

PR #1006, opened 2021-09-09
Upgrade chromedriver to fix Circle job

cherewaty:
## Motivation
Related to https://github.com/thefrontside/bigtest/issues/1005

## Approach
I don't love kicking this can down the road again, but I really don't want to block https://github.com/thefrontside/bigtest/pull/1004 on this issue.

## Need to follow up on
`@interactors/html` needs to drop its dependency on `@bigtest/globals` and `@bigtest/performance`. Caused some `yarn.lock` additions that should've already been in the `v0` branch.

### DIFF
--- a/.changeset/lovely-jobs-joke.md
+++ b/.changeset/lovely-jobs-joke.md
@@ -1,0 +1,5 @@
+---
+"@bigtest/webdriver": minor
+---
+
+Require chromedriver 93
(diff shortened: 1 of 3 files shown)


minkimcello (approved):

cowboyd:
@cherewaty The dependency on `@bigtest/performance` should be a no-brainer since it was implemented as a ponyfill for the web performance API in node. Since `@interactors/html` is browser-only, we can just drop the dependency and use https://developer.mozilla.org/en-US/docs/Web/API/Performance

<blockquote><div><strong><a href="https://developer.mozilla.org/en-US/docs/Web/API/Performance">Performance - Web APIs | MDN</a></strong></div><div>The Performance interface provides access to performance-related information for the current page. It's part of the High Resolution Time API, but is enhanced by the Performance Timeline API, the Navigation Timing API, the User Timing API, and the Resource Timing API.</div></blockquote>

PR #1021, opened 2021-10-13
Extract UI

cowboyd:
## Motivation

In order to really perceive the value of what BigTest brings to the table, a picture is worth a thousand words. While the raw GraphQL dashboard is certainly powerful in what it can display (which is almost everything). Learning the system, and see where the dials and knobs are is still quite opaque. Eventually, we're going to want to make a UI that can do all of that for us.

## Approach

Following the pattern of `@simulacrum/ui`, This creates a separate `@bigtest/ui` package that is a web-app bundled with parcel that can connect to _any_ bigtest server. At the moment, all it does is load the GraphiQL component and point it at the orchestrator, and so functionally, the behavior is exactly the same as before. However, the difference is that we now have a standalone React app that we can modify to our liking.


jnicklas (approved):

jnicklas:
Great idea!

PR #1032, opened 2021-10-20
decoupling interactors from bigtest globals

wKich:
## Motivation

The second step of decoupling interactors from bigtest https://github.com/thefrontside/interactors/issues/3

## Approach

- Use `setInteractionWrapper` to add `check` method to interactions to restrict running actions in assertions.
- Move `visit` action from the `Page` interactor as a separate function
- `@bigtest/globals` uses `@interactors/globals` for some properties


jnicklas:
Since this also includes the `visit` function, it should include a changeset for that as well. Or better yet, maybe we could make that commit a separate PR?

I'm not sure which I prefer, but maybe `visit` could be spelled with a capital letter, like: `Visit` that way it feels similar to other interactors. I feel like that'd be more natural when using it with the step DSL:

```typescript
.step(Visit("/"))
.step(Button("Foo").click())
```

But maybe less so when using regular `async/await`:

```typescript
.step("visit homepage and click button", async() => {
  await Visit("/");
  await Button("Foo").click();
});
```

wKich:
@jnicklas I renamed PR to make the `visit` function fit better.

I'm not sure about naming `visit` from capital letter, because it's not an interactor it is just action.

packages/agent/app/run-lane.ts
```diff
@@ -112,6 +113,12 @@ export function* runLane(config: LaneConfig): Operation<TestImplementation> {
   }
 
   setLogConfig({ events: [] });
+  setInteractionWrapper((interaction) => ({
+    ...interaction,
+    check() {
+      throw new Error(`tried to ${interaction.description} in an assertion, actions/perform should only be run in steps`)
+    }
+  }))
```
jnicklas: If we make interaction wrappers "stack" then this way of setting and resetting the interaction wrapper won't work anymore. Also, I'm pretty sure that overriding the `check` method like this will only work when using the DSL form and not when using `async/await`.

wKich: I don't understand why it won't work?

About the `async/await`, where is it used in `run-lane`?

jnicklas: Sorry, I did not explain this very well. When an interaction is created there are two ways of consuming it:

1. with the DSL:
  ``` typescript
  .step(Button('Foo').click())
  ```
2. As a promise:
  ``` typescript
  .step(async() => {
    await Button('Foo').click()
  })
  ```

This way of overriding `check` will work when the interaction is consumed through the DSL, but I don't think it will work when used as a promise. This is where we are creating the interaction in interactors:

https://github.com/thefrontside/interactors/blob/c98dafe58226df133d56a45fc91f59d8b8cfc683/packages/html/src/interaction.ts#L45-L65

As you can see, when using the interaction as a promise we always call `action`, never `check`, and so overriding `check` has no effect. Additionally, `action` is closed over here, so reassigning the property on the object would have no effect on how it is consumed as a promise.

wKich: Oh, yeah. I didn't notice how step consumes an interaction. I thought about that. The problem is we wrap interaction, but we need to wrap an action instead. Made the fix for that https://github.com/thefrontside/interactors/pull/108

jnicklas (approved):

PR #1033, opened 2021-10-21
update eslint-config and typescript

wKich:
## Motivation

The eslint-config was updated to use new version of eslint

## Approach

Update the eslint-config dependency and bump `typescript` with `tsconfig` versions as well


packages/agent/package.json
```diff
@@ -28,9 +28,9 @@
   "devDependencies": {
     "@bigtest/suite": "^0.12.1",
     "@bigtest/webdriver": "^0.11.0",
-    "@frontside/eslint-config": "^2.0.0",
+    "@frontside/eslint-config": "2.1.0-2fd3c42",
```
jnicklas: Why does it use a preview package?

wKich: Because the update PR isn't merged yet https://github.com/thefrontside/javascript/pull/36

jnicklas (approved):
LGTM!

PR #1041, opened 2021-11-16
Add Effection labels

jnicklas:
Adds some labels to our various components, so they render better in the effection inspector.


cowboyd (approved):
labels? yes plz!

PR #1050, opened 2021-11-25
Remove @bigtest/effection dependency

jnicklas:
We missed this when removing @bigtest/effection. It might be causing the changeset action to fail.


cowboyd (approved):
Do we need to remove it from changesets entirely and also maybe the workspaces list? Or is that done already

jnicklas:
> Do we need to remove it from changesets entirely and also maybe the workspaces list? Or is that done already

I think this is already done.

PR #1065, opened 2021-12-13
Update interactors to 1.0rc

wKich:
## Motivation

HP uses bigtest with old interactors version. And installing the last version of `@interactors/with-cypress` there makes a little mess with hosting `@interactors/html` package.

## Approach

Update `@interactors/html` in bigtest


cowboyd (approved):

taras:
@wKich we should really remove bigtest

PR #1067, opened 2021-12-21
spawn request handler directly on proxy task

cowboyd:
## Motivation

The `spawn(operation).within(task))` syntax is going to be deprecated, as it is no longer needed. There is a `Task#spawn` operation which can be used to spawn a task directly from its parent.

## Approach

1. spawn directly off of the proxy task
2. Remove the `blockParent` option since the `onEmit` operation is infinite and it is impossible for the handler to _ever_ finish before its parent and is therefore redundant.

### DIFF
--- a/.changeset/remove-deprecated-spawn-form.md
+++ b/.changeset/remove-deprecated-spawn-form.md
@@ -1,0 +1,4 @@
+---
+"@bigtest/server": patch
+--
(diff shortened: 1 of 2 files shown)


dagda1 (approved):
👍

taras (approved):
👍

PR #1068, opened 2021-12-21
Use task group for collecting tasks with the same lifetime

cowboyd:
### Motivation
Instead of using an option 'blockParent'. We can explicitly spawn different tasks into collections of groups that can be waited upon individually. This not only makes which group a task is part of explicit, but it allows flexibility in how you consume the blocking tasks.

This is fairly common in other structured concurrency frameworks like Swift or ember-concurrency, and can be expanded upon to add other aspects like "how many concurrent things can be going at the same time" or "what do we do when the max concurrency is attained... do we drop incoming operations? do we respond with a "503 operation"? do we cancel the currently running things? do we enqueue the new operations? How many do we enqueue?"

So this is an experiment in what that might look like.

### DIFF
--- a/packages/server/src/command-server.ts
+++ b/packages/server/src/command-server.ts
@@ -122,7 +122,7 @@ function handleSocketConnection(options: CommandServerOptions): (socket: Socket<
   }
 
   return (socket) => function*() {
-    let handlers: TaskGroup = yield createTaskGroup('message handlers');
+    let handlers = yield* createTaskGroup('message handlers');
 
     yield socket.forEach(function*(message) {
       if (isQuery(message)) {
@@ -136,6 +136,6 @@ function handleSocketConnection(options: CommandServerOptions): (socket: Socket<
       }
     });
 
-    yield handlers;
+    yield handlers.allSettled();
   }
 }
--- a/packages/server/src/command-server.ts
+++ b/packages/server/src/command-server.ts
@@ -11,6 +11,7 @@ import { GraphqlContext } from './schema/context';
 import { Slice } from '@effection/atom';
 import { OrchestratorState, CommandServerStatus } from './orchestrator/state';
 import { Runner } from './runner';
+import { TaskGroup, createTaskGroup } from './task-group';
 
 import { Variables, Message, Response, QueryMessage, MutationMessage, SubscriptionMessage, isQuery, isMutation, isSubscription } from '@bigtest/client';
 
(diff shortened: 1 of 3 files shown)


packages/server/src/task-group.ts
```diff
@@ -0,0 +1,38 @@
...
+    name,
+    *init(scope) {
+      return {
+        *spawn<T>(operation: Operation<T>) {
+          let task: Task<T> | undefined;
+          yield spawn(function*() {
+            let t: Task<T> = task = yield scope.spawn(operation);
+            tasks.add(t);
+            try {
+              yield t.future;
```
cowboyd: We don't care how this task ended, only that it did. Since it was spawned in its parent, it will fail it if needs be.

packages/server/src/task-group.ts
```diff
@@ -0,0 +1,38 @@
...
+              tasks.delete(t);
+            }
+          });
+          assert(!!task, "task was not initialized");
+          return task;
+        },
+        *init() {
+          for (let task of tasks) {
+            yield task.future;
+          }
```
cowboyd: Probably need to wrap this in a `while (!!tasks.size) { }` in case new tasks are added to the group while it is shutting down.

packages/server/src/task-group.ts
```diff
@@ -0,0 +1,38 @@
...
+  let tasks = new Set<Task>();
+
+  return {
+    name,
+    *init(scope) {
+      return {
+        *spawn<T>(operation: Operation<T>) {
+          let task: Task<T> | undefined;
+          yield spawn(function*() {
+            let t: Task<T> = task = yield scope.spawn(operation);
```
dagda1: why is there a local `t` and `task` var?

dagda1:
Looks like a good idea

Should this not be part of effection itself?

At the moment, I don't see any max task restriction, is this still to be added?

packages/server/src/task-group.ts
cowboyd: Excellent point! It was mainly there to help with the typing because the type of `task` has to be `Task<T> | undefined` and I can't add it to the set or delete it even if it's undefined. In truth though, it is probably better not to have two storage locations in memory when one will suffice.

cowboyd:
> Looks like a good idea
> 
> Should this not be part of effection itself?
> 
> At the moment, I don't see any max task restriction, is this still to be added?

Yes, this is more of a proving ground for the idea, and then when we're happy with it we can add it to Effection proper. I was thinking something like

```ts
function* run() {
  let group = yield createTaskGroup('socket handlers', {
    maxConcurrency: 200,
    drop: 'latest'
  });
}
```

cowboyd:
@dagda1 Yeah, this is something that I think should eventually be added to Effection itself, but that I was using BigTest as a testing ground.  I'd imagine it would be something like:

```ts
let group = yield createTaskGroup('socket handlers', {
  maxConcurrency: 10,
  drop: 'latest'
});
```

packages/server/src/command-server.ts
```diff
@@ -122,7 +122,7 @@ function handleSocketConnection(options: CommandServerOptions): (socket: Socket<
   }
 
   return (socket) => function*() {
-    let handlers: TaskGroup = yield createTaskGroup('message handlers');
+    let handlers = yield* createTaskGroup('message handlers');
```
dagda1, line 125: 💯

